PR #2, opened 2019-04-03
updated eslint config, updated jsdoc

ivadimko:
sorry, wrong commit message, pay attention to PR name

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -2,7 +2,8 @@
 
 /**
  * Implement convertToObject function:
- * Function gets string with styles
+ *
+ * Function takes string with styles
  * and returns object where CSS properties are keys
  * and values are the values of related CSS properties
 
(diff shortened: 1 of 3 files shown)


src/convertToObject.js
```diff
@@ -2,7 +2,8 @@
 
 /**
  * Implement convertToObject function:
- * Function gets string with styles
+ *
+ * Function takes string with styles
```
roman-mate: Nit: each line should be almost 80 characters long.

mgrinko: I agree with splitting based on meaning rather than just length

roman-mate: https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/AbstractIterator.java#L87 -- this is an example of good documentation. 

Different meaning separated into paragraphs. Lines within each paragraph are as close to 80 characters as possible.

PR #6, opened 2019-04-09
Solution

KonstVolkovnitsky:


src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const separateStyles = sourceString.split(';\n');
+  const stylesObject = {};
+  function addToObject(style, index, arr) {
+    const transitionalArray = style.trim().split(': ');
+    stylesObject[transitionalArray[0]] = transitionalArray[1];
+  }
+  separateStyles.forEach(addToObject);
+  delete stylesObject[''];
```
YuriiHoliuk: It is bad practice to use delete [prefer assign](https://mate-academy.github.io/style-guides/javascript.html#objects--rest-spread)
I suggest you add `if` inside `addToObject`

PR #12, opened 2019-04-10
solved by valerii tovstyk

valeriitovstyk:


src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  let splitedArray = sourceString.split(';');
```

src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
...
+  for (let item of splitedArray) {
```
YuriiHoliuk: use forEach. Always prefer array methods

src/convertToObject.js
valeriitovstyk: done

src/convertToObject.js
valeriitovstyk: done, thanks

PR #15, opened 2019-04-10
Yevhen Kryvdiuk

Yevhen-Kryvdiuk:


src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceObj = {};
+  let sourceArray = sourceString.split(';');
```

PR #17, opened 2019-04-10
'Solution'

dvitalijb:


src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let result = {};
+  let properties = sourceString.split(';');
```

src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
...
+  for (let property of properties) {
```
YuriiHoliuk: use forEach

PR #24, opened 2019-04-11
'fs_apr19_viacheslav solution'

YojikYo:


src/convertToObject.js
```diff
@@ -12,7 +12,14 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const propertyObj = {};
+  const propertiesArray = sourceString.split(';');
+
+  for (let property of propertiesArray) {
```
YuriiHoliuk: will be better to use array methods with arrow function(forEach and arrow callback)

YojikYo: Done

PR #30, opened 2019-04-14
Ostap_Solution

ost-ptk:


src/convertToObject.js
```diff
@@ -12,7 +12,13 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';').map(el => el.slice(3).split(': '))
```
roman-mate: 1. splice(3) -- someone may pass a string into this function which uses 4 spaces, or tabs. 

we should use trim() instead. 

2. same with ':  ' -> use ':' and use trim().

src/convertToObject.js
```diff
@@ -12,7 +12,13 @@
...
+    .reduce((sum, curent) => {
```
roman-mate: sum is a bad name. we are not really summing anything.

res or obj would be better.

roman-mate (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,13 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';').map(el => el.split(':'))
+    .reduce((obj, curent) => {
+      if (curent[1] !== undefined) {
```
danmysak: `current`

PR #74, opened 2019-08-16
solution

shklyarigor2015:


src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const array = sourceString.split(';');
```
YuriiHoliuk: What if you will have another array here? It will be `array2`?
Variable name should describe content inside, not type.

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
...
+  const resultConvert = {};
```
YuriiHoliuk: `convertResult` will be better, but still bad
which convert? result of what? function result or temporary result?
`stylesObject` - will be better name

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
...
+  let properties = [];
```
YuriiHoliuk: instead of declaring `let` variables, declare const in place where you need it

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
...
+  for (let i = 0; i < array.length; i++) {
+    array[i] = array[i].trim();
```
YuriiHoliuk: Do not mutate array without necessary, create constant instead

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
...
+    properties = array[i].split(': ');
```
YuriiHoliuk: properties not the best name, because variable contains 1 css property, `nameValuePair`, `style`
Or use destruction and create 2 variables instead

src/convertToObject.js
```diff
@@ -12,7 +12,14 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArrFrom = sourceString.split(';');
```
irenhh: `sourceArr` is enough

PR #78, opened 2019-08-16
Kseniia's solution

KseniiaMazan:


src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceObj = {};
+  const sourceArr = sourceString.split('\n  ').join('').split(';');
```
irenhh: instead of splitting the string twice and then shorten it, you can split once and then trim() each element in the for loop

PR #84, opened 2019-08-16
Vyacheslav-P style-to-object solution

Vyacheslav-P:


src/convertToObject.js
```diff
@@ -12,7 +12,28 @@
...
+
+  const arrFromSourse = sourceInLine.split(';');
+  const arrayOfProperties = [];
+  for (const item of arrFromSourse) {
+    if (item !== '') {
+      arrayOfProperties.push(item);
+    }
+  }
+
+  const ListOfProperties = {};
```
irenhh: variable names starts from lowercase letter

Vyacheslav-P: fixed

PR #85, opened 2019-08-16
Vika Tsykun solution

ViktoriiaTsykun:


src/convertToObject.js
```diff
@@ -11,8 +11,17 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+  const sourceArray = sourceString.slice(1).split(';');
```
irenhh: why do you need slice here?

PR #86, opened 2019-08-16
Obj styles

Vladyslav223:


src/convertToObject.js
```diff
@@ -13,6 +13,15 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const sourceStringNned = sourceString.replace(/\n/g, '');
+  const obj = {};
+  const arr = sourceStringNned.split(';');
+  for (let i = 0; i < arr.length; i++) {
```
irenhh: new line before and after for loop for better readability

PR #92, opened 2019-08-17
Solution

marchukmiroslava88:


src/convertToObject.js
```diff
@@ -12,7 +12,13 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString.split(';');
+  const stylesObj = {};
+  for (let i = 0; i < styles.length - 1; i++) {
```
irenhh: new line before and after for loop for better readability

PR #93, opened 2019-08-17
solution

youvovk:
[LINK](https://github.com/mate-academy/js_style-to-object/compare/master...youvovk:fs_jul19_yurii?expand=1)

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const createPropObj = {};
+  const splitKeyValue = sourceString.split(';');
+
+  for (let i = 0; i < splitKeyValue.length; i++) {
+    const styleCssProp = splitKeyValue[i].trim().split(': ');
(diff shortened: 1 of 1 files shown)


src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const createPropObj = {};
```
irenhh: verbs are used to name functions

src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
...
+  const splitKeyValue = sourceString.split(';');
```
irenhh: `stylesArr` is a better name

src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const propsObj = {};
```
danheim:
```suggestion
  const stylesObject = {};
```

PR #94, opened 2019-08-18
Solution

kosbielozorov:


KosBelozyorov:
прибрав метод map

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectResult = {};
+  const inputString = sourceString.split(';');
+  for (let i = 0; i < inputString.length; i += 1) {
```
irenhh: you can write just `i++`

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectResult = {};
+  const inputString = sourceString.split(';');
```
irenhh: you will have an array in the end, don't name it as string

irenhh (changes requested):

src/convertToObject.js
KosBelozyorov: i know it, but airbnb and google recommend to write "i += 1"

KosBelozyorov:
add fix

src/convertToObject.js
danheim: I agreed with you

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectResult = {};
```
danheim:
```suggestion
  const stylesObject = {};
```

danheim (approved):

KosBelozyorov:
Fixed

PR #95, opened 2019-08-18
sol

Taisiiapn:


src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.split(';');
+  const obj = {};
```
irenhh: what if you have another `arr` and `obj` in your code? give more specific names

PR #96, opened 2019-08-18
Solution

ArtemMarkevych1:


src/convertToObject.js
```diff
@@ -12,7 +12,14 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const solutionArr = sourceString.split(';'); // get a split array
+  const solutionObj = {}; // object for getting solution
+  for (let i = 0; i < solutionArr.length; i++) {
```
irenhh: new line before for loop

ArtemMarkevych1: New empty line is added:
image

PR #97, opened 2019-08-18
'Solution'

Serhii-P:


src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+
+  const arr = sourceString.split(';');
+  for (let i = 0; i < arr.length; i++) {
+    const elements = arr[i].trim().split(': ');
+
```
irenhh: no need for empty line here

irenhh:
what about other comments?

PR #98, opened 2019-08-18
solution by Chernij

ChernijAnna:


src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectFromString = {};
+  const splitString = sourceString.trim().split(';');
```
irenhh: this is not a string, rename a variable

irenhh:
what about another comment?

PR #100, opened 2019-08-19
solution

Volodsher:
Good morning! Thank you very much for testing my small program creatures!

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -1,18 +1,16 @@
 'use strict';
 
-/**
- * Implement convertToObject function:
- *
- * Function takes string with styles
- * and returns object where CSS properties are keys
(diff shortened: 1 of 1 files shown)


src/convertToObject.js
```diff
@@ -1,18 +1,16 @@
...
-
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const newObj = {};
+  let property;
+  const newArr = sourceString.split(';\n');
```
irenhh: what if there is no \n in the source string? split by `;` better, you will always have it in css code

Volodsher: Thanks, corrected that.

PR #101, opened 2019-08-19
my sol

koretskyi:


src/convertToObject.js
```diff
@@ -12,7 +12,13 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const strToObj = {};
+  const strToArr = sourceString.split(';');
+  for (let i = 0; i < strToArr.length; i++) {
```
irenhh: new line before and after for loop (for better readability)

PR #128, opened 2019-10-01
convert

sosibona:


src/convertToObject.js
```diff
@@ -12,7 +12,14 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const a = sourceString.split('\n  ').join('').split(';');
+  a.pop(a.length);// delete last element (\n)
```
gestap: did `pop` accepts arguments?

src/convertToObject.js
```diff
@@ -12,7 +12,14 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const a = sourceString.split('\n  ').join('').split(';');
```
gestap: what is this `a`??
never name variable like this.

src/convertToObject.js
```diff
@@ -12,7 +12,14 @@
...
+  a.pop(a.length);// delete last element (\n)
+  const css = {};
+  for (let i of a) {
```
gestap: what is `i`?
rename

PR #131, opened 2019-10-02
'Solution'

Xandr1:


src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const expected = {};
+  const propertiesAndKeys = sourceString.split(';');
+  propertiesAndKeys.pop();
+
+  propertiesAndKeys.forEach(function(element) {
```
gestap: better to write 
```suggestion
  propertiesAndKeys.forEach(element => {
```

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
...
+    const styles = element.split(': ');
```
gestap: use `trim` on styles[1]
```suggestion
    const styles = element.split(':');
```

gestap: If it would be with 2 whitespaces `:  `?

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const expected = {};
+  const propertiesAndKeys = sourceString.split(';');
+  propertiesAndKeys.pop();
```
gestap:
```suggestion

  propertiesAndKeys.pop();
```

PR #169, opened 2020-01-29
AnastasiiaBardyuzha: js_style-to-object solution

AnastasiiaBardyuzha:


src/convertToObject.js
```diff
@@ -12,7 +12,23 @@
...
+    .split(';');
+
+  correctStringFormat.splice(-1);
+
+  const convertedObj = {};
+
+  correctStringFormat.forEach((item) => {
+    let splitItem = item;
+
+    splitItem = splitItem.split(/:\s?/);
```
irenhh: no need to do this action separately

PR #173, opened 2020-01-29
Natalia Lypovyk: Style to Object

natalia-lypovyk:


src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const list = sourceString
+    .split(';')
+    .map(item => item
+      .trim())
```
irenhh: leave this on the prev line

src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
...
+    .filter(el => el.length > 0)
+    .map(item => item.split(':')
+      .map(i => i.trim()));
+
+  const map = new Map(list);
```
irenhh: no need to create map

src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
...
+
+  const stylesObject = Object.fromEntries(map);
```
irenhh: you can return this directly

PR #178, opened 2020-01-29
convertToObject git test

cursorksu:


src/convertToObject.js
```diff
@@ -10,9 +10,25 @@
...
+ * git test
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const cssObj = {};
+
+  sourceString.split(';')
+    .forEach(prop => {
+      const coupleProp = prop.split(':');
```
irenhh:
```suggestion
      const splittedProp = prop.split(':');
```

src/convertToObject.js
```diff
@@ -10,9 +10,25 @@
...
+
+      if (coupleProp.length > 1) {
+        const rule = coupleProp[0].substr(1, coupleProp[0].length).trim();
```
irenhh: do you really need `substr` method here?

PR #263, opened 2020-06-02
My solution

solaryasha:
Hello there. Thanks for the review.

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -13,6 +13,15 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const styleProperties = {};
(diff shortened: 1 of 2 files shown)


src/convertToObject.js
```diff
@@ -13,6 +13,15 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const styleProperties = {};
+
+  const regex = /([^\s;`]+)\s*:\s*([#\w\d-]*\s?[!.#\w\d-]*\s?[#\w\d-]+)/g;
```
Dead-TR, line 18: 👍

PR #292, opened 2020-08-11
sol

karinaKudatska:


src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
...
+
+  const obj = {};
+
+  arr.map(item => {
+    const prop = item.trim().split(':');
+
+    if (prop[0]) {
+      obj[prop[0].trim()] = prop[1].trim();
+    }
+  });
```
irenhh, line 25: map should return something, it's not a good idea to do side effects in map

PR #293, opened 2020-08-11
solution

Andrii256:


src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  sourceString
+    .split(';')
+    .map(property => property
+      .split(':')
+      .map(elem => elem.trim()))
+    .forEach(elem => (elem[0]) && (result[elem[0]] = elem[1]));
```
irenhh, line 22: 1. forEach should not return anything
2. don't do assignment when returning something even if it works with parentheses, it's confusing

PR #295, opened 2020-08-11
solved

nanccyy:


src/convertToObject.js
```diff
@@ -8,11 +8,21 @@
...
+//  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  const finalArray = sourceString.split(';')
+    .map(element => (element.split(':')))
+    .filter(group => (group.length > 1));
```
irenhh, line 19: parentheses around return statement are not necessary since the statement is short

PR #296, opened 2020-08-11
solution

varseniuk:


src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArr = sourceString.split(';');
+  const result = {};
+
+  stylesArr.forEach((style) => {
+    const property = style.split(':');
+
+    if (property[1]) {
```
irenhh, line 21: why do you check only the second part?

varseniuk: Because in some cases, splitted string created an empty element of array of length = 1, which contained only one space and nothing more. For example:
correct element - [" padding-right", " 30px"] (length = 2)
incorrect element- [" "] (length = 1)
Because of that, I've checked, if THERE IS second element in that subarray, and if it is not - I've skipped that element at all.

But ok. Just in case added condition for checking property[0] as well))

PR #298, opened 2020-08-12
Solution

aliceblunt8:


src/convertToObject.js
```diff
@@ -11,8 +11,21 @@
...
+
+  sourceString
+    .split(';')
+    .map(prop => {
+      const [ key, value ] = prop.split(':');
+
+      if (value) {
+        stylesObj[key.trim()] = value.trim();
+      }
+    });
```
irenhh, line 26: map should return something with no side effects

PR #299, opened 2020-08-12
soltuion

masyatya:


src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resObj = {};
+  const arr = sourceString
+    .split(';')
+    .map(elem => elem.split(':'))
+    .map(elem => elem.map(a => a.trim()))
```
irenhh, line 19: a is a bad name, nobody will understand what's going on here

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
+    .filter(elem => elem[0] !== '');
+
+  for (const [key, value] of arr) {
+    resObj[key] = value;
+  }
```
irenhh, line 24: use arr methods

PR #300, opened 2020-08-12
solution

oggyrock:


src/convertToObject.js
```diff
@@ -12,7 +12,12 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const s = sourceString
+    .split(';')
+    .map(e => e.split(':').map(x => x.trim()))
+    .filter(z => z[1]);
```
irenhh, line 18: e, x, z 🤯

src/convertToObject.js
```diff
@@ -12,7 +12,12 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const s = sourceString
+    .split(';')
+    .map(e => e.split(':').map(x => x.trim()))
```
irenhh, line 17:
```suggestion
    .map(e => e.split(':')
    .map(x => x.trim()))
```
be consistent

oggyrock: it's inside first method, does this make it better?

PR #303, opened 2020-08-12
solution

furude-rika:


src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
...
+    .map(item => item.split(':')
+      .map(elem => elem.trim())
+    );
+
+  const filteredArray = trimmedArray.filter(item => item.length !== 1);
+
+  filteredArray.forEach(item => {
+    result[item[0]] = item[1];
+  });
+
```
irenhh, line 28: those two steps could be combined

PR #304, opened 2020-08-12
solution

tykhan:


src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splitedProp = sourceString.split(';');
```
irenhh, line 15:
```suggestion
  const splittedProp = sourceString.split(';');
```

src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
...
+
+  const objectProp = {};
+
+  splitedProp.map(item => {
+    const eachProp = item.split(':');
+
+    if (eachProp[0] && eachProp[1]) {
+      objectProp[eachProp[0].trim()] = eachProp[1].trim();
+    }
+  });
```
irenhh, line 25: map is not for side effects, use forEach

tykhan: changed
but not quite understood, what does it mean hear 'for side effects'?

vkryvytskyy: 'side effects' means that for each element in array you do some action. Map is for `changing each element in array`

PR #305, opened 2020-08-12
solution

oivannikov:


src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objStyle = sourceString.split(';')
+    .map(space => space.trim())
+    .filter(space => space)
```
irenhh, line 17:
```suggestion
    .filter(Boolean)
```

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
+    .map(pair => pair.split(':')
+      .map(keyValue => keyValue.trim()))
```
irenhh, line 19: this could be done in the next step

PR #309, opened 2020-08-12
Solution

KaterinaSavvova:


src/convertToObject.js
```diff
@@ -13,6 +13,18 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const ans = {};
```
leonbohdan, line 16: give an understandable name for variables

Denchepornyuk: Погоджуюсь!

PR #314, opened 2020-08-12
first solution

leonbohdan:


src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const refactored = {};
```
irenhh, line 15: refactored what?

leonbohdan: refactoredStyles

PR #325, opened 2020-08-14
Solution

1nsider21:


src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+
+  const arr = sourceString.split(';').map(elem => elem.trim());
+
+  const newArr = arr.filter(char => char);
```
irenhh, line 19:
```suggestion
  const newArr = arr.filter(Boolean);
```

src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
...
+
+  newArr.map(item => {
+    const [key, value] = item.split(':');
+
+    obj[key.trim()] = value.trim();
+  });
+
+  return obj;
```
irenhh, line 27: don't do side effects in map, it has to return something

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+
+  const arr = sourceString.split(';').map(elem => elem.trim());
+
+  const newArr = arr.filter(char => Boolean(char));
```
irenhh, line 19:
```suggestion
  const newArr = arr.filter(Boolean);
```
you can write even shorter

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
...
+
+  newArr.map(item => {
+    const [key, value] = item.split(':');
+
+    obj[key.trim()] = value.trim();
+
+    return item;
+  });
```
irenhh, line 27: you're trying to do actions that are usually done in forEach

where do you return `item`? you don't assign the result anywhere

PR #357, opened 2020-09-30
solution convert

Andrii1996:


goncharukop (approved):

src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const parameters = sourceString.split(';');
+  const styleParameters = {};
+
+  parameters.map(parameter => {
+    const splitParameter = parameter.split(':');
+    const first = splitParameter[0];
+    const second = splitParameter[1];
```
Vladimir-Beliy, line 15: key, value

Vladimir-Beliy: никакх first, second

Vladimir-Beliy (changes requested):

src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
+
+  parameters.map(parameter => {
+    const splitParameter = parameter.split(':');
+    const key = splitParameter[0];
+    const value = splitParameter[1];
+
+    if (key && value) {
+      styleParameters[key.trim()] = value.trim();
+    }
+  });
```
Vladimir-Beliy, line 20:
```suggestion
  parameters.map(parameter => {
    const splitParameter = parameter.split(':');
    const key = splitParameter[0].trim();
    const value = splitParameter[1].trim();

    if (key && value) {
      styleParameters[key] = value;
    }
  });
```

Vladimir-Beliy (approved):

PR #397, opened 2020-11-28
Add solution convertToObject

nevskyy:


vkryvytskyy (changes requested):
try to simplify the solution. It's very complicated for now

vkryvytskyy (approved):

PR #403, opened 2020-12-17
Convert to Object solution

VinnieJ-2k20:


zhenya-mezhueva (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  let arr = sourceString.split(';');
+
+  arr = arr.map(property => property.trim());
+  arr = arr.filter(property => property.length > 0);
+
+  arr = arr.map(function makeArrays(property) {
+    return property.split(':');
+  });
```
varseniuk, line 22:
```suggestion
  let arr = sourceString.split(';')
    .map(property => property.trim())
    .filter(property => property.length > 0)
    .map(property => property.split(':'));
```

don't write functions inside methods. They already take callbacks that you can use

varseniuk (changes requested):

varseniuk (approved):

PR #405, opened 2020-12-18
Solution

vorfolomeevav:


Lara-l-art (approved):
Great!

varseniuk (approved):
Nice solution.
But it would be more readable if you split elements by ';' first and then by ':' (moving from bigger to smaller, and not vice-versa) But it's ok

PR #406, opened 2020-12-18
solution

Lara-l-art:


vorfolomeevav (approved):
Good job!

src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayFromString = sourceString
+    .split(';')
+    .map(item => item.split(':'))
+    .filter(item => item.length > 1);
```
vsm2k17, line 18: you can avoid this, checking if elements exist before adding them to an object.

src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayFromString = sourceString
```

vsm2k17 (changes requested):

vsm2k17 (approved):
nice one

PR #407, opened 2020-12-19
Solution

vovador:


Serhii-Alieksieievych (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const ofSArr = sourceString.split(';');
```
vsm2k17, line 16: ofSArr bad name

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
+  const opArray = ofSArr.map(x => x.trim().split(':').map(y => y.trim()));
```
vsm2k17, line 17: opArray bad name, don't use types as a name.
x, y => bad names, don't use letters as a name.
no need to use trim on every step, use it before adding it into an object.

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
+  const filteredArray = opArray.filter(el => el.length > 1);
```
vsm2k17, line 18: just before adding check if the element exists

vsm2k17 (changes requested):
good job but you can make it better

vsm2k17 (approved):
good job

PR #410, opened 2020-12-20
Solution

YuriiFitsyk:


Roman-Matsuk (approved):
Well done!

src/convertToObject.js
```diff
@@ -12,7 +12,23 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArray = sourceString
```

src/convertToObject.js
```diff
@@ -12,7 +12,23 @@
...
+    .split('\n')
+    .join('')
+    .split(';')
+    .map(param => param.split(':')
+      .map(part => part
```
vsm2k17, line 20: try to avoid using map inside another map

src/convertToObject.js
```diff
@@ -12,7 +12,23 @@
...
+        .split(' ')
+        .filter(x => x.length > 0)
```
vsm2k17, line 22: x => bad name. don't use letters as names

src/convertToObject.js
```diff
@@ -12,7 +12,23 @@
...
+        .join(' ')))
+    .filter(param => param.length === 2);
```
vsm2k17, line 24: redundant, just check if an element exists before adding to an object.

vsm2k17 (changes requested):

src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
...
-  // write your code here
+  const sourceOfStyles = sourceString
+    .split(';')
+    .map(param => param
+      .split(':')
+    );
+  const result = {};
+
+  for (const param of sourceOfStyles) {
+    if (param[1]) {
```
vsm2k17, line 23: logic is understandable = > if you got the second element then you have first by default, but better to check the existence of both of them

vsm2k17 (approved):
good job

PR #412, opened 2020-12-20
Solution

AlexysUa:


src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  let pairs = sourceString.split(';');
+
+  pairs = pairs.map(e => e.trim()).filter(el => el.length > 0);
```
vsm2k17, line 18: use trim only once. not here.

vsm2k17 (changes requested):
good job

src/convertToObject.js
AlexysUa: Done!

vsm2k17 (approved):

PR #416, opened 2020-12-21
Solution

Roman-Matsuk:


YuriiFitsyk (approved):
Good job!

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObj = {};
```

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
...
+  const sourceStringArr = sourceString.split('\n').join('').split(';')
```
vsm2k17, line 16:
```suggestion
  const sourceStringArr = sourceString.split(';')
```

vsm2k17: redundant

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
...
+    .map(item => item.split(':')).filter(item => item.length > 1)
```
vsm2k17, line 17: the filter can be replaced by checking if elements exist before adding them to an object.

vsm2k17 (changes requested):

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+  const sourceStrings = sourceString.split(';')
+    .map(item => item.split(':'))
+    .map(item => item.map(part => part.trim()));
```
vsm2k17, line 18: too many map methods, you can delete at least one

vsm2k17 (changes requested):

vsm2k17 (approved):
good job

PR #417, opened 2020-12-21
Solution

vitalya2626:


NikitaTudakov (approved):
Great!

src/convertToObject.js
```diff
@@ -12,7 +12,11 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const final = sourceString.split(';')
+    .map(rule => rule.split(':')
+      .map(rulePiece => rulePiece.trim()))
+    .filter(rule => rule.length > 1);
```
varseniuk, line 18:
```suggestion
  const final = sourceString.split(';')
    .map(rule => rule.split(':')
    .map(rulePiece => rulePiece.trim()))
    .filter(rule => rule.length > 1);
```
Why 'length > 1' but not > 0?

varseniuk (changes requested):

src/convertToObject.js
vitalya2626: The test doesn't work with 'length > 0' and linter is not happy with that change as well)

varseniuk: Oh, I haven't noticed that you use the second 'map' INSIDE the first 'map', not AFTER it. Sorry. Ok. then the formatting is proper.

varseniuk (approved):

src/convertToObject.js
vitalya2626: Thank you for the review!

PR #419, opened 2020-12-21
solution

HaidaiAndrii:


zhenya-mezhueva (approved):

src/convertToObject.js
```diff
@@ -11,8 +11,21 @@
...
+  let stringCSS = sourceString.split(';');
+  const result = {};
+
+  stringCSS = stringCSS.map(x => x.replace(/\n|\s\s+/g, ''));
+  stringCSS = stringCSS.filter(x => x.length > 1);
+
+  stringCSS.forEach(el => {
+    const [rule, value] = el.split(':');
+
+    result[rule] = value.trim();
```
vsm2k17, line 25: you can also use a trim method for rule

src/convertToObject.js
```diff
@@ -11,8 +11,21 @@
...
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  let stringCSS = sourceString.split(';');
+  const result = {};
+
+  stringCSS = stringCSS.map(x => x.replace(/\n|\s\s+/g, ''));
+  stringCSS = stringCSS.filter(x => x.length > 1);
```
vsm2k17, line 20: don't use letters as a names

vsm2k17: no need to use filter just check if an element exists before adding to an object and use a trim.

src/convertToObject.js
```diff
@@ -11,8 +11,21 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  let stringCSS = sourceString.split(';');
```
vsm2k17, line 16: don't reuse one variable, better to create a couple of constants than reassigning one.

vsm2k17 (changes requested):

src/convertToObject.js
```diff
@@ -11,8 +11,21 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  sourceString.split(';')
+    .map(item => item.replace(/\n|\s\s+/g, ''))
```
vsm2k17, line 19: no need to use regexp

vsm2k17 (approved):

PR #426, opened 2020-12-21
style-to-object solution

Nvalja:


src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let strArray = sourceString.replace(/[;]/g, '').split('\n');
```
alexmogwaiii, line 15: good 🙂

alexmogwaiii (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
+  const result = {};
+
+  strArray = strArray
+    .filter(x => (x.length > 0) && (x !== '  '))
+    .map(x => x.split(':'));
```
varseniuk, line 20: It's never a good idea to use such kind of checks `x !== '  '`. What if there would be 3 spaces? Or 4? Then your code would not work. You can cut all unnecessary spaces in such way:
```suggestion
  strArray = strArray
    .map(x => x.trim())
    .filter(x => x.length > 0)
    .map(x => x.split(':'));
```

varseniuk (changes requested):

src/convertToObject.js
Nvalja: done

varseniuk (approved):
👍

PR #430, opened 2020-12-21
Solution

khrulenko:


dima-bond (approved):
well done!

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
```

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
...
+
+  const arr = sourceString.split(';').map(
+    el => el.split(':')
+  ).filter(a => a.length > 1);
```
vsm2k17, line 19: you can avoid filter method by simply checking if elements exist before adding to an object.

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
...
+
+  const arr2 = arr.map(el =>
```
vsm2k17, line 21: change names, arr2 === bad name

vsm2k17 (changes requested):

vsm2k17 (approved):
good job

PR #431, opened 2020-12-21
solution

Droshik:


taniazt (approved):
Cool!

src/convertToObject.js
```diff
@@ -11,8 +11,19 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .map(rule => rule
+      .trim()
```
vsm2k17, line 19: no need to trim here, the second trim is enought

vsm2k17 (changes requested):
nice solution, but you got an O(x^2) algorithm complexity here because of the map method inside of another map method,  try to use some machine memory and make it faster, O(x) maybe

src/convertToObject.js
```diff
@@ -11,8 +11,21 @@
...
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObj = {};
+
+  sourceString
+    .split(';')
+    .map(stile => stile.trim())
+    .filter(style => style.length > 0)
```
vsm2k17, line 21: you can remove this line and check if elements exist before adding them to an object

vsm2k17 (approved):
good job

PR #432, opened 2020-12-21
solution

dima-bond:


khrulenko (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cleanProps = sourceString
+    .split(';')
+    .map(prop => prop.split(':'))
+    .filter(x => x.length > 1)
+    .map(prop => [prop[0].trim(), prop[1].trim()]);
```
vsm2k17, line 19: you can implement these actions before adding properties for an object and remove it from here

vsm2k17 (approved):

PR #435, opened 2020-12-22
solution

alexmogwaiii:


Nvalja (approved):
good job)))

src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrOfStrings = sourceString.split(';')
+    .map(item => item.split(':')
+      .map(prop => prop.trim()));
```
vsm2k17, line 17: don't use map inside another map, you can avoid it here

vsm2k17 (changes requested):

src/convertToObject.js
alexmogwaiii: Why i can't use it? :c

vsm2k17: you can, but that makes your code very slow, reed about algorithm complexity.

vsm2k17 (approved):

PR #436, opened 2020-12-22
Solution

Layrovell:


5Mountains (approved):
Good solution!

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+
+  sourceString.split(';')
+    .forEach(el => {
+      const str = el.split(':');
+
+      if (str.length > 1) {
```
varseniuk, line 21: Why str.length > 1? Why 1? Not 0, not 3, not 100?
How did you get this number?

varseniuk (changes requested):

varseniuk (approved):
Using check `if (str.length > 1)` is not a good practice. You should check if there both elements in the array and it may be done in a more understandable way.
E.g. `if (str[0] && str[1])`
Take it into consideration for future

PR #437, opened 2020-12-22
style to object

kerpichonka:


Pasha5555 (approved):
Great!

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stringsWithProperty = sourceString.split(';');
+  const trimStrings = stringsWithProperty.map(prop => prop.trim());
+  const filterStrings = trimStrings.filter((prop) => prop !== '');
+  const styles = filterStrings.map(prop =>
+    prop.split(':').map(value => value.trim()));
```
varseniuk, line 19:
```suggestion
  const styles = sourceString.split(';')
    .map(prop => prop.trim())
    .filter((prop) => prop !== '')
    .map(prop =>
      prop.split(':').map(value => value.trim()));
```

Your variant is ok. It is very clear and readable, but you could also simplify it like this.
Just FYI

varseniuk (approved):
Good job

PR #441, opened 2020-12-22
solution

5Mountains:


Pasha5555 (approved):
Great solution!

src/convertToObject.js
```diff
@@ -12,7 +12,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+
+  sourceString.split(';').forEach(item => {
+    const property = item.trim().split(':');
+
+    if (property.length > 1) {
```
varseniuk, line 20: Why exactly `property.length > 1`? Why not 0 or 4 or 10?

varseniuk (changes requested):
why property.length > 1 ?

src/convertToObject.js
5Mountains: > why property.length > 1 ?

because "property" is an array that should contain 2 values in order to fill CSS **"property: value"** Via this method I cut all unnecessary items that do not contain such rules. 😕 Do you believe that this is the magic number?

5Mountains: or you meant that I should write the exact number (property.length === 2) that is required for the condition?

varseniuk: Yes, that's exactly what I meant. If you want to check the EXACT AMOUNT of elements, it's always better to SPECIFY this exact amount directly. `>1`, in this case, can mean anything. Let's try to make the code clearer and more understandable.
But for this task, you could use anything else. E.g. to check `if (property[0] && property[1])`. That would make your code more understandable for others. In the case of `> 1` somebody can misunderstand the purpose of this part of the code

varseniuk (approved):
👍

src/convertToObject.js
5Mountains: I got it! 😀 Many thanks for the clarification!

PR #442, opened 2020-12-23
Solution

andrew-zaycev:


sashalarin01 (approved):
Great job👍

src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  sourceString.split(';')
+    .map(value => value.split(':'))
+    .filter(value => value.length > 1)
```
varseniuk, line 19: Why `value.length > 1`? 
Why do you use exactly number 1? Why not 0? Not 3? Not 50?

varseniuk (changes requested):

src/convertToObject.js
andrew-zaycev: Becouse in tests need filtered ';' (length ';' = 1)

varseniuk: Let's follow your code:
1. You split the string into an array, getting rid of the ';' sign. After that, you have an array of strings, which were early divided with ';'. No ';' left anymore.
2. Next step - you are mapping your array, creating an array of arrays, where there are two strings, which were early divided with ':'.

So now you have an array of a kind [['property', 'value'],['property', 'value'],....]

And you're trying to filter this array with length > 1. So actually what you are doing - is checking the length of the array ['property', 'value'].
There's no ';' sign anymore, you split it on your first step and you don't need to filter it. So that's why I'm wondering, why you using length > 1

I guess you want to filter only elements, which are pairs 'property-value', and get rid of redundant elements. And If so, it would be better to use length === 2, or some other checks. length > 1 is not understandable at all

varseniuk (changes requested):

varseniuk (approved):
It's ok. But can be better. 
Please, read my comment and take it to consideration

src/convertToObject.js
andrew-zaycev: > It's ok. But can be better.
> Please, read my comment and take it to consideration

Thank you for your review. As for the "length", I understood.
When I talked about ';' I meant this - https://prnt.sc/w8la5p.

PR #463, opened 2021-02-17
solution

antoniakstepan:


Ruslanwp (approved):
better use for of, and u can filter this array instead check (i === undefined)

Vladimir-Beliy (approved):

PR #465, opened 2021-02-17
solution

MorozOleh:


Ruslanwp (approved):
not simple way to solve it and understand for what reduce) reduce method looks really cool 😎

Vladimir-Beliy (approved):

PR #499, opened 2021-03-11
solution

Uanov:


danylo-onishchuk (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,22 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+  const sourceStringArray = sourceString.split('\n');
```
VinnieJ-2k20, line 16: Theoretically, styles can be written in one line, so split them by '\n' - is not the best idea.
But styles will always end with ';' so you should use it for splitting.

Additionally - this will allow you not to perform additional slice in the end.

VinnieJ-2k20 (changes requested):

VinnieJ-2k20 (approved):
👍 
Good job!

PR #500, opened 2021-03-11
solution

OlegHlukh:


haiduk2019 (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,22 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+  const splitSouresSting = sourceString.split('\n');
```
VinnieJ-2k20, line 16: Theoretically, styles can be written in one line, so splitting them by '\n' - is not the best idea.
But styles will always end with ';' so you should use it for splitting.

Additionally - this will allow you not to perform additional slice in the end.

src/convertToObject.js
```diff
@@ -12,7 +12,22 @@
...
+  const propertyList = [];
+
+  for (let i = 0; i < splitSouresSting.length; i++) {
+    propertyList.push(splitSouresSting[i].split(': '));
```
VinnieJ-2k20, line 20:
```suggestion
    propertyList.push(splitSouresSting[i].split(':'));
```

VinnieJ-2k20 (changes requested):
Please fix solution according to comments.

Denchepornyuk (approved):

PR #506, opened 2021-03-11
solution

InnaSoro:


danylo-onishchuk (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let array = sourceString.split(';');
+
+  const result = {};
+
+  array = array.map(item => (item.replace('"').trim()));
```
VinnieJ-2k20, line 19:
```suggestion
  array = array.map(item => (item.trim()));
```

This 'replace' does nothing in this case.

src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
...
+
+  while (array.indexOf('') !== -1) {
+    array.splice(array.indexOf(''), 1);
+  }
```
VinnieJ-2k20, line 23: Here you are overcomplicating things. As a result you will have three cycles inside one another.
I would suggest using filter to simply filter out empty array elements.

src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
...
+
+  array.map(item =>
+    (result[item.split(':')[0].trim()] = item.split(':')[1].trim()));
```
VinnieJ-2k20, line 26: Again, overcomplicated function calls. You have two same split calls.
Also, if you don't use the result of map method (new array) - it's logically better to use forEach
```suggestion
  array.forEach(item => {
    const [property, value] = item.split(':');

    result[property.trim()] = value.trim();
  });
```

VinnieJ-2k20 (changes requested):
Please, fix according to comments.
Also, try to use more meaningful variables names:
array -> propertiesArray
result -> stylesObject
etc

Denchepornyuk (approved):

PR #509, opened 2021-03-11
Solution

Julietta-M:


vladdezhniuk (approved):

VinnieJ-2k20 (approved):
Good job!

But try to use more meaningful variables names. Focus on what it is from business logic point of view, not from data types.
Bad names: string, array, object, result etc.
Better names: stylesObject, propertyAndValue, filteredProperties etc.

PR #510, opened 2021-03-11
Solution

svyat-krit:


OlenaMelnyk (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,25 @@
...
+  for (let i = 0; i < properties.length; i++) {
+    if (properties[i].trim().length === 0) {
+      continue;
+    }
+
+    pair = properties[i].split(':');
+    pair[0] = pair[0].trim();
+    pair[1] = pair[1].trim();
+
+    result[pair[0]] = pair[1];
```
VinnieJ-2k20, line 30: You can simplify this section with destructuring:
```suggestion
    const [property, value] = properties[i].split(':');

    result[property.trim()] = value.trim();
```

VinnieJ-2k20 (approved):
👍 
Good job!

For the future - always try to use meaningful names for variables:
result -> stylesObject
pair -> propertyValuePair

PR #532, opened 2021-04-26
Solution

Taras-Konovalov:


Burunduk003 (approved):
Nice simple solution. You could be more specific with namings. Names like 'delTrim' sound not ok to me.

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  const semicolon = sourceString.split(';');
+  const colon = semicolon.map(i => i.split(':'));
+  const delTrim = colon.map(del => {
+    return del.map(elem => {
+      return elem.trim();
+    });
+  });
```
mykola-yarchuk, line 22: you can use `trim` when you mapping semicolon

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
```
mykola-yarchuk, line 15: `obj` that's a bad name

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
...
+  const semicolon = sourceString.split(';');
+  const colon = semicolon.map(i => i.split(':'));
+  const delTrim = colon.map(del => {
```
mykola-yarchuk, line 18: delTrim it's hard to understand what it is

mykola-yarchuk (changes requested):
almost done

src/convertToObject.js
```diff
@@ -12,7 +12,22 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objResult = {};
```
mykola-yarchuk, line 15:
```suggestion
  const result = {};
```

mykola-yarchuk: don't use types in name

mykola-yarchuk (approved):

PR #541, opened 2021-04-29
develop

Andrey203:


nazarmatsevych (approved):
Nice solution 👍

src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
...
+  const stylesObj = {};
+
+  convertStylesToArr.forEach(el => {
+    const style = el.slice(0, el.lastIndexOf(';')).split(':');
+    const [key, value] = [style[0].trim(), style[1].trim()];
+
+    stylesObj[key] = value;
+  });
+
+  return stylesObj;
```
aliceblunt8, line 27: let's separate concerns here
fully prepare your array inside convertStylesToArr variable (split, trim, filter and etc), so then in forEach loop you will only need to assign values to an object

src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertStylesToArr = sourceString
+    .split('\n')
+    .filter(prop => prop.includes(':'));
+  const stylesObj = {};
```
aliceblunt8, line 18: move this variable declaration to the top of the function

aliceblunt8 (changes requested):

src/convertToObject.js
```diff
@@ -12,7 +12,24 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObj = {};
```
```suggestion
  const stylesObj = {};
  
```

src/convertToObject.js
```diff
@@ -12,7 +12,24 @@
...
+  const convertStylesToArr = sourceString
+    .split('\n')
```
aliceblunt8, line 17: maybe split by ';'? it's a separator for each rule

src/convertToObject.js
```diff
@@ -12,7 +12,24 @@
...
+    .filter(prop => prop.includes(':'))
+    .map(el => {
+      const style = el.slice(0, el.lastIndexOf(';')).split(':');
```
aliceblunt8, line 20: split by ';' and you won't need this lastIndexOf here

src/convertToObject.js
```diff
@@ -12,7 +12,24 @@
...
+
+      style[0] = style[0].trim();
+      style[1] = style[1].trim();
```
aliceblunt8, line 23: use map after split(':') for trimming each value in resulting array of prop and it's value

src/convertToObject.js
```diff
@@ -12,7 +12,24 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObj = {};
+  const convertStylesToArr = sourceString
```
aliceblunt8, line 16:
```suggestion
  const styles = sourceString
```

aliceblunt8 (changes requested):

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObj = {};
+
+  const styles = sourceString
+    .split(';')
+    .filter(prop => prop.includes(':'))
+    .map(el => el.split(':').map(prop => prop.trim()));
```
aliceblunt8, line 20:
```suggestion
  const styles = sourceString
    .split(';')
    .filter(prop => prop.includes(':'))
    .map(el => el
      .split(':')
      .map(prop => prop.trim())
    );
```

aliceblunt8 (approved):

PR #547, opened 2021-04-30
first solution

scarlettjohanssons:


IvanKolodiazhnyi (commented):
Nice!!!

VinnieJ-2k20 (changes requested):
This solution works, but it is not the most reliable one in case of irregulars in `sourceString`.

VinnieJ-2k20 (approved):

PR #548, opened 2021-04-30
solution

TimurHrynenko:


src/convertToObject.js
```diff
@@ -12,7 +12,23 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let cssRulesObject;
+  const cssRulesArr = sourceString.split(';');
+
+  cssRulesArr.map(rule => {
+    if (rule.trim()) {
+      const property = rule.split(':')[0].trim();
+      const value = rule.split(':')[1].trim();
```
aliceblunt8, line 21: mb before assigning values to object first split and trim each property in array?
you can do this inside cssRulesArr variable assignment

aliceblunt8 (changes requested):

VinnieJ-2k20 (approved):
Good job 👍

PR #549, opened 2021-05-01
Solution

Nazar-Kudatsky:


src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  const strArr = sourceString.split(':').join(';').split(';')
```
illisogori, line 16:
```suggestion
  const properties = sourceString.split(':').join(';').split(';')
```

illisogori (approved):
Good job.

VinnieJ-2k20 (changes requested):
This solution works, but it is not the most reliable one in case of irregulars in `sourceString.`

Also, try to use more meaningful variable names.

VinnieJ-2k20 (approved):
Good job 👍

PR #550, opened 2021-05-01
solution

victor-arutiunov:


Dima-Batiuk (commented):
Мені здається , краще використовувати методи масивів `map , filter`  замість  ` if ` )

src/convertToObject.js
```diff
@@ -12,7 +12,27 @@
...
+  const stylesList = sourceString.split('\n');
+
+  const stylesObj = {};
+
+  stylesList.forEach((style, ind, arr) => {
+    arr[ind] = style.split(':');
+
+    if (arr[ind].length > 1) {
+      arr[ind][0] = arr[ind][0].trim();
+      // console.log(arr[ind][1]);
```
aliceblunt8, line 24: delete comments

src/convertToObject.js
```diff
@@ -12,7 +12,27 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesList = sourceString.split('\n');
```
aliceblunt8, line 15: maybe separate by ';'?

src/convertToObject.js
```diff
@@ -12,7 +12,27 @@
...
+
+    if (arr[ind].length > 1) {
+      arr[ind][0] = arr[ind][0].trim();
+      // console.log(arr[ind][1]);
+      arr[ind][1] = arr[ind][1].slice(0, arr[ind][1].length - 1).trim();
+    } else {
+      arr[ind][0] = arr[ind][0].trim();
+    }
+    stylesObj[arr[ind][0]] = arr[ind][1];
+  });
```
aliceblunt8, line 30: let's separate concerns here
first, prepare an array like this (you can use array iteration methods here instead of double indexes)
`[ [ background-color, #fff] , [-webkit-border-radius, 5px] ]`

and then in a separate loop create an object out of it

aliceblunt8 (changes requested):

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
...
+    .split(';')
+    .forEach((style, ind, arr) => {
+      const property = style
+        .split(':')
+        .map(elem => elem.trim());
+
+      stylesObj[property[0]] = property[1];
+    });
+
+  delete stylesObj[''];
```
aliceblunt8, line 27: check property value before assigning it yo object, so you won't need to delete it later

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #553, opened 2021-05-03
created css obj

YuriiLiso:


vadym-liubych (approved):
Good job!

src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
...
+    .map(property => property
+      .split(':').map(styleValue => styleValue.trim())
+    )
+    .filter(property => property.length > 1);
+
+  const cssPropertyesObj = {};
+
+  clearArr.map(prop => {
+    cssPropertyesObj[prop[0]] = prop[1];
+  });
```
VinnieJ-2k20, line 26: You don't use map results anywhere, so logically it would be better to use general loop or `forEach` here.

VinnieJ-2k20 (changes requested):
Almost done. Fix one detail.

VinnieJ-2k20 (approved):

PR #556, opened 2021-05-03
solution

RuslanaPodshyvalova:


aionova (changes requested):
Nice, but let's try to refactor to make it shorter?

aionova (approved):
Great job!

PR #561, opened 2021-05-04
Solution

IvanVasylenko:


src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const properties = new Set(sourceString.split(';'));
+
+  properties.delete('\n');
+  properties.delete('\n  ');
+  properties.delete('\n\n  ');
```
aionova, line 19: Can we refactor this?

aionova (changes requested):
Great, but can be refactored

PR #562, opened 2021-05-04
Style to object

rauzok:


Burunduk003 (approved):
Good solution!

VinnieJ-2k20 (changes requested):
This solution works, but it is not the most reliable one in case of irregulars in `sourceString.`

VinnieJ-2k20 (approved):

PR #567, opened 2021-05-04
solution

amithos:


Namart (approved):
good job

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const css = {};
```
```suggestion
  const stylesObject = {};
```

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
+  const rules = sourceString.split(';\n');
```
aliceblunt8, line 16: why don't just split by ';'? it does the same work

aliceblunt8 (approved):

PR #571, opened 2021-05-04
Solution

BoikoEva:


src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayOfStyles = sourceString.split(':').join(';').split(';').map(item =>
```
illisogori, line 15:
```suggestion
  const styles= sourceString.split(':').join(';').split(';').map(item =>
```

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
+    item.trim());
+  const css = {};
```
illisogori, line 17:
```suggestion
  const styleObject = {};
```

illisogori (approved):
Godd job.

src/convertToObject.js
aliceblunt8, line 15: what's going on here? you join by ; to split again by ;?
maybe just split by ; at the beginning?

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
+
+  for (let i = 0; i < arrayOfStyles.length; i++) {
+    if (arrayOfStyles[i] !== '') {
+      css[arrayOfStyles[i]] = arrayOfStyles[i + 1];
+      i++;
+    }
```
aliceblunt8, line 23: let's create array of arrays in the first place, so you don't need to use indexes that much

let's try to create such array out of the initial string
`[[property, value], [property, value], etc]`

aliceblunt8 (changes requested):

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString.split(';').map(s => s.split(':'))
+    .map(s => s.map(x => x.trim()))
+    .filter(x => x.length === 2);
```
aliceblunt8, line 17:
```suggestion
  const styles = sourceString
  .split(';')
  .map(item => item
    .split(':')
    .map(prop => prop.trim()))
    .filter(element => element.length === 2);
```

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
...
+  const styleObject = {};
+
+  for (const style of styles) {
+    styleObject[style[0]] = style[1];
+  }
```
aliceblunt8, line 22: use forEach here
destructure key and value from each array

aliceblunt8 (changes requested):

src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
...
+    .split(';')
+    .map(item => item
+      .split(':')
+      .map(prop => prop.trim()))
+    .filter(element => element.length === 2);
+  const styleObject = {};
+
+  styles.forEach(element => {
+    styleObject[element[0]] = element[1];
+  });
```
aliceblunt8, line 25:
```suggestion
  styles.forEach(element => {
   const [key, value ] = element;
  
    styleObject[key] = value;
  });
```

src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .map(item => item
+      .split(':')
+      .map(prop => prop.trim()))
+    .filter(element => element.length === 2);
+  const styleObject = {};
```
aliceblunt8, line 21: move this variable to the top of the function

aliceblunt8 (changes requested):

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
```
```suggestion
  const styleObject = {};
  
```

aliceblunt8 (approved):
GJ

PR #572, opened 2021-05-05
Solution

vova-myronenko:


VinnieJ-2k20 (approved):
Good solution 👍 

For the future - try to use more meaningful variable names.
`arr` basically says nothing from the business logic point of view.

PR #575, opened 2021-05-05
solution

stas-dukach:


src/convertToObject.js
```diff
@@ -13,6 +13,18 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const stringToArr = sourceString.split(':').join(';').split(';')
```
aliceblunt8, line 16: omg what's happening here? maybe first split by ';'?

src/convertToObject.js
```diff
@@ -13,6 +13,18 @@
...
+    .map(item => item.trim());
+  const cssRules = {};
```
aliceblunt8, line 18: move to the top of the function

src/convertToObject.js
```diff
@@ -13,6 +13,18 @@
...
+
+  for (let i = 0; i < stringToArr.length; i++) {
+    if (stringToArr[i] !== '') {
+      cssRules[stringToArr[i]] = stringToArr[i + 1];
+      i++;
+    }
```
aliceblunt8, line 24: let's create array of arrays in the first place, so you don't need to use indexes that much

let's try to create such array out of the initial string
[[property, value], [property, value], etc]

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #578, opened 2021-05-05
Solution

vitaliy-buchinchik:


src/convertToObject.js
```diff
@@ -12,7 +12,13 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return Object.fromEntries(
```
victor-arutiunov, line 15: Wow! Very nice solution. Thanks for idea)

vitaliy-buchinchik: Thank U for the review)

aliceblunt8 (approved):

PR #583, opened 2021-05-05
js_style-to-object

Bogdan-Kotsupey:


src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const trimed = sourceString.split(';').map(element => element.trim());
+  const noEmpty = trimed.filter(word => word !== '');
+  const pairs = noEmpty.map(element => element.split(':'));
```
VinnieJ-2k20, line 18: You could do this using one variable and chained methods.
```suggestion
  const pairs = sourceString
    .split(';')
    .map(element => element.trim());
    .filter(word => word !== '');
    .map(element => element.split(':'));
```

VinnieJ-2k20 (approved):
Good job 👍 

Check the comment for some advice.

PR #585, opened 2021-05-05
Solution

MinchukO:


src/convertToObject.js
```diff
@@ -13,6 +13,14 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+
+  return sourceString.split(';')
+    .map(each => each.split(':'))
```
VinnieJ-2k20, line 18: `each` is not the best name for a variable.

VinnieJ-2k20 (approved):
Good solution 👍 

Check the comment - try to use more meaningfule variable names.

PR #675, opened 2021-08-06
Solution

alex-glotov67:


src/convertToObject.js
```diff
@@ -13,6 +13,17 @@
...
+  return sourceString
+    .split(';')
+    .filter(field => field.trim().length > 2)
+    .map(keyValue => keyValue.split(':').map(prop => prop.trim()))
+    .reduce((object, property) => {
+      const keyValue = {};
+
+      keyValue[property[0]] = property[1];
+
+      return Object.assign(object, keyValue);
```
illiarerun, line 25: FYI: You can make new object right in assign method. Of course it's better to do it when object isn't big and complex.

illiarerun (approved):
Great job 🔥

src/convertToObject.js
alex-glotov67: Good point! Thanks)

Denchepornyuk (approved):

PR #678, opened 2021-08-08
done

SashaLifashkin:


SashaRudenko (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const arrayFromString = sourceString.split('\n').filter(
+    item => item).map(
+    item => item.split(':')).filter(
+    item => item.length > 1);
```
herkoss, line 18:
```suggestion
  const arrayFromString = sourceString.split('\n')
    .filter(item => item)
    .map(item => item.split(':'))
    .filter(item => item.length > 1);
```

src/convertToObject.js
```diff
@@ -12,7 +12,17 @@
...
+
+  const resultObject = {};
 
+  arrayFromString.forEach(item => {
+    resultObject[item[0].trim()] = item[1].slice(0, -1).trim();
```
herkoss, line 23: Too much operations in one row. Let's refactor it to make your code more clear

herkoss (changes requested):
Good solution!👍👌
But it needs some refactoring
Details in comments)

herkoss (approved):
Well done!👍✅

PR #679, opened 2021-08-08
solution

SashaRudenko:


ivanchenko-kirill (approved):
Все круто 👍

src/convertToObject.js
```diff
@@ -13,6 +13,22 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const strToArr = sourceString.split(';').filter(item => item.length > 4);
```
VinnieJ-2k20, line 16: Why 4? Use more obvious condition.
Think about how to get rid of unnecessary spaces.

src/convertToObject.js
```diff
@@ -13,6 +13,22 @@
...
+
+  const formattedArr = strToArr
+    .map(element => element.split(':')
+      .map(strItem => strItem.trim()));
```
VinnieJ-2k20, line 20: Don't use shortened words in var names.

```suggestion
  const formattedArray = stringToArray
    .map(element => element
      .split(':')
      .map(strItem => strItem.trim());
    );
```

VinnieJ-2k20 (changes requested):
You need to adjust logic a little bit.
Your code must work not only with test cases.

src/convertToObject.js
```diff
@@ -13,6 +13,24 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const strToArray = sourceString.split(';');
```
VinnieJ-2k20, line 16:
```suggestion
  const stringToArray = sourceString.split(';');
```

src/convertToObject.js
```diff
@@ -13,6 +13,24 @@
...
+
+  const formattedArr = strToArray
```
VinnieJ-2k20, line 18:
```suggestion
  const formattedArray = stringToArray
```

src/convertToObject.js
```diff
@@ -13,6 +13,24 @@
...
+    .map(element => element
+      .split(':')
+      .map(strItem => strItem.trim()))
+    .filter(item => item.length > 1);
+
+  const arrToObj = formattedArr
```
VinnieJ-2k20, line 24:
```suggestion
  const arrayToObject = formattedArray
```

VinnieJ-2k20 (changes requested):

VinnieJ-2k20 (approved):

PR #681, opened 2021-08-09
solution

VeronikaLushchik:


alex-glotov67 (approved):
Very nice, but you can do this with one chain of methods and without any extra variables.
like 
`return sourseString.split(';').map()x=> x.split(';')...` end go on!

PR #683, opened 2021-08-10
Solution

denyschemeriakov:


alex-glotov67 (approved):
Very nice, but you can do this with one chain of methods and without any extra variables.
like
return `sourseString.split(';').map()x=> x.split(';')... ` end go on!
even without cycle for!

denyschemeriakov:
> Very nice, but you can do this with one chain of methods and without any extra variables.
> like
> return `sourseString.split(';').map()x=> x.split(';')... ` end go on!
> even without cycle for!

Cool solution! Thanks!)

Denchepornyuk (approved):

PR #684, opened 2021-08-10
Develop

ihor-jpeg:


nastyyvolk (approved):
Nice! There's nothing to suggest

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .map(item => item.split(':')
+      .map(line => line.trim()))
```
herkoss, line 18:
```suggestion
    .map(line => line.trim()))
```

herkoss (approved):
Well done!💪👌

src/convertToObject.js
ihor-jpeg: that map is called on item.split(':'). should we not indent it in this case?

PR #688, opened 2021-08-11
Solution

lera-yaskevych:
I solved it using reduce method, however I found there's a simpler way of doing the last two steps of the function. It seems to me, that these two lines of code make it look more straightforward.

`const filteredStylesArray = stylesArray.filter(entry => entry.length > 1);
 const stylesObject = Object.fromEntries(filteredStylesArray);`

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -17,9 +17,13 @@ function convertToObject(sourceString) {
     .map(entries => entries.split(':')
       .map(property => property.trim()));
 
+  const callback = (prev, [key, value]) => {
+    return Object.assign(prev, { [key]: value });
+  };
+
   const stylesObject = stylesArray
     .filter(entry => entry.length > 1)
-    .reduce((prev, [key, value]) => Object.assign(prev, { [key]: value }), {},);
+    .reduce(callback, {});
 
   return stylesObject;
 }
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -12,7 +12,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString
+    .split(';')
+    .map(entries => entries.split(':')
+      .map(property => property.trim()));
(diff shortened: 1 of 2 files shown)


src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
...
+    .map(entries => entries.split(':')
+      .map(property => property.trim()));
+
+  const callback = (prev, [key, value]) => {
+    return Object.assign(prev, { [key]: value });
+  };
+
+  const stylesObject = stylesArray
+    .filter(entry => entry.length > 1)
+    .reduce(callback, {});
```
lera-yaskevych, line 26:
```suggestion
  const filteredStylesArray = stylesArray.filter(entry => entry.length > 1);
  const stylesObject = Object.fromEntries(filteredStylesArray);
```

ihor-jpeg: brilliant use of `Object.fromEntries`, saves from pain of writing a reduce callback function. 
but i don't think you need `stylesObject` here. just immediately return `Object.fromEntries(filteredStylesArray)`

Another thing could be chaining the filter method to `stylesArray`

src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString
+    .split(';')
+    .map(entries => entries.split(':')
+      .map(property => property.trim()));
+
+  const callback = (prev, [key, value]) => {
+    return Object.assign(prev, { [key]: value });
+  };
```
ihor-jpeg, line 22: I guess you don't need this anymore

ihor-jpeg (approved):
good code! i left my two cents below;

VinnieJ-2k20 (approved):
I agree with Ihor, solution with `Object.fromEntries(...) ` is a very clever idea. 👍 
Just keep in mind that all data should be prepared in appropriate way for it to work.

PR #692, opened 2021-08-12
solution

Vova-Stoliar:


DmytroMykoliv (approved):
Good work ))

src/convertToObject.js
```diff
@@ -13,6 +13,19 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const sourceStringToArray = sourceString
+    .split('\n')
```
VinnieJ-2k20, line 17: It is a bad idea to split by newline here. Test don't cover it here, but styles can all be written in one line.
More sustainable solution is to split by `;`

VinnieJ-2k20 (changes requested):
Solution looks good, but you need to fix one important detail.

VinnieJ-2k20 (approved):

PR #703, opened 2021-08-13
Style to Object

andrii-hryhoruk:


MaksymBeliever (approved):
Цікавий варіант, мені сподобався 👍

src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const uniqueProperties = new Set(sourceString.split(';'));
+
+  uniqueProperties.delete('\n');
+  uniqueProperties.delete('\n  ');
+  uniqueProperties.delete('\n\n  ');
```
VinnieJ-2k20, line 19: It may seem like easy solution, but what if you'll have more non-valid elements?
You need to find a different way to get rid of useless spaces and new lines.

VinnieJ-2k20 (changes requested):
You need to restructure your solution a little bit.

varseniuk (approved):

PR #705, opened 2021-08-13
js_style-to-object

RomanFedusevych:


src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const formatString = sourceString
+    .split(';')
+    .map(property => property.trim().split(':'))
+    .filter(property => property.length === 2);
+
+  const cssObj = {};
+
+  for (const key of formatString) {
```
Clonvhs, line 22: can use ForEach

Clonvhs (approved):
good one

src/convertToObject.js
VinnieJ-2k20: Yep, we need to learn how to use array methods instead of loops. Rewrite using `forEach`

VinnieJ-2k20 (changes requested):
Almost done, fix just one detail.

VinnieJ-2k20 (approved):
Good job 👍

PR #707, opened 2021-08-13
Solution

Illia-Boiko:


LazurNiko (approved):
nice solution)

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayWithStyles = sourceString.split(';')
+    .map(properties => properties.split(':')
+      .map(element => element.trim()))
+    .filter(x => x.length > 1);
```
VinnieJ-2k20, line 18:
```suggestion
  const arrayWithStyles = sourceString
    .split(';')
    .map(properties => properties.split(':')
      .map(element => element.trim()))
    .filter(x => x.length > 1);
```

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
...
+
+  return arrayWithStyles.reduce((object, value) =>
+    ({
+      ...object,
+      [value[0]]: value[1],
+    }), 0);
```
VinnieJ-2k20, line 24: Correct starting value here will be empty object `{}`.

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment about `reduce` method.

PR #710, opened 2021-08-14
solution

derelicttt:


Clonvhs (approved):
Good one

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
+  const properties = sourceString
+    .split(';')
+    .map(property => property.split(':'))
+    .filter(property => property.length === 2);
+  const addProperty = (styles, [key, value]) => {
+    return {
+      ...styles,
+      [key.trim()]: value.trim(),
+    };
+  };
```
VinnieJ-2k20, line 24: You don't use this callback anywhere else, therefore it is better to write it inside reduce call.

VinnieJ-2k20 (changes requested):
Almost done, just fix one code style detail.

varseniuk (approved):

PR #713, opened 2021-08-15
solution

LazurNiko:


RomanFedusevych (approved):
good solution👍🏻

src/convertToObject.js
```diff
@@ -12,7 +12,14 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const string = sourceString.split(';');
+  const strTrim = string.filter(empty => empty.trim().length);
+  const arrObj = strTrim.map(key => key.split(':').map(value => value.trim()));
```
VinnieJ-2k20, line 17: No need to create separate variables here for each step.
Use chained methods.

src/convertToObject.js
```diff
@@ -12,7 +12,14 @@
...
+
+  return arrObj.reduce((obj, value) => ({
+    ...obj,
+    [value[0]]: value[1],
+  }), 0);
```
VinnieJ-2k20, line 22: Correct starting value here will be empty object `{}`.

VinnieJ-2k20 (changes requested):
Almost done, just need to fix some details.

src/convertToObject.js
```diff
@@ -12,7 +12,16 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  return sourceString
+    .split(';')
+    .filter(empty => empty.trim().length)
+    .map(key => key.split(':').map(value => value.trim()))
+
+    .reduce((obj, value) => {
```
VinnieJ-2k20, line 20:
```suggestion
    .map(key => key.split(':').map(value => value.trim()))
    .reduce((obj, value) => {
```

VinnieJ-2k20 (approved):
Good job 👍

PR #718, opened 2021-08-16
Solution

Clonvhs:
v1.0

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -1,18 +1,55 @@
 'use strict';
 
-/**
- * Implement convertToObject function:
- *
- * Function takes string with styles
- * and returns object where CSS properties are keys
- * and values are the values of related CSS properties
-
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedObj = {};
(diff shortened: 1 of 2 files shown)


Andrii-Kalaida (commented):
Гарна робота)

src/convertToObject.js
```diff
@@ -1,18 +1,16 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedObj = {};
+
+  sourceString.split(';').map(stroke =>
+    stroke.split(':').map(element => element.trim()))
+    .filter(array => array.length > 1)
+    .forEach(array => {
+      convertedObj[array[0]] = array[1];
+    });
```
VinnieJ-2k20, line 11: Pay attention to indentations. Each chained method call should be on the same level.
```suggestion
  sourceString
    .split(';')
    .map(stroke =>
      stroke.split(':').map(element => element.trim()))
    .filter(array => array.length > 1)
    .forEach(array => {
      convertedObj[array[0]] = array[1];
    });
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

src/convertToObject.js
Clonvhs: got it, in future will no make these mistakes

PR #721, opened 2021-08-16
Solution

Ingvvvar:


Andrii-Kalaida (commented):
Чудова робота!)

src/convertToObject.js
```diff
@@ -11,8 +11,11 @@
...
  */
-function convertToObject(sourceString) {
-  // write your code here
-}
+
+const convertToObject = sourceString => Object.fromEntries(sourceString
+  .split(';')
+  .map(element => element.split(':')
+    .map(property => property.trim()))
+  .filter(item => item.length > 1));
```
VinnieJ-2k20, line 19: Good use of Object.fromEntries 👍 
Just watch the code style.
```suggestion
const convertToObject = sourceString => Object
  .fromEntries(sourceString
    .split(';')
    .map(element => element.split(':')
      .map(property => property.trim()))
    .filter(item => item.length > 1));
```

VinnieJ-2k20 (approved):
Good job 👍

PR #754, opened 2021-09-22
стайл ту обджект

yungvitik:


caelum29 (approved):
👍

src/convertToObject.js
```diff
@@ -12,7 +12,17 @@
...
+  const result = {};
+  const stylesPairs = sourceString.split(';').map(pair => pair.split(':'));
+
+  stylesPairs.forEach(convert);
 
+  function convert(pair) {
+    if (pair.length === 2) {
+      result[pair[0].trim()] = pair[1].trim();
+    }
+  }
```
aliceblunt8, line 24: no need to create an additional function convert, do everything inside forEach

src/convertToObject.js
```diff
@@ -12,7 +12,17 @@
...
-}
+  const result = {};
+  const stylesPairs = sourceString.split(';').map(pair => pair.split(':'));
+
+  stylesPairs.forEach(convert);
 
+  function convert(pair) {
+    if (pair.length === 2) {
+      result[pair[0].trim()] = pair[1].trim();
+    }
```
aliceblunt8, line 23:
```suggestion
    if (pair.length === 2) {
    const [key, value ] = pair;
      result[key.trim()] = value.trim();
    }
```

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #757, opened 2021-09-23
solution

caelum29:


AVronskyi (approved):
look like good!

src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
...
+  return arrOfLines.reduce((result, line) => {
+    let [key, value] = line.split(':');
+
+    key = key.trim();
+    value = value.trim();
+
+    return {
+      ...result,
+      [key]: value,
+    };
```
aliceblunt8, line 27:
```suggestion
    return {
      ...result,
      [key.trim()]: value.trim(),
    };
```

src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const regex = /(\S.*?:)(\s.*\b)/g;
```
aliceblunt8, line 15: why do you need this regex here?

aliceblunt8 (changes requested):

Ihor-Karpyn (approved):

PR #787, opened 2021-09-29
Develop

IshchenkoAnton:


src/convertToObject.js
```diff
@@ -11,8 +11,23 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const splited = sourceString.split(';');
+  const obj = {};
```
Anna-Tkachenko, line 17: Create more meaningful variable name instead of obj

Anna-Tkachenko (approved):
Good job!👍

PR #792, opened 2021-10-14
Task solved

cyb1ove:


src/convertToObject.js
```diff
@@ -13,6 +13,15 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  return sourceString.split(';').reduce((obj, codeline) => {
```
Olexandr6, line 16: I would rename variable `codeline` to `cssProperty`

src/convertToObject.js
```diff
@@ -13,6 +13,15 @@
...
+    const propertyAndValue = codeline.split(':');
```
Olexandr6, line 17: I would create two variables for property key and value.
```suggestion
    const [key, value]= cssProperty.split(':');
```

G-YaSha (approved):
Good solution🔥

src/convertToObject.js
cyb1ove: Thanks for suggestion!

src/convertToObject.js
```diff
@@ -13,6 +13,15 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  return sourceString.split(';').reduce((obj, cssProperty) => {
```
solaryasha, line 16:
```suggestion
  return sourceString.split(';').reduce((сssProperties, cssProperty) => {
```

solaryasha: Уникай назв змінних, типу `obj`, `arr`, `array`, `object`

solaryasha (approved):

PR #797, opened 2021-10-15
solution

podlesnyi-pavel:


src/convertToObject.js
```diff
@@ -1,18 +1,20 @@
...
-
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrStyles = sourceString.split(';').map(style => {
+    return style.trim();
+  }).filter(style => style.length > 0);
```
VinnieJ-2k20, line 6: You can write it simpler.
Also, if we use chained methods, each call should be on the same level.
```suggestion
  const arrStyles = sourceString
    .split(';')
    .map(style => style.trim())
    .filter(style => style.length > 0);
```

src/convertToObject.js
```diff
@@ -1,18 +1,20 @@
...
+
+  const result = arrStyles.reduce((prev, current) => {
```
VinnieJ-2k20, line 8: You can also call this `reduce` in cahin with previous methods.
And, you create result variable and return it right away.

VinnieJ-2k20 (changes requested):
You code can still be improved.

PR #798, opened 2021-10-15
solved

arielg0rl:


src/stylesString.js
```diff
@@ -2,7 +2,7 @@
 
 const stylesString = `
   background-color:      #fff;
--webkit-border-radius: 5px;
+  -webkit-border-radius: 5px;
```
cyb1ove, line 5: Source file changing is not right. You should pass test with original source file.

src/convertToObject.js
```diff
@@ -12,7 +12,33 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let sourceArray = [];
+
+  sourceArray = sourceString.split(';');
+
+  let trimmedSourceArray = [...sourceArray];
```
VinnieJ-2k20, line 19: You created this array. Don't need to make copy before each operation.
Moreover, methods like `filter` or `map` don't mutate original array.

src/convertToObject.js
```diff
@@ -12,7 +12,13 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resultObject = {};
+
+  sourceString.split(';').map(line => line.trim()).filter(x => x.length > 0)
+    .map(property => property.split(':').map(x => x.trim()))
+    .map(element => {resultObject[element[0]] = element[1]});
```
VinnieJ-2k20, line 19: 1. You don't write anything to a variable, so last `map` method should be replaced with `forEach`.
2. All chained methods called on same level should have same indentation.

```suggestion
  sourceString
    .split(';')
    .map(line => line.trim())
    .filter(x => x.length > 0)
    .map(property => property.split(':').map(x => x.trim()))
    .forEach(element => {resultObject[element[0]] = element[1]});
```

VinnieJ-2k20 (changes requested):
Almost done.
Just need to improve your code.

PR #801, opened 2021-10-17
[JS] Styles-to-Object STRZH

DionStrzh:


src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splittedSourceString = sourceString.split(';');
+
+  return splittedSourceString.reduce((acc, el) => {
+    const [key, values] = el.trim().split(':');
```
cyb1ove, line 18:
```suggestion
    const [key, values] = el.split(':');
```
You trim the same elements in several places. This `trim` is extra.

cyb1ove (changes requested):
Need some fix.

PR #804, opened 2021-10-18
solution

denys-anosov:


src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
...
+
+  sourceString.split(';')
+    .map(element => element.trim())
+    .filter(element => element !== '')
+    .map(element => element.split(':'))
+    .map(element => element.map(subElement => subElement.trim()))
+    .map(function(element) {
+      output[element[0]] = element[1];
+
+      return element;
```
HaqpaHR, line 25: просто интересно, чего одной строкой не записал?

VinnieJ-2k20: You don't write anything from this piece of code to a variable, so last map should be safely replaced with `forEach`

VinnieJ-2k20 (changes requested):
Almost done. Need to fix one array method.

VinnieJ-2k20 (approved):
Good job 👍

PR #805, opened 2021-10-18
solutions

Linetstg:


src/convertToObject.js
```diff
@@ -12,7 +12,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const filteredArray = sourceString.split(';').map(el => el
+    .trim()).filter(x => x.length > 0).map(property =>
+    property.split(':').map(x => x.trim()));
```
VinnieJ-2k20, line 17: All methods called on one level should have same indentation.
```suggestion
  const filteredArray = sourceString
    .split(';')
    .map(el => el.trim())
    .filter(x => x.length > 0)
    .map(property => property.split(':').map(x => x.trim()));
```

VinnieJ-2k20 (approved):
I'll approve, but try to make your code more readable.

PR #807, opened 2021-10-18
solution

HaqpaHR:


src/convertToObject.js
```diff
@@ -13,6 +13,12 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const cleanArray = sourceString.split(';')
+    .map((element) => element.trim())
+    .filter((element) => element.length > 0)
+    .map((element) => element.split(':').map(el => el.trim()));
```
VinnieJ-2k20, line 19: All chained methods called on the same level should have same indentation. This makes code much more readable.
```suggestion
  const cleanArray = sourceString
    .split(';')
    .map((element) => element.trim())
    .filter((element) => element.length > 0)
    .map((element) => element.split(':').map(el => el.trim()));
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

PR #808, opened 2021-10-18
Solution

DarinaSavytska:


src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
...
+  sourceString.split(';')
+    .map(element => element.trim())
+    .filter(element => element !== '')
+    .map(element => element.split(':'))
+    .map(element => element.map(subEl => subEl.trim()))
+    .map(function(element) {
+      result[element[0]] = element[1];
+
+      return element;
+    });
```
VinnieJ-2k20, line 26: You don't write anything from this piece of code to a variable, so last map should be safely replaced with `forEach`

VinnieJ-2k20 (changes requested):
Almost done. Need to fix one thing.

PR #809, opened 2021-10-18
done

DammneD:


src/convertToObject.js
```diff
@@ -13,6 +13,21 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+
+  const sourceStringToArray = sourceString.split(';');
+
+  const objectOfProperties = {};
+
+  sourceStringToArray.forEach(property => {
```
cyb1ove, line 21: I think this task needs to make with iteration methods.

cyb1ove (changes requested):

cyb1ove (approved):

PR #825, opened 2021-11-15
Object with styles completed

Vitaliy-0:


src/convertToObject.js
```diff
@@ -12,7 +12,28 @@
...
+    normalized = normalized.replace(/\n/g, '');
+  }
+
+  normalized = normalized.split(';');
+
+  for (let i = 0; i < normalized.length; i++) {
+    if (normalized[i].trim()) {
+      const index = normalized[i].indexOf(':');
+      const key = normalized[i].slice(0, index);
+      const value = normalized[i].slice(index + 1);
```
alex-kostomarov, line 30: normalized[i] used for many times and variable could be used instead of it

alex-kostomarov (approved):
Good job!

src/convertToObject.js
```diff
@@ -12,7 +12,28 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+  let normalized = sourceString;
+
+  while (normalized.includes('  ')
+  || normalized.includes('\n')) {
+    normalized = normalized.replace(/\s\s/g, '');
+    normalized = normalized.replace(/\n/g, '');
+  }
```
ihor-jpeg, line 22: let's think about time complexity. the `includes` method goes over the entire array (worst case).
the `replace` method does the same. it goes over the entire array. 
just in this while loop you go over the same array 4 times. i don't think it is necessary. 
think about time complexity and how you can do the same thing doing less

ihor-jpeg (changes requested):
Think about ways to optimise your code

ihor-jpeg (approved):
Well done! This is a much more optimised solution!

PR #826, opened 2021-11-15
Solution

alex-kostomarov:


Vitaliy-0 (approved):
Nice code, interesting way to solve this task

src/convertToObject.js
```diff
@@ -12,7 +12,24 @@
...
-  // write your code here
+  const css = {};
+
+  const propertiesAndValues = [];
+
+  sourceString
+    .split(';')
+    .forEach(prop => propertiesAndValues.push(prop.split(':')));
+
+  for (const property of propertiesAndValues) {
```
aionova, line 23: let's try to do it with an array iteration method instead of a loop

aionova (changes requested):

aionova (approved):
reduce would be better here instead of forEach ;)

PR #828, opened 2021-11-15
Solution

SerhiiKirik:
**I will do this later with reduce :D**

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -12,7 +12,26 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.split(';').filter((el) =>
+    el !== '' || el !== ' ');
+  const result = {};
+  const newArr = [];
+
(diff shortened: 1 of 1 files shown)


MagicBox15 (commented):
good!

MagicBox15 (approved):
clever 👍

src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const styles = sourceString.split(';');
+
+  return styles.reduce((prev, el) => {
+    if (el.trim() !== '' && el.trim() !== ' ') {
```
aionova, line 18: do you really need this check?` el.trim() !== '   '`

aionova (approved):
Nice!

src/convertToObject.js
SerhiiKirik: Oh, I think no need, thank you)

aionova (approved):

PR #830, opened 2021-11-15
Solution

kurtkobein:


JamesSunderlanddd (approved):
Great and intelligent way to solve the task, well done!

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+  const strArr = sourceString.trim().replace(/\s+/g, ' ').split(';');
+
+  strArr.forEach(string => {
```
aionova, line 18: you could also try to do it with reduce, where you pass an empty object as an initial value of the accumulator.

aionova (approved):

aionova (approved):

PR #831, opened 2021-11-15
solution was added

JamesSunderlanddd:


kurtkobein (approved):
Good job

src/convertToObject.js
```diff
@@ -12,7 +12,24 @@
...
+  const arrValues = [];
+  const resultObject = {};
+
+  for (let i = 0; i < filterSrc.length; i++) {
+    arrValues.push(filterSrc[i]
+      .slice(filterSrc[i].indexOf(':') + 1, filterSrc[i].indexOf(';')).trim());
+    arrKeys.push(filterSrc[i].slice(0, filterSrc[i].indexOf(':')).trim());
+
+    resultObject[arrKeys[i]] = arrValues[i];
+  }
```
aionova, line 30: let's try to do it with array iteration methods instead of the loop. For example, with reduce

aionova (changes requested):

src/convertToObject.js
JamesSunderlanddd: @aionova done)

aionova (changes requested):
.

aionova (approved):
Nice!

src/convertToObject.js
aionova: that was fast)

PR #835, opened 2021-11-16
css styles

DudarDaryna:


Val-space (approved):
I like your code! I didn't come up with the solution to use reduce to create an object

PR #836, opened 2021-11-17
Develop

Val-space:


src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+  const parseSourceString = sourceString.split(';');
+
+  for (const style of parseSourceString) {
+    const parseStyle = style.split(':');
```
DudarDaryna, line 19: Interesting way

DudarDaryna (approved):

aionova (changes requested):
GJ! let's try to do it with array methods now. For example, reduce

aionova (approved):
Nice!

PR #837, opened 2021-11-17
Solution

4qwerty:


src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const splitedString = sourceString.split(';');
```
WorkAccCom, line 16: this constant actually contains an array, so, I think there is should be a more appropriate constant name.

WorkAccCom (approved):
Good

aionova (approved):
Nice!

PR #840, opened 2021-11-17
Solution

Vitalii-Lytvyn01:
Not ideal solution , since i still end up with one empty key in object.

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -12,7 +12,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.trim().split(';');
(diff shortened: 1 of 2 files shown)


oleksiy-pochkayev (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.trim().split(';');
+  const cssObject = {};
+
```
aionova, line 17: You can simply check if element.trim() is an empty string before splitting it.

aionova (approved):

PR #842, opened 2021-11-17
solution

oleksiy-pochkayev:


msaikun (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,17 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+
+  for (const style of sourceString.split(';')) {
+    const [key, value] = style.split(':');
+
+    if (key.trim() !== '' && value.trim() !== '') {
+      styles[key.trim()] = value.trim();
```
ihor-jpeg, line 21: you are using the same command twice (`key.trim()`). At this point you can create a variable for that. the same goes for `value.trim()`.

ihor-jpeg (changes requested):
It is a good and working solution, but let's use the array iteration methods only. Without loops.

ihor-jpeg (approved):
Good work!

PR #844, opened 2021-11-17
Solution of style-to-object

Dima-Ziakhor:


mariia-bila (approved):
good work 👍

src/convertToObject.js
```diff
@@ -12,7 +12,27 @@
...
+  const sourceStringArr = sourceString.split(';').map(
+    item => item.split(':')
+  ).map(
+    ([prop, value]) => (
+      {
+        [prop.trim()]: (value + '').trim(),
+      }
+    )
+  );
+
```
aionova, line 24:
```suggestion
  const sourceStringArr = sourceString
    .split(';')
    .map(item => item
      .split(':'))
      .map(([prop, value]) => ({
        [prop.trim()]: (value + '').trim()
      })
    );

```

src/convertToObject.js
```diff
@@ -12,7 +12,27 @@
...
+  const result = {};
+
+  Object.assign(result, ...sourceStringArr);
+
+  for (const key in result) {
+    if (!key) {
+      delete result[key];
+    }
+  }
```
aionova, line 33: let's try to use reduce method instead of this loop

aionova (changes requested):

ihor-jpeg (changes requested):
Let's fix some mistakes

src/convertToObject.js
```diff
@@ -12,7 +12,28 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceStringArr = sourceString
+    .split(';')
+    .map(item => {
+      const newItem = item.split(':');
+
+      return {
+        [newItem[0].trim()]: (newItem[1] + '').trim(),
```
aionova, line 21: you could leave it as array here and creaet object only inside of reduce

aionova (approved):
GJ!

PR #846, opened 2021-11-17
solution

volodymyr-soltys97:


msaikun (approved):
Well done!

src/convertToObject.js
```diff
@@ -12,7 +13,19 @@
...
+
+  const propertiesArray = sourceString
+    .split(';')
+    .map(property => property.split(':'))
+    .filter(property => property.length >= 2)
+    .map(property => [property[0].trim(), property[1].trim()]);
+
+  propertiesArray.map(element => {
+    cssObject[element[0]] = element[1];
+  });
```
aionova, line 26: could you do 1 map instead of 2?

aionova (changes requested):

src/convertToObject.js
volodymyr-soltys97: I did not succeed. I replace map on forEach, but I don't know is it a good idea :)

ihor-jpeg (changes requested):
There are a few things you can do. You can implement all the logic of map in just one map method. Or you can change the last map method to reduce, and implement the for each logic there.

src/convertToObject.js
```diff
@@ -11,8 +11,20 @@
...
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const cssObject = {};
+  const propertiesArray = sourceString.split(';');
+
+  propertiesArray.map(property => {
+    const arrayProperties = property.split(':').map(element => element.trim());
```
ihor-jpeg, line 20:
```suggestion
    const arrayProperties = property
      .split(':')
      .map(element => element.trim());
```

ihor-jpeg (approved):
Good work!

PR #848, opened 2021-11-17
First solution

WorkAccCom:


ImIkator (approved):
Well done !

aionova (changes requested):
good job! let's try to use array methods instead of loop now

src/convertToObject.js
```diff
@@ -11,8 +11,23 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
AlinaMatuschak, line 16: you can take a better name

src/convertToObject.js
```diff
@@ -11,8 +11,23 @@
...
+
+  const properties = sourceString.split(';');
+
+  properties.forEach((property) => {
+    const propertyArr = property
+      .split(':')
+      .map(item => item.trim());
+
+    if (propertyArr[0] !== '') {
+      result[propertyArr[0]] = propertyArr[1];
```
AlinaMatuschak, line 26: better create separate constants for propertyArr[0] and propertyArr[1]

AlinaMatuschak (approved):
Nice work 👌

PR #849, opened 2021-11-18
completed style to object

i-stanko:


Nadiyahr (approved):

src/convertToObject.js
```diff
@@ -11,8 +11,17 @@
  *
  * @return {object}
  */
-function convertToObject(sourceString) {
-  // write your code here
-}
+const convertToObject = (sourceString) => {
+  const array = sourceString
+    .split(';')
+    .filter(style => style.length > 5)
```
ihor-jpeg, line 17: what is 5?

src/convertToObject.js
```diff
@@ -11,8 +11,17 @@
...
+    .map(style => style.split(':'));
+
+  const styles = array.reduce((object, [key, value]) => ({
+    ...object, [key.trim()]: value.trim(),
+  }), {});
```
ihor-jpeg, line 22: you can do this in one go

ihor-jpeg (changes requested):

ihor-jpeg (changes requested):
Good work, but let's fix the comments

aionova (approved):
LGTM!

PR #851, opened 2021-11-18
init

ghost:


IrynaHerasymchuk (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
...
+  const splitedStr = sourceString.split(';');
+
+  splitedStr.forEach(element => {
+    const values = element.split(':');
+    const [key, value] = values;
+
+    if (key.trim()) {
+      obj[key.trim()] = value.trim();
+    }
+  });
```
ihor-jpeg, line 26: For each is alright, but can we do it with other array methods?

ihor-jpeg (approved):
Good work!

PR #855, opened 2021-11-18
solution

lanebx:


tmrsl (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,22 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  sourceString.split(';')
+    .map(item => item.trim().split(':'))
+    .filter(item => {
```
aionova, line 19: reduce would fit better here than filter

aionova (changes requested):

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';')
+    .map(item => item.trim().split(':'))
+    .reduce((next, item) => {
```
ihor-jpeg, line 17:
```suggestion
  return sourceString
    .split(';')
    .map(item => item.trim().split(':'))
    .reduce((next, item) => {
```

ihor-jpeg (approved):
Nicely done!

PR #857, opened 2021-11-18
implemented task

tmrsl:


lanebx (approved):
GJ

src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';').reduce((acc, cur) => {
+    const [key, value] = cur.split(':');
+
+    if (key && value) {
```
ihor-jpeg, line 18: this will also work if my key is '   ' and my value is '     '; this expression will only equal to false if one of two strings is an empty string. You have not covered the case where both strings contain only spaces

ihor-jpeg (changes requested):
Good work! But let's fix the comment

src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';').reduce((acc, cur) => {
+    const [key, value] = cur.split(':');
+
+    if (/\S/.test(key) && /\S/.test(value)) {
+      acc[key.trim()] = value.trim();
+    }
```
ihor-jpeg, line 20: i meaaaan it works. But i was talking about trimming the key and value before the if block

ihor-jpeg (changes requested):

src/convertToObject.js
```diff
@@ -12,7 +12,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';').reduce((acc, cur) => {
+    const [rawKey, rawValue] = cur.split(':');
+    const key = rawKey && rawKey.trim();
```
AlinaMatuschak, line 17:
```suggestion
    const key = rawKey.trim();
```

src/convertToObject.js
```diff
@@ -12,7 +12,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';').reduce((acc, cur) => {
```
AlinaMatuschak, line 15: 🔥

AlinaMatuschak (approved):
Greate job 🚀

PR #861, opened 2021-11-19
Solution

Veronika-Korona:
Fixed

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stringArr = sourceString.split(';').map(string => (
+    string.split(':').map(e => e.trim()))
+  );
+
+  const callback = (css, style) => {
+    return {
+      ...css,
+      [style[0]]: style[1],
+    };
+  };
+
+  return stringArr.filter(element => element.length > 1).reduce(callback, {});
 }
 
 module.exports = convertToObject;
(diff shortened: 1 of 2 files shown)


alinaShchavinska (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stringArr = sourceString.split(';').map(string => (
+    string.split(':').map(e => e.trim()))
+  );
```
aionova, line 17:
```suggestion
  const stringArr = sourceString
    .split(';')
    .map(string => (
      string
        .split(':')
        .map(e => e.trim())
    ));
```

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
+
+  const callback = (css, style) => {
+    return {
+      ...css,
+      [style[0]]: style[1],
+    };
+  };
+
+  return stringArr.filter(element => element.length > 1).reduce(callback, {});
```
aionova, line 26:
```suggestion
  return stringArr
    .filter(element => element.length > 1)
    .reduce(callback, {});
```

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stringArr = sourceString.split(';').map(string => (
+    string.split(':').map(e => e.trim()))
```
aionova, line 16: no need to use map inside of map. You can trim it inside of reduce callback

aionova (changes requested):

illiarerun (approved):
Good job 🔥

PR #862, opened 2021-11-19
Solution

alinaShchavinska:


VladDyrenkov (approved):
GJ

src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssObj = {};
+  const cssArrey = sourceString.split(';');
+
+  cssArrey.map(function(lineCss) {
```
AlinaMatuschak, line 18: you can there use callback
```suggestion
  cssArrey.map((lineCss) => {
```

AlinaMatuschak, line 18: use forEach if you don’t return something

AlinaMatuschak (changes requested):
Almost perfect 👍

AlinaMatuschak (approved):
🔥

PR #875, opened 2021-12-08
Solution

victorgri:


src/convertToObject.js
```diff
@@ -12,7 +12,22 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+ const newObject = {};
+
+  const filterString = sourceString.split(';');
+
+  for (let i = 0; i < filterString.length; i++) {
```
Anastasiia-Svintsova, line 19: Instead of using a standard for loop, use a for...of loop which is more concise and easier to read. Or you can think about array methods.

src/convertToObject.js
```diff
@@ -12,7 +12,22 @@
...
+    let propertyAndValue = filterString[i].trim();
+
+    if (propertyAndValue === '') {
+      continue;
+    }
+
+    propertyAndValue = propertyAndValue.split(':');
```
Anastasiia-Svintsova, line 26: Use destructuring to directly extract the key and value from each key-value pair

src/convertToObject.js
```diff
@@ -12,7 +12,22 @@
...
+ const newObject = {};
+
+  const filterString = sourceString.split(';');
+
+  for (let i = 0; i < filterString.length; i++) {
+    let propertyAndValue = filterString[i].trim();
+
+    if (propertyAndValue === '') {
+      continue;
+    }
```
Anastasiia-Svintsova, line 24: Seems redundant

Anastasiia-Svintsova (changes requested):
- You can remove commented code to get rid of the errors 
image

maxim2310 (changes requested):
Seems like you forgot to deploy your changes, deploy it and get new request for review

Anastasiia-Svintsova (approved):

PR #887, opened 2021-12-22
Solution

Bohdan-Maksymiuk:


src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
vladyslav-harkusha, line 15: чеклист

vladyslav-harkusha (changes requested):

vladyslav-harkusha (approved):

danheim (approved):

PR #892, opened 2021-12-22
Solution🤨

destroythemoon:


src/convertToObject.js
```diff
@@ -12,7 +12,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const res = {};
```
Basil-Panasiuk, line 15: maybe rename variable

Basil-Panasiuk (approved):

src/convertToObject.js
danheim: rename variable

danheim (changes requested):

danheim (approved):

PR #919, opened 2021-12-26
solution_v1.0

Rozdorozhnii:


ViktorBachynskyi (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const attributes = sourceString.split(';').map(rule => rule.trim());
```
Olexandr6, line 16:
```suggestion
  const attributes = sourceString
  	.split(';')
  	.map(rule => rule.trim());
```

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
+
+  for (let i = 0; i < attributes.length; i++) {
+    if (attributes[i]) {
+      const entry = attributes[i].split(':').map(rule => rule.trim());
```
Olexandr6, line 20:
```suggestion
      const entry = attributes[i]
      	.split(':')
      	.map(rule => rule.trim());
```

Olexandr6 (approved):
Good job 👍  Check comments 👇

PR #920, opened 2021-12-27
sm msg

Mikedacheka:
thank you for review!

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -11,8 +11,21 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
   // write your code here
+  const convertedObject = sourceString
+    .split(';')
+    .filter(property => property.length > 4)
+    .reduce(function(CSSObject, property) {
+      const propPair = property.split(':');
+
+      CSSObject[propPair[0].trim()] = propPair[1].trim();
+
+      return CSSObject;
+    }, {});
+
+  return convertedObject;
 }
 
 module.exports = convertToObject;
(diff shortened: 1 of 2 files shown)


IrynaZolko (approved):

src/convertToObject.js
```diff
@@ -11,8 +11,21 @@
...
+  const convertedObject = sourceString
+    .split(';')
+    .filter(property => property.length > 4)
+    .reduce(function(CSSObject, property) {
+      const propPair = property.split(':');
+
+      CSSObject[propPair[0].trim()] = propPair[1].trim();
+
+      return CSSObject;
+    }, {});
```
Olexandr6, line 26: Use arrow functions as callback, it looks shorter and more readable 
```suggestion
    .reduce((CSSObject, property) => {
      const propPair = property.split(':');

      CSSObject[propPair[0].trim()] = propPair[1].trim();

      return CSSObject;
    }, {});
```

src/convertToObject.js
```diff
@@ -11,8 +11,21 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
   // write your code here
+  const convertedObject = sourceString
```
Olexandr6, line 17: Avoid words `obj, arr, str` in var naming

src/convertToObject.js
```diff
@@ -11,8 +11,21 @@
...
+    .split(';')
+    .filter(property => property.length > 4)
+    .reduce(function(CSSObject, property) {
+      const propPair = property.split(':');
+
+      CSSObject[propPair[0].trim()] = propPair[1].trim();
```
Olexandr6, line 23: Why `CSSObject` is yellow, what is this? 🤔

Olexandr6 (approved):
Good job 👍

PR #953, opened 2022-01-28
solution

Tsekhmister:


013g (approved):

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleCss = {};
+
+  const filteredString = sourceString
```
Anna-Tkachenko, line 17: Just create more meaningful variable name and don't use data type name in variable names

Anna-Tkachenko (approved):
🔥

PR #956, opened 2022-01-29
Solution

AndryKK:


src/convertToObject.js
```diff
@@ -12,7 +12,22 @@
...
+  const convertedToObject = {};
+
+  const TempArray = sourceString.split(';').map(item => item.trim());
+
+  const TempArrayFiltred = TempArray.filter(item => item.length !== 0);
+
+  const TempArrayOfArrayes = TempArrayFiltred
+    .map(
+      item => item.split(':').map(itemIn => itemIn.trim())
+    );
```
Anna-Tkachenko, line 24: You can use iteration methods in chain instead of using extra variables
```suggestion
  const TempArray = sourceString.
  	split(';')
  	.map(item => item.trim())
  	.filter(item => item.length !== 0)
  	.map(
      item => item.split(':').map(itemIn => itemIn.trim())
    );
```

Anna-Tkachenko (changes requested):
Just fix code style, delete redundant variables and fix variable names: it should start from small letter

PR #960, opened 2022-01-30
Done

tt-denis:


src/convertToObject.js
```diff
@@ -12,7 +12,12 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const formatedArr = sourceString.split(';')
+    .map(propertys => propertys.split(':')
+      .map(values => values.trim()))
+    .filter(element => element.length > 1);
```
```suggestion
  const formatedArr = sourceString
  	.split(';')
    .map(propertys => propertys.split(':')
      .map(values => values.trim()))
    .filter(element => element.length > 1);
```

Anna-Tkachenko (approved):
Create more meaningful variable names instead of formattedArr

PR #962, opened 2022-01-30
Solution

maxim2310:


src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const properties = sourceString.split(';');
+  const trimed = properties.map(prop =>
+    prop.split(':').map(elem => elem.trim()));
```
Anna-Tkachenko, line 17: You can do it without extra variable
```suggestion
  const properties = sourceString
  	.split(';')
  	.map(prop =>
    	prop.split(':').map(elem => elem.trim())
    );
```

Anna-Tkachenko (approved):

PR #964, opened 2022-01-30
sol

Kengat:


src/convertToObject.js
```diff
@@ -12,7 +12,12 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const formatedArray = sourceString.split(';')
+    .map(properties => properties.split(':')
+      .map(keyValue => keyValue.trim()))
+    .filter(element => element.length > 1);
```
```suggestion
  const formatedArray = sourceString
  	.split(';')
    .map(properties => properties.split(':')
      .map(keyValue => keyValue.trim()))
    .filter(element => element.length > 1);
```

Anna-Tkachenko (approved):
Good job
It's better to create more meaningful variable names instead of formattedArray

PR #973, opened 2022-02-05
convertToObject function

MykhailoIvchenko:


src/convertToObject.js
```diff
@@ -3,16 +3,46 @@
...
+      index++;
+    }
+  }
+
+  propValuesArray = propValuesArray.map(property => property.split(':'));
+
+  for (let i = 0; i < propValuesArray.length; i++) {
+    propValuesArray[i][0] = propValuesArray[i][0].trim();
+    propValuesArray[i][1] = propValuesArray[i][1].trim();
+  }
```
aliceblunt8, line 37: don't use loops in this task, use array methods
also no need to create a new array

aliceblunt8 (changes requested):

MykhailoIvchenko:
The solution was fixed. For the object properties assignment forEach method was used. In fact it is also a loop, but applied as  array's method)))

danylo-onishchuk (approved):

PR #981, opened 2022-02-14
solution

SanyaBratashchuk:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+
+  const result = {};
+
+  normalize.map(propValuePair => {
+    const [key, value] = propValuePair.split(':');
+
+    result[key] = value;
+
+    return propValuePair;
+  });
```
GoDfreE25, line 26: Nice option I thought how to do it myself

GoDfreE25 (approved):
Nice Work

solaryasha (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const normalizedSourceString = sourceString.split(';').map(propValuePair => {
+    return propValuePair
+      .split(':')
+      .map(propOrValue =>
+        propOrValue.trim()).join(':');
+  }).filter(propValuePair => propValuePair.length > 1);
```
solaryasha, line 18:
```suggestion
  const normalizedSourceString = sourceString
  .split(';')
  .map(propValuePair => {
    return propValuePair
      .split(':')
      .map(propOrValue =>
        propOrValue.trim()).join(':');
  })
  .filter(propValuePair => propValuePair.length > 1);
```

solaryasha (approved):

PR #982, opened 2022-02-15
Sulution

Andr1yK:


aspurgilus (approved):
Interesting solution.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splitedSource = sourceString.split(';');
+  const cssStyle = {};
```
solaryasha, line 18:
```suggestion
  const cssStyles = {};
```

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  for (const style of splitedSource) {
+    const [ key, ...properties ] = style
```
solaryasha, line 21:
```suggestion
    const [key, ...properties] = style
```

solaryasha (approved):
Well done !

PR #995, opened 2022-02-17
JS style-to-object HW

Ostafii4uk:


src/convertToObject.js
```diff
@@ -3,16 +3,24 @@
...
+ * and values are the values of related
+ *  CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';').map(pair => {
```
RomanKushyk, line 17: Maybe You can use .reduce method

RomanKushyk (approved):
So good!

src/convertToObject.js
solaryasha, line 17:
```suggestion
  return sourceString
   .split(';')
   .map(pair => {})
   .filter()
   .reduce()
```

solaryasha (approved):

PR #999, opened 2022-02-18
soltion

RomanKushyk:


krolich500grn (approved):
it seems really great in compare with my solution! good job! 👍

PR #1000, opened 2022-02-18
complete task

irreligious86:


krolich500grn (approved):
for me it's okay, but Vova will reject it, because we practicing at array iteration methods, that you didn't use here. 

You can start rewrite this)

Ihor-Karpyn (changes requested):
don't use loops here

Ihor-Karpyn (approved):

PR #1007, opened 2022-02-23
Solution for js_style-to-object

murat-yesimbietov:


src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString.split(';');
+
+  const withoutSpaces = styles.map(style => {
+    let elems = style.split(':');
+
+    elems = elems.map(element => element.trim());
```
vitaliy-martynyuk, line 22:
```suggestion
    let elems = style.split(':').map(element => element.trim());
```

vitaliy-martynyuk (approved):
well done! 🚀

PR #1008, opened 2022-03-04
solution

djtyrf312:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const elements = sourceString.split(';');
+  const filteredElements = [];
+  const result = {};
+
+  for (let i = 0; i < elements.length; i++) {
+    if (elements[i].length > 4) {
```
Bohdan-Maksymiuk, line 22: what are you doing here?

Bohdan-Maksymiuk (changes requested):

src/convertToObject.js
djtyrf312: Here I have filtered spaces( check pls new solution I have been made

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+  for (let i = 0; i < elements.length; i++) {
+    const trimedLine = elements[i].trim();
+
+    if (trimedLine !== '') {
+      filteredElements.push(trimedLine.split(':'));
+    }
+  }
+
+  for (let j = 0; j < filteredElements.length; j++) {
+    stylesObject[filteredElements[j][0].trim()] = filteredElements[j][1].trim();
```
ArtemTopolskyi, line 30: Лучше вынести `filteredElements[j][0].trim()` в отдельные переменные

ArtemTopolskyi (approved):

PR #1030, opened 2022-04-03
Solution

mikhail-shchuka:


src/convertToObject.js
```diff
@@ -3,16 +3,46 @@
...
+    items = items.map(preparationProp);
+
+    return {
+      [items[0]]: items[1],
+    };
+  };
+
+  const resultAraay = properties.map(divideProp);
+
+  return resultAraay.reduce((pre, item) => (
```
aliceblunt8, line 41: too complicated, where it's unnecessary 
start by splitting by `;`, then each element by `:`, then just trim each element
then loop through the resulting array and assign key and value to prepared in advance empty object

aliceblunt8 (changes requested):

src/convertToObject.js
Bohdan-Maksymiuk: > then loop through the resulting array and assign key and value to prepared in advance empty object

Bohdan-Maksymiuk (changes requested):
You iterate twice through the array, but can do it only ones.
Create an empty object first, then in the loop pass there key/properties pair
And then there will be no need in additional reduce method

ArtemTopolskyi (approved):

PR #1032, opened 2022-04-04
Solution

yaremenko-maksym:


OksanaTyshchenko (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,63 @@
...
+  const lastIndex = string.indexOf(':');
+
+  return string.slice(firstIndex, lastIndex).trim();
+}
+
+function splitValue(string) {
+  const valuePart = string.slice(string.indexOf(':') + 1);
+
+  return valuePart.replace(/\s+/g, ' ').trim();
+}
```
AlinaMatuschak, line 27: try to use the split method by :. First element will be key, second value

src/convertToObject.js
```diff
@@ -3,16 +3,63 @@
...
+    } else if (b.indexOf('-') === 0) {
+      second = b.slice(1);
+    }
+
+    return first - second;
+  }).reduce((obj, key) => {
+    obj[key] = cssProperties[key];
+
+    return obj;
+  }, {});
```
AlinaMatuschak, line 60: is it necessary?

AlinaMatuschak (changes requested):

src/convertToObject.js
yaremenko-maksym: Doesn`t pass the sorting test and i had to correct the sorting.

src/convertToObject.js
yaremenko-maksym: With the initial split (';') we get hyphens (/n) at the beginning of the line, and if we do another split (as I tried), then these /n get into the keys and we still have to get rid of them.

src/convertToObject.js
yaremenko-maksym: image

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+}
+
 function convertToObject(sourceString) {
-  // write your code here
+  const cssProperties = {};
+  const splitedStr = sourceString.split(';');
+
+  for (const string of splitedStr) {
+    if (splitKey(string) && splitValue(string)) {
+      cssProperties[splitKey(string)] = splitValue(string);
```
AlinaMatuschak, line 30: you 2 times get the key and value

```suggestion
    const key = splitKey(string);
    if (key && splitValue(string)) {
      cssProperties[key] = splitValue(string);
```

AlinaMatuschak (approved):

PR #1038, opened 2022-04-08
convertToObject function was created

OksanaTyshchenko:


yaremenko-maksym (commented):
Very Good!

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+  for (const item of newPartsArr) {
+    Object.assign(result, item);
+  }
+
+  return result;
+}
+
+function changesParts(part) {
+  const index = part.indexOf(':');
+  const key = part.slice(0, index).trim();
```
AlinaMatuschak, line 30:
```suggestion
  const key = part[0].trim();
```

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+  const result = {};
+
+  for (const item of newPartsArr) {
+    Object.assign(result, item);
+  }
+
+  return result;
+}
+
+function changesParts(part) {
```
AlinaMatuschak, line 28: you can use it with forEach instead of a map and add new values straight to result object. You don't need 2 cycles

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):

PR #1039, opened 2022-04-08
Solution

strukk311:


HelenPetrovska (approved):
Good job!

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const cssObject = {};
+  const properties = sourceString.split(';');
+
+  properties.map(property => {
```
AlinaMatuschak, line 21: the map method returns a new array and doesn't change anything outside, while the forEach method doesn't return anything, but he can change something outside. Here better use forEach

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
```
AlinaMatuschak, line 17: you can delete comment

AlinaMatuschak (changes requested):

PR #1041, opened 2022-04-08
Create solution

ivan-kadykalo:


HelenPetrovska (approved):
I like it!

src/convertToObject.js
```diff
@@ -1,18 +1,20 @@
...
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const splitString = sourceString.split(';');
+  const styleList = {};
+
+  splitString.forEach(item => {
+    if (item.trim().length > 0) {
```
AlinaMatuschak, line 8: 0 converts to false

```suggestion
    if (item.trim().length) {
```

AlinaMatuschak (approved):

PR #1043, opened 2022-04-09
styles were created

HelenPetrovska:


strukk311 (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
- *
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const arrStrings = sourceString.split(';')
+    .map(el => el.split(':'));
```
AlinaMatuschak, line 13: it isn’t an array with strings. It is an array with arrays, right? Better take a name like arrauProperties

AlinaMatuschak (changes requested):
be more attentive with names

AlinaMatuschak (approved):

PR #1045, opened 2022-04-10
solution

barannikovArtem:


EvgenBabin1 (approved):
ok, good job

src/convertToObject.js
```diff
@@ -3,16 +3,22 @@
...
+    .split(';')
+    .map(el => el.split(':'))
+    .filter(el => el.length > 1);
+
+  return listOfStyles.reduce((styles, [key, value]) => {
+    return {
+      ...styles,
+      [key.trim()]: value.trim(),
+    };
+  }, {});
```
AlinaMatuschak, line 21:
```suggestion
  return listOfStyles.reduce((styles, [key, value]) => ({
    ...styles,
    [key.trim()]: value.trim(),
  }), {});
```

or you can check here key and value existing and delete the filter from getting listOfStyles (-1 cycle)

AlinaMatuschak (approved):

PR #1046, opened 2022-04-11
Solution

ElizaChernysh:


src/convertToObject.js
```diff
@@ -3,15 +3,30 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
+  const result = {};
+  const newString = sourceString.split(';').map(elem => elem.split(': '));
+
+  for (const i in newString) {
+    if (newString[i].length > 1) {
+      const key = newString[i][0].replace(/\s/g, '');
+      const value = newString[i][1].replace(/^ +| +$|( ) +/g, '$1');
```
Anastasiia-Svintsova, line 23: You can also use reduce method instead of for-of loop. But your solution is good too)

Anastasiia-Svintsova (approved):
Good!

src/convertToObject.js
```diff
@@ -3,15 +3,30 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
+  const result = {};
+  const newString = sourceString.split(';').map(elem => elem.split(': '));
```
AlinaMatuschak, line 18: Is it a string? I thought that array 🤔

```suggestion
  const betterName = sourceString.split(';').map(elem => elem.split(': '));
```

src/convertToObject.js
```diff
@@ -3,15 +3,30 @@
...
+    if (newString[i].length > 1) {
+      const key = newString[i][0].replace(/\s/g, '');
+      const value = newString[i][1].replace(/^ +| +$|( ) +/g, '$1');
+
+      result[key] = value;
+    }
+  }
+
+  return result;
   // write your code here
```
AlinaMatuschak, line 30: delete comment

AlinaMatuschak (approved):

PR #1054, opened 2022-04-12
Solution

plankton9991:


misha200119 (approved):
good job

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+  const keys = sourceString.split(';').map(ex => ex.trim())
+    .filter(e => e.length > 0);
+
+  keys.forEach(key => {
+    const part = key.split(':');
+    const prop = part[0].trim();
+    const value = part[1].trim();
+
+    res[prop] = value;
+  });
```
AlinaMatuschak, line 28: 2 cycles instead of 4 )

```suggestion
  const keys = sourceString.split(';');

  keys.forEach(prop => {
    const trimmedProperty = prop.trim()
    
    if (trimmedProperty.length > 0) {
      const part = key.split(':');
      const prop = part[0].trim();
      const value = part[1].trim();

      res[prop] = value;
    }
  });
```

AlinaMatuschak (approved):

PR #1072, opened 2022-04-17
solut

KostiaO:


Bohdan-Maksymiuk (changes requested):
You can use this for preparing lines and then add properties and values to the object
```
  const lines = sourceString.split(';')
    .map(line => line.split(':'))
    .filter(line => line.length === 2);
```

KostiaO:
Changed

PR #1075, opened 2022-04-19
Solution

derenivska:


src/convertToObject.js
```diff
@@ -3,16 +3,67 @@
...
+        value = value.slice(1);
+      }
+
+      result[key] = value;
+    }
+
+    return result;
+  }
+
+  return {};
```
Bohdan-Maksymiuk, line 66: why?

Bohdan-Maksymiuk (changes requested):
you should try to use iterative array methods
```
  const lines = sourceString.split(';')
    .map(line => line.split(':'))
    .filter(line => line.length === 2);
```
you may use this to prepare lines and then add properties to style object (use `trim()` to remove unnecessary indentation)

src/convertToObject.js
vkryvytskyy: remove this return

vkryvytskyy (approved):

PR #1081, opened 2022-05-01
Solution

Mikhail-Danylenko:


src/convertToObject.js
```diff
@@ -3,16 +3,41 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.split(';');
+
+  const arr1 = [];
```
aliceblunt8, line 19: no need to create a new array and push there, use  .map method to split by : and then trim each item

src/convertToObject.js
```diff
@@ -3,16 +3,41 @@
...
+    const str1 = str.split(':');
+
+    arr1.push(str1);
+  });
+
+  arr1.forEach((str) => {
+    const key = str[0].trim();
+
+    objectResult[key] = typeof str[1] === 'string' ? str[1].trim() : '';
+  });
```
aliceblunt8, line 32: this could be done inside map

src/convertToObject.js
```diff
@@ -3,16 +3,41 @@
...
+
+  for (const key in objectResult) {
+    if (key.length === 0 && objectResult[key].length === 0) {
+      delete objectResult[key];
+    }
+  }
```
aliceblunt8, line 38: ad only after you prepare a two-dimensional array use forEach to get first and last value from each array and write them as key and value to object

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #1085, opened 2022-05-02
ConverToObject was done

vlasiuk-anatolii:


EnjSadman (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const res = {};
+  const arr = sourceString
```
Anna-Tkachenko, line 18: Pay attention to variable names

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    .replace(/\s{2,}|^\s/g, '')
+    .split(';')
+    .filter(a => a.length > 2);
+
+  for (let i = 0; i < arr.length; i++) {
+    arr[i] = arr[i].trim().replace(/\s(?=:)|^\s/gm, '');
```
Anna-Tkachenko, line 24: You can split all elements by : separator, trim them using map method and then just add final values to the object
Such way will be more understandable and help to avoid so many matches

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const res = {};
+  const arr = sourceString
+    .replace(/\s{2,}|^\s/g, '')
+    .split(';')
+    .filter(a => a.length > 2);
+
+  for (let i = 0; i < arr.length; i++) {
```
Anna-Tkachenko, line 23: Don't use cycle in array iteration methods topic

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job, but you can make it easier, check solutions of other students)

PR #1088, opened 2022-05-03
string to an object with properties

koshachina:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const result = {};
+  // preparing array to data extraction
```
ArtemTopolskyi, line 19: don't left comments

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+  let arrOfProp = sourceString.split(';');
+
+  arrOfProp = arrOfProp.map(value => value.replace(/\n*/g, ''));
```
ArtemTopolskyi, line 22: create separate variable for the result of map

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+
+  arrOfProp.forEach(element => {
```
ArtemTopolskyi, line 24: use reduce method

ArtemTopolskyi (changes requested):

ArtemTopolskyi (approved):

PR #1089, opened 2022-05-04
solution

alexandra-protyanova:


src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+ * and values are the values of related CSS properties
+ *  (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.split(';').map(x => x.split(':'));
```
vlad-demchuk, line 17: [CODE STYLE] - All chained methods called on the same level should have same indentation. This makes code much more readable (checklist)

vlad-demchuk (approved):
Well done)

src/convertToObject.js
Anna-Tkachenko, line 17:
```suggestion
  const arr = sourceString
    .split(';')
    .map(x => x.split(':'));
```

Anna-Tkachenko: Pay attention to variable names: let's create more meaningful name instead of arr)

Anna-Tkachenko (approved):
⭐️

PR #1094, opened 2022-05-04
solution

ZUBOV-ILLIA:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  const inpArr = sourceString.split(':').map(x => x.trim()).join(':')
+    .split(';').map(x => x.trim()).filter(x => x !== '');
+
+  for (let i = 0; i < inpArr.length; i++) {
```
alexandra-protyanova, line 21: you can use map here

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  const inpArr = sourceString.split(':').map(x => x.trim()).join(':')
```
alexandra-protyanova, line 18: [CODE STYLE] - All chained methods called on the same level should have same indentation. This makes code much more readable (checklist)

alexandra-protyanova (approved):
Good job!

Anna-Tkachenko (changes requested):
Fix @alexandra-protyanova comments

src/convertToObject.js
Anna-Tkachenko: Don't use cycles in array iteration methods task)

Anna-Tkachenko (changes requested):

denys-cheporniuk (approved):

PR #1095, opened 2022-05-04
solution

yelnikov-andrii:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayOfValues = sourceString.split(';');
+  const obj = {};
+
+  for (let i = 0; i < arrayOfValues.length; i++) {
```
alexandra-protyanova, line 20: you can use forEach instead of for loop

alexandra-protyanova (approved):
👍

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+ * and values are the values of related CSS properties (see an exampl in
+ * [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayOfValues = sourceString.split(';').map(item => item.split(':'));
```
Anna-Tkachenko, line 17:
```suggestion
  const arrayOfValues = sourceString
    .split(';')
    .map(item => item.split(':'));
```

Anna-Tkachenko (approved):
⭐️

PR #1097, opened 2022-05-05
Solution

amitpatiljc:


src/convertToObject.js
```diff
@@ -3,16 +3,23 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = [];
+
+  const properties = sourceString.split(';')
+    .map(item => item.trim().split(':'))
+    .filter(item => !item.includes(''));
```
rHashTag, line 17: For better readability, nest the methods that are applied.

```
    .split(';')
    .map(item => item.trim().split(':'))
    .filter(item => !item.includes(''));

rHashTag (changes requested):
Looks Good!

One change requested.

rHashTag (approved):
LGTM!

PR #1098, opened 2022-05-05
solution

SavMary:


src/convertToObject.js
```diff
@@ -12,7 +13,21 @@
...
+  const obj = {};
+  const arr = sourceString
+    .split(';')
+    .map(el => el.trim())
+    .filter(el => el.length !== 0);
+
+  arr.forEach(element => {
+    const index = element.split('').findIndex(el => el === ':');
+    const key = element.slice(0, index).trim();
+    const value = element.slice(index + 1).trim();
```
vlad-demchuk, line 25: in my opinion, 'index' is redundant here. you can avoid this using only split(':')[index] for key and value

vlad-demchuk (approved):

Anna-Tkachenko (approved):
⭐️

PR #1099, opened 2022-05-05
task completed

andreisucman:


59MrRobot (approved):
Good job 👍

src/convertToObject.js
```diff
@@ -1,18 +1,21 @@
...
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  const someMagic = sourceString.split(';')
+    .map(element => element.split(':'));
+
+  someMagic.forEach(string => {
+    if (string.length > 1) {
```
meforoma, line 10: nice, but one important semantic comment - `someMagic` is an array of arrays, right?
so forEach loop interacts not with the string, but with an array.
Let's keep code properly descriptive 😉

meforoma (changes requested):

src/convertToObject.js
andreisucman: True, fixed.

PR #1102, opened 2022-05-06
solution

AnastasiyaSopotnytska:


vlad-demchuk (approved):
Good job!) i guess, it'll be better to make more variables to understand better the code)

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const result = {};
+  const arrayWithoutSpaces = sourceString
+    .split(';').map(x => x.trim()).filter(x => x.length > 0);
```
Denchepornyuk, line 20:
```suggestion
    .split(';')
    .map(x => x.trim())
    .filter(x => x.length > 0);
```

Denchepornyuk (approved):

PR #1103, opened 2022-05-06
Solution

huberciak:


andreisucman (approved):
Great job.

src/convertToObject.js
```diff
@@ -1,18 +1,33 @@
...
+properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString
+    .split(';')
```
meforoma, line 18:
```suggestion
    .split(';').filter(item => item)
    // if item is falsy (equivalent of false) filter will leave it out
```

src/convertToObject.js
```diff
@@ -1,18 +1,33 @@
...
+    .map(style => style.split(':')
+      .map(value => value.trim()));
+
+  const stylesObject = {};
+
+  for (const style of stylesArray) {
+    if (style[0] !== undefined && style[1] !== undefined) {
+      stylesObject[style[0]] = style[1];
+    }
+  }
```
meforoma, line 28: if applied the above filter, you won't need this loop and an object
```suggestion
```

meforoma (changes requested):

VinnieJ-2k20 (approved):

PR #1105, opened 2022-05-07
solution

lytvyshko:


src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  sourceString
+    .split(':')
+    .map(item => item.trim())
+    .join(':')
+    .split(';')
+    .map(item => item.trim())
```
dellachaise, line 11: why it necessary to use trim() twice at 8 and 11 lines?

dellachaise (approved):
Please answer for comment.

src/convertToObject.js
lytvyshko: This is definitely not the best solution and it violates DRY rule.  But I wanted to solve it in one methods chain.

Answer to your question. Try this code in your IDE:

```
let string = "font-family: inherit       ;        font-size: 14px;";

let arr = string
            .split(':')
            .map(item => item.trim());

console.log(arr);    // ["font-family", "inherit       ;        font-size", "14px;"]
```

There are still spaces before and after the semicolon. So I have to join it back, split it by semicolons and use trim again.

Denchepornyuk (approved):

PR #1109, opened 2022-05-08
css to obj

elkhayate:


proznerka (approved):
I think, you can do this task with just one map and one filter.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    .map(elm => elm.trim())
+    .filter(elm => elm.length !== 0)
+    .map(elm => elm.split(':')
+      .map(val => val.trim()));
+
+  for (const prop of objProp) {
+    cssObj[prop[0]] = prop[1];
+  }
+
+  return cssObj;
```
meforoma, line 29: since you get an array of arrays, you could use `Object.fromEntries`
```suggestion
  return Object.fromEntries(objProp);
```

meforoma (changes requested):

VinnieJ-2k20 (approved):

PR #1110, opened 2022-05-09
solution

Uktam19980416:


efdalyalcin (approved):
nicely done!

src/convertToObject.js
```diff
@@ -13,6 +13,19 @@
...
 function convertToObject(sourceString) {
   // write your code here
+  const strSplit = sourceString.split(';');
+  const strFilter = strSplit.filter(el => el.trim());
+  const strToObj = {};
+
+  strFilter.forEach(x => {
+    const strChildSplit = x.split(':');
+    let key = strChildSplit[0].trim();
+    let value = strChildSplit[1].trim();
```
meforoma, line 23: love your implementation, just an advise to make it better - de-structure stuff, i'll be more readable:
```suggestion
    const [property, value] = x.split(':');
    let key = property.trim();
    let value = value.trim();
```

meforoma (changes requested):

meforoma (approved):

PR #1111, opened 2022-05-09
Solution

DNK-forexample:


src/convertToObject.js
```diff
@@ -1,18 +1,42 @@
...
+  const chaoticArray = sourceString.split(';');
+  const keysArray = [];
+  const valuesArray = [];
+  const stylesObject = {};
+
+  chaoticArray.forEach(el => {
+    const x = el.split(':');
 
+    keysArray.push(x[0].trim());
+    valuesArray.push(x[1]);
```
valerii-derkach, line 30: why do you .trim keys in one place and values you .trim later? You can .trim them both in this loop.

src/convertToObject.js
```diff
@@ -1,18 +1,42 @@
...
-  // write your code here
-}
+  /* const str = sourceString.replaceAll('\n', ''); */
+  const chaoticArray = sourceString.split(';');
+  const keysArray = [];
+  const valuesArray = [];
+  const stylesObject = {};
+
+  chaoticArray.forEach(el => {
+    const x = el.split(':');
```
valerii-derkach, line 27: what is x? you should use more readable names for variables

valerii-derkach (changes requested):

src/convertToObject.js
```diff
@@ -1,18 +1,42 @@
...
+
+  chaoticArray.forEach(element => {
+    const partOfElement = element.split(':');
 
+    keysArray.push(partOfElement[0].trim());
+    valuesArray.push((partOfElement[1] || '').trim());
+  });
+
+  for (let i = 0; i < keysArray.length; i++) {
+    if (keysArray[i] !== '') {
```
Denchepornyuk, line 34: use method forEach instead of for loop

Denchepornyuk (approved):

PR #1114, opened 2022-05-10
Solution

PRO-GRAM-MER:


proznerka (approved):
Looks good to me

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
   // write your code here
+  const requiredArray = [];
+
+  const properties = sourceString
+    .split(';')
+    .map(element => element.trim().split(':'))
+    .filter(element => !element.includes(''));
+
+  properties.forEach(property => requiredArray
+    .push([property[0].trim(), property[1].trim()]));
```
meforoma, line 26: minor but relevant comment - de-structured approach would make it more readable:
```suggestion
  properties.forEach([property, value] => requiredArray
    .push([property.trim(), value.trim()]));
```

meforoma (approved):

PR #1116, opened 2022-05-10
Solution

RomekIvantsiv:


Denchepornyuk (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
-  // write your code here
+  const result = {};
+  const arrayStyle = sourceString
+    .split(';')
+    .map(element => element.trim())
+    .filter(el => el.length !== 0);
+
+  arrayStyle.forEach(element => {
+    const key = element.split(':')[0].trim();
+    const value = element.split(':')[1].trim();
```
Denchepornyuk, line 25: you can create a variable for `element.split (':')` instead of splitting the element twice

PR #1121, opened 2022-05-10
style to obj

kitaminary:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const obj = {};
+  const arr = sourceString.split(';').map(a => a.trim());
```
Anna-Tkachenko, line 19:
```suggestion
  const arr = sourceString
    .split(';')
    .map(a => a.trim());
```

Anna-Tkachenko: You can also add map and filter here in methods chain

Anna-Tkachenko (approved):
Pay attention to variable names

PR #1122, opened 2022-05-10
solution

proznerka:


src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splited = sourceString.split(/:|;/).map(el => el.trim())
+    .filter(el => el !== '');
+  const newObj = {};
```
MabuzaM, line 19: I like the implementation here. But, what's el?

MabuzaM (changes requested):
Great work. keep it up? Please check the comment I left below.

meforoma (approved):
non trivial but working approach
👍 👍 👍

PR #1128, opened 2022-05-11
js_style-to-object

anastasiia-nikita:


EnjSadman (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+// Через звичайний цикл це набагато простіше написати)
+
 function convertToObject(sourceString) {
-  // write your code here
+  const resultObject = {};
+
+  const arrayString = sourceString.split(';')
+    .map(element => element.split(':'))
+    .map(item => item.map(subItem => subItem.trim()))
+    .filter(value => value[0] !== '');
```
Denchepornyuk, line 24: oes this filter do anything?

Denchepornyuk (approved):

src/convertToObject.js
anastasiia-nikita: Yes, it leaves in our array only those elements in which 0 element(our future key) is not equal to an empty string.
Without this filter, the tests do not pass)

PR #1129, opened 2022-05-11
solution

Sardorbek1122:


proznerka (approved):
Looks good to me

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
-  // write your code here
+  const newArray = [];
+
+  const properties = sourceString.split(';')
+    .map(item => item.trim().split(':'))
+    .filter(item => !item.includes(''));
+
+  properties.forEach(value =>
+    newArray.push([value[0].trim(),
+      value[1].trim()]));
```
meforoma, line 25: legit approach, but let's use de-structuring:
```suggestion
  properties.forEach([property, value] =>
    newArray.push([property.trim(),
      value.trim()]));
```

meforoma (changes requested):

VinnieJ-2k20 (approved):

PR #1140, opened 2022-05-18
converted

Havrysh:


Bohdan-Maksymiuk (approved):
better use semantic naming of variables, not only obj and resultArr

PR #1141, opened 2022-05-21
Solution

stepband:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  */
 function convertToObject(sourceString) {
   // write your code here
+  const object = {};
+  let cssString = sourceString.split(';');
+
+  cssString = cssString.map(element => element.trim());
+  cssString = cssString.map(element => element.split(':'));
+  cssString = cssString.map(element => element.map(str => str.trim()));
+  cssString = cssString.filter(element => element[0] !== '');
```
vkryvytskyy, line 24: use chaining to write method after method

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+
+  for (const data of cssString) {
```
vkryvytskyy, line 26:
```suggestion
  for (const [key, value] of cssString) {
```

vkryvytskyy (changes requested):

PR #1156, opened 2022-06-02
Solution

dmytro-lebedchenko:


src/convertToObject.js
```diff
@@ -1,18 +1,20 @@
...
+  const sourceProperty = sourceSplitted.map(property => property.split(':'));
+
+  sourceProperty.forEach(property => {
+    if (property.length === 2) {
+      const key = property[0].trim();
+      const value = property[1].trim();
+
+      styles[key] = value;
+    }
+  });
```
ArtemTopolskyi, line 15: use reduce method instead

ArtemTopolskyi (changes requested):

src/convertToObject.js
dmytro-lebedchenko: Thank you for your recommendation. I'll act accordingly. 👍

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceSplitted = sourceString.split(';');
+  const sourceProperty = sourceSplitted.map(property => property.split(':'));
+  const sourceFiltered = sourceProperty.filter(
+    property => property.length === 2);
```
aliceblunt8, line 7:
```suggestion
  const sourceArray = sourceString.split(';')
  . map(property => property.split(':')
   .filter(property => property.length === 2);
```

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
+
+  const styles = sourceFiltered.reduce((prev, property) => {
+    const [key, value] = property;
+    const styleKey = key.trim();
+    const styleValue = value.trim();
+
+    return {
+      ...prev,
+      [styleKey]: styleValue,
```
aliceblunt8, line 16:
```suggestion

    return {
      ...prev,
      [key.trim()]: value.trim(),
```

aliceblunt8 (approved):

src/convertToObject.js
dmytro-lebedchenko: thank you! Well noted! ☺

PR #1163, opened 2022-06-09
Style to object

vsupruniuk:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayOfStyles = sourceString
+    .split(';')
+    .map((string) => string.trim())
+    .filter((string) => string.length !== 0);
```
jstmpelowycz, line 20: suggestion: 
```
.filter(Boolean);
```
you can get more [here](https://michaeluloth.com/filter-boolean)

jstmpelowycz (approved):

src/convertToObject.js
vsupruniuk: Good suggestion👍

PR #1168, opened 2022-06-10
Solution

mira1zu:


AntonYegorov (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    .map((rule) => {
+      let [property, value] = rule.trim().split(': ');
+
+      property = property.trim();
+      value = value.slice(0, -1).trim();
+
+      return [property, value];
+    });
+
+  return Object.fromEntries(rules);
```
Olexandr6, line 29: You used trim() 4 times🤯 Let's improve your solution

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an example in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const rules = sourceString
+    .split('\n')
```
Olexandr6, line 18: Let's split by `;` instead, in such case you don't need to use `slice` later

Olexandr6 (changes requested):

Olexandr6 (approved):
Looks better, well done👍

PR #1173, opened 2022-06-12
sokution

vinnibea:


src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const formatedStyles = {};
+
+  let propertyArray = sourceString.replace(/[\n]/gi, '').split(';');
+
```
AndriiNychaiuk, line 7: Wow! It looks great! But this topic is about array iteration methods, not regex 😮. And even Misha said it's strongly not recomended to use it. I can approve it but I'm sure that mentors will tell you to remake it using array iteration methods.

AndriiNychaiuk (changes requested):

src/convertToObject.js
vinnibea: It`s just made code a lot simpler and in that case I thought that it would be great option to make string look in a proper way with that kind of solution ) Also I have used methods where it was necessary

AndriiNychaiuk (approved):
👍

src/convertToObject.js
illiarerun: It's okay to use `regex` here)

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const formatedStyles = {};
+
+  let propertyArray = sourceString.replace(/[\n]/gi, '').split(';');
```
illiarerun, line 6: but let's not use `let` for arrays/objects, etc. Either create a copy with consts, or just chain methods like in [checklist](https://github.com/mate-academy/js_style-to-object/blob/master/checklist.md) 4th item

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
+
+  propertyArray = propertyArray.filter(element => element.length > 2);
+
+  propertyArray.map((element) => {
+    const key = element.slice(0, element.indexOf(':')).trim();
+    const value = element
+      .slice(element.indexOf(':') + 1, element.length)
```
illiarerun, line 13: you look for `indexOf(':')` 2 times, just split string by that symbol, it'll be easier

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
+      .trim();
+
+    formatedStyles[key] = value;
```
illiarerun, line 16: `map` should return something, it's redundant without returning any values, use `forEach` here

illiarerun (changes requested):
GJ, now let's work on fixes:

jstmpelowycz (approved):

PR #1184, opened 2022-06-15
Solution

GavrilyukYura:


src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const result = {};
+  const styles = sourceString.split(';').map(str => str.trim().split(':'));
```
illiarerun, line 19: [Checklist](https://github.com/mate-academy/js_style-to-object/blob/master/checklist.md) 4th item, use proper indentation for your method calls
```suggestion
  const styles = sourceString
    .split(';')
    .map(str => str.trim()
    .split(':'));
```

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+  const stylesFilterd = styles.filter(arr => arr.length > 1);
```
illiarerun, line 20: Do you really need this copy? Maybe it would be easier to just chain this method, like you did before?

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+
+  for (const style of stylesFilterd) {
+    result[style[0].trim()] = style[1].trim();
+  }
```
illiarerun, line 24: [Checklist](https://github.com/mate-academy/js_style-to-object/blob/master/checklist.md) 5th item, you have iteration methods to do this, let's use them, I suggest you to use `reduce` to build your object

illiarerun (changes requested):
GJ, now let's work on some fixes:

PR #1185, opened 2022-06-15
solution

illavodonis:


Tsymbal-Viacheslav (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const classes = sourceString.split(';');
+  const classesCss = classes.reduce((cls, rule) => {
```
Olexandr6, line 18: It's not classes, rather properties

Olexandr6, line 18:
```suggestion
  const cssProperties = classes.reduce((cls, rule) => {
```

Olexandr6: or styles

Olexandr6, line 18:
```suggestion
  const classesCss = classes.reduce((convertedCssStyles, rule) => {
```

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    const clss = rule.split(':');
```
Olexandr6, line 19: What is this `clss`?

Olexandr6 (changes requested):
Pay attention to variable's naming

jstmpelowycz (approved):

PR #1187, opened 2022-06-15
Compleated

ant2nta:


src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedObj = {};
+
+  sourceString
+    .split(';')
+    .map(item => item.trim().split(':'))
+    .filter(item => item.length > 1)
+    .forEach(item => {
```
illiarerun, line 16: would be great to use descturcturing here to improve readability
```suggestion
    .forEach([prop, value] => {
```

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedObj = {};
```
illiarerun, line 10: also, to avoid creating object beforehand, you could've used `reduce`

illiarerun (approved):
GJ, left some comments:

PR #1191, opened 2022-06-16
Solution

LesyaSvk:


SpaceHook (approved):
GJ

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    .map(style => style.split(':'));
+
+  stylesArr.forEach(style => {
+    if (style.length === 2) {
+      const kay = style[0].trim();
+      const value = style[1].trim();
+
+      stylesObj[kay] = value;
+    }
+  });
```
Olexandr6, line 29: Actually, it would be better to use `reduce` here, instead of forEach. Let's rewrite this solution using reduce

Olexandr6 (changes requested):
Almost done👍

PR #1193, opened 2022-06-16
Solution

G0odvin:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  //  write your code here
+  const rezult = {};
+  const stringToArr = sourceString.split(';');
+  const arrSplit = stringToArr.map(item => item.split(':'));
+  const arrTrim = arrSplit.map(item => item.map(elem => elem.trim()));
+  const joinArr = arrTrim.filter(item => !item.every(items => items === ''));
```
aliceblunt8, line 22: you can use them one by one to the same variable, in chains, no need to create a new constant each time

like this arr.map().filter()

aliceblunt8 (approved):

PR #1197, opened 2022-06-17
done

k0nstrukt-or:


src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString.split(';').map(item => item.split(':'))
+    .filter(element => element.length > 1)
+    .map(property => ([property[0].trim(), property[1].trim()]));
```
illiarerun, line 19: let's use proper indents for chained methods
```suggestion
  const styles = sourceString
    .split(';')
    .map(item => item.split(':'))
    .filter(element => element.length > 1)
    .map(property => ([property[0].trim(), property[1].trim()]));
```

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+
+  return styles.reduce((previous, line) => {
```
illiarerun, line 21: you can use destructuring here to make code more readable
```suggestion
  return styles.reduce((previous, [property, value]) => {
```

illiarerun (changes requested):
GJ, few fixes left:

illiarerun (approved):

PR #1205, opened 2022-06-18
Solution

yanaost:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+ * an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArray
+    = sourceString.split(';').filter(el => el.indexOf(':') !== -1);
```
omelika, line 18: this string is long and better to use indentation as in a checklist 4

omelika (changes requested):
Good solution! only one thing from the checklist

omelika (approved):
super!

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+    .split(';')
+    .filter(el => el.indexOf(':') !== -1);
+  const finObj = {};
+
+  sourceArray.map(callback);
+
+  function callback(element) {
+    const clearElement = element
+      .replace(' ', '')
+      .replace('\n', '');
```
illiarerun, line 27: what if there are more then  1 whitespace or `\n` - line break? there is a function to `trim` such values`

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArray = sourceString
+    .split(';')
+    .filter(el => el.indexOf(':') !== -1);
+  const finObj = {};
+
+  sourceArray.map(callback);
+
+  function callback(element) {
```
illiarerun, line 24: it's weird that you're declaring callback after it's call, and do both things inside 1 function. Either move method out of your `converToObject` func or make it a function `expression` and call it after

src/convertToObject.test.js
```diff
@@ -37,7 +37,7 @@ test('Convert string to object', () => {
     .toEqual(EXPECTED_OBJECT);
 });
 
-test('Object should have same count of keys as styles in css', () => {
+test.only('Object should have same count of keys as styles in css', () => {
```
illiarerun, line 40: Don't change tests
```suggestion
test('Object should have same count of keys as styles in css', () => {
```

illiarerun (changes requested):
GJ, let's work on fixes:

src/convertToObject.test.js
yanaost: @illiarerun just forgot to delete(

src/convertToObject.js
yanaost: @illiarerun from mozilla.org 
[Function declaration hoisting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function#function_declaration_hoisting)
Function declarations in JavaScript are hoisted to the top of the enclosing function or global scope. You can use the function before you declared it:

PR #1209, opened 2022-06-18
solution was added

Bielichenko:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  let arr = sourceString.split(/[:;\n]/);
+
+  arr = arr.filter(item => item && item !== '  ');
+
+  for (let i = 0; i < arr.length; i = i + 2) {
```
illiarerun, line 23: [Checklist](https://github.com/mate-academy/js_style-to-object/blob/master/checklist.md) 5th item, you have iteration methods to do this

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
```
illiarerun, line 18: if you use `reduce` with initial value of `{}`, you won't need obj beforehand, also `obj` is a bad name for a variable

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+  let arr = sourceString.split(/[:;\n]/);
```
illiarerun, line 19: why is it `let`? you just change it with a `filter` which is non-mutating method, make it a const and you can actually chain methods together - look at [checklist](https://github.com/mate-academy/js_style-to-object/blob/master/checklist.md) 4th item to find out how

illiarerun: also, `arr` is a bad name for a variable

illiarerun (changes requested):
GJ, now let's work on fixes:

Olexandr6 (approved):

PR #1214, opened 2022-06-19
solution

AKomashchuk:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+  const stylesAndProp = sourceString
+    .split(';')
+    .map(item => item.split(': '));
+
+  const reducer = stylesAndProp
+    .filter(item => item.length > 1)
+    .reduce((acc, item) => {
+      const key = item[0].replace('\n', '').trim();
+      const values = item
+        .splice(1).map((value) => value.replace('\n', '').trim());
```
illiarerun, line 26: why don't you just do the same thins as you did with the key?
```suggestion
      const key = item[0].replace('\n', '').trim();
      const value = item[1].replace('\n', '').trim();
```

illiarerun (changes requested):
GJ, now let's work on fixes:

jstmpelowycz (approved):

PR #1218, opened 2022-06-20
implemented convertObject function

omelika:


yanaost (approved):
nice) optionally you could use map to to gather final object

illiarerun (approved):
GJ

PR #1219, opened 2022-06-20
Solution

dimabelotskyi:


ant2nta (approved):

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayOfStyles = sourceString.split(';')
+    .map(str => str.trim().split(':'))
+    .filter(arr => arr.length > 1);
```
illiarerun, line 12: lest improve readability abit
```suggestion
  const arrayOfStyles = sourceString
    .split(';')
    .map(str => str.trim().split(':'))
    .filter(arr => arr.length > 1);
```

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
+
+  const objectOfStyles = {};
+
+  for (const style of arrayOfStyles) {
+    objectOfStyles[style[0].trim()] = style[1].trim();
+  }
```
illiarerun, line 18: let's use array iteration methods instead, try to od it with reduce with an initial value of `{}`

illiarerun (changes requested):
GJ, let's work on fixes now:

illiarerun (approved):

PR #1239, opened 2022-07-05
Solution

alexbardyshev:


src/convertToObject.js
```diff
@@ -1,18 +1,20 @@
...
+    .split(';')
+    .map(el => el.trim().split(':'))
+    .filter(arr => arr.length > 1);
+
+  for (const style of styleArr) {
+    const key = style[0].trim();
+    const value = style[1].trim();
+
+    cssProperties[key] = value;
+  }
```
solaryasha, line 15: Let's practice `reduce` method (create object and add specific `key`- `value` pairs)

src/convertToObject.js
```diff
@@ -1,18 +1,20 @@
...
-  // write your code here
+  const cssProperties = {};
+  const styleArr = sourceString
+    .split(';')
+    .map(el => el.trim().split(':'))
+    .filter(arr => arr.length > 1);
+
+  for (const style of styleArr) {
+    const key = style[0].trim();
+    const value = style[1].trim();
```
solaryasha, line 12: Use destructuring for these values

solaryasha (changes requested):

src/convertToObject.js
```diff
@@ -1,18 +1,17 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const callback = (prev, elem) => ({
```
VETALEK, line 4:
```suggestion
  const getStyle = (prev, elem) => ({
```

`callback` is not descriptive name

VETALEK (changes requested):
GJ 👍

VETALEK (approved):
🔥

PR #1242, opened 2022-07-07
Solution

Sergii2705:


src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split('\n')
```
VETALEK, line 17: better split by `;`,
now you can't parse inline styles

image

VETALEK (changes requested):
Good start 🚀

src/convertToObject.js
VETALEK: yeah, but splitting by `\n` is totally unreliable, cause it is just formatting, and your function shouldn't depend on whether or not I formatted my styles correctly

VETALEK (approved):
🚀

PR #1244, opened 2022-07-07
Solution

frontend112:


meforoma (approved):
really smart 🚀 

the only thing I'd mention is instead of `.map` better use `.forEach` loop, - just because the loop's purpose is to iterate, while `.map` returns a new array.
So might simply be confusing for another developer 🙂

PR #1248, opened 2022-07-10
solution

ClaudiaKalnitskaya:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  const resultArr = sourceString.split(';')
+    .map(element => element.split(':'))
+    .map(element => element.map(value => value.trim()))
```
```suggestion
    .map(element => element.split(':').map(value => value.trim()))
```
You can map only once

BudnikOleksii (approved):
:+1:

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+    .filter(element => element.length > 1);
```
ivan-kadykalo, line 21: It is a possible solution, but, you start loop 4 times. Do you really need it? 
You can make all operations from lines 19 to 21 in your forEach loop.

ivan-kadykalo (changes requested):

PR #1250, opened 2022-07-10
Solution

TwoAndMore:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedString = {};
+  const styles = sourceString
+    .split(';')
+    .map(i => i.trim())
```
DashaPoliazz, line 20: Its better to replace `i` with semantic name

DashaPoliazz (approved):

src/convertToObject.js
ivan-kadykalo: Yep. It would be more understandable.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    .filter(i => i.length > 0);
```
ivan-kadykalo, line 21: here the same

ivan-kadykalo (changes requested):

ivan-kadykalo (approved):
GJ💪

PR #1254, opened 2022-07-12
Solution

LeonidPustovit:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+  const convertedStyles = {};
+
+  sourceString
+    .split(';')
+    .map(element => element.trim())
+    .filter(element => element.includes(':'))
+    .map(property => {
+      return [property.split(':')[0].trim(),
+        property.split(':')[1].trim()];
+    })
```
Ihor-Karpyn, line 26: better to remove return, and move this logic to some function

Ihor-Karpyn (approved):
Good job

PR #1259, opened 2022-07-13
done

AntonChetverykMA:


src/convertToObject.js
```diff
@@ -12,7 +13,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+  const arr = sourceString.split(';');
+  const arr1 = arr.map(e => e.trim().split(':'));
+
+  for (const prop of arr1) {
```
alexandra-protyanova, line 20: you can use forEach instead of loop

PR #1261, opened 2022-07-13
solution

denys-danyliuk:


src/convertToObject.js
```diff
@@ -1,18 +1,31 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const callback = (prev, elem ) => ({
+      ...prev, [elem[0].trim()]: elem[1].trim(),
+  });
```
ivan-kadykalo, line 20: Move this callback outside of the convertToObject function, to avoid creating it on each call.

src/convertToObject.js
```diff
@@ -1,18 +1,31 @@
...
+
+  const properties = sourceString
+    .split(';')
+    .map(el => el.trim().split(':'))
+    .filter(arr => arr.length > 1)
+    .reduce(callback, {});
+    
+  return properties;
```
ivan-kadykalo, line 28:
```suggestion
  return sourceString
    .split(';')
    .map(el => el.trim().split(':'))
    .filter(arr => arr.length > 1)
    .reduce(callback, {});
```

ivan-kadykalo (changes requested):
Good start👍

ivan-kadykalo (approved):
🔥

PR #1263, opened 2022-07-13
solution

AleksandrYuschenko:


src/convertToObject.js
```diff
@@ -2,17 +2,42 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const sourceArray = sourceString
+    .split('\n')
+    .filter(x => x !== '' && x !== '  ;')
+    .join('').split(';').filter(x => x !== '');
+
+  const test1 = [];
```
PavloMarkov, line 22: names of arrays are not understandable. 
maybe you should name them like "keyValueStylesWithSpaces" & "keyValueStyles" or something else.

src/convertToObject.js
```diff
@@ -2,17 +2,42 @@
...
+  const test2 = [];
+  const result = {};
+
+  sourceArray.forEach(element => {
+    test1.push(element.trim().split(':'));
```
PavloMarkov, line 27: maybe the work of this trim() method will do trim() in test2 - 'test2.push(key.trim());'

PavloMarkov (approved):
good job! 🚀 🔥 
have some comments to correct

src/convertToObject.js
Ihor-Karpyn: agree, fix it pls

src/convertToObject.js
```diff
@@ -2,17 +2,42 @@
...
+
+  test1.forEach(element => {
+    element.forEach(key => {
+      test2.push(key.trim());
+    });
+  });
+
+  for (let i = 0; i < test2.length; i += 2) {
+    result[test2[i]] = test2[i + 1];
+  }
```
Ihor-Karpyn, line 38: Let's optimize this solution, and don't use loops here

Ihor-Karpyn (changes requested):

src/convertToObject.js
```diff
@@ -2,17 +2,28 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const result = {};
+
+  sourceString.split(';\n')
```
solaryasha, line 19: Let's practice `reduce` method

AleksandrYuschenko: Я не маю взагалі фантазіі куди той reduce тут можна засунути. Sorry, I can't fix that

VETALEK (approved):
GJ 🚀 
but it is better to split by `;`, not `;\n`

PR #1264, opened 2022-07-13
Solutioin

yaroslav-lizogub:


oleksandr-kotliarov (approved):
an interesting solution, but hard to understand by looking at it

PR #1275, opened 2022-07-14
Convert to object func was created

yuliia-dronova:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArray = sourceString.split(';')
+    .map(element => element.trim().split(':'))
+    .filter(arr => arr.length > 1);
+
+  const callback = (acc, element) =>
+    ({
+      ...acc, [element[0].trim()]: element[1].trim(),
+    });
```
ivan-kadykalo, line 24: If it is a callback function, will be better to move it outside of your convertToObject function, to avoid creating it on each call.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+
+  const style = sourceArray.reduce(callback, {});
+
+  return style;
```
ivan-kadykalo, line 28:
```suggestion
  return sourceArray.reduce(callback, {});
```

ivan-kadykalo (changes requested):

ivan-kadykalo (approved):
GJ🚀

PR #1276, opened 2022-07-14
solution

h4p4l4:


oleksandr-kotliarov (approved):
good

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let styles = sourceString.split(';').map(string => string.trim())
+    .filter(string => string !== '');
+
+  styles = styles.map(style => style.split(':')
```
ivan-kadykalo, line 20: Why do you need to rewrite this let?
You can create separate const for each value.
You should use const variables when it is possible.
```suggestion
  const styles = sourceString.split(';').map(string => string.trim())
    .filter(string => string !== '');

  const splitedStyle = styles.map(style => style.split(':')
```
for example.

ivan-kadykalo (approved):
🚀

PR #1277, opened 2022-07-14
Solution

kata96451:


oleksandr-kotliarov (approved):
good but so hard to understand

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .map(style => style
+      .trim()
```
ivan-kadykalo, line 13: Perhaps you don't need this trim because you trim every element in your map, but I'm not sure, need to test.

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
+      .split(':')
+      .map(el => el.trim())
```
ivan-kadykalo, line 15: Use more informative names, not el

ivan-kadykalo (approved):

PR #1297, opened 2022-07-24
solution

oleksiy-uchaev:


src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .map((item) => item.split(':').map(i => i.trim()))
+    .filter(keyValArr => keyValArr.length === 2)
+    .sort((a, b) => a[0].localeCompare(b[0]))
```
alexandra-protyanova, line 21: why do u need to sort them?

alexandra-protyanova (changes requested):

src/convertToObject.js
oleksiy-uchaev: Thanks!

alexandra-protyanova (approved):

PR #1307, opened 2022-08-03
Solution style-to-obgect

Serhii-Vovk:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayStyles = sourceString.split(';')
+    .map(property => property.split(':'));
```
Anna-Tkachenko, line 18: When you use several methods in chain, each of them should be moved to the new line
```suggestion
  const arrayStyles = sourceString
    .split(';')
    .map(property => property.split(':'));
```

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  const deleteGap = arrayStyles.map(items =>
+    items.map(element => element.trim())).filter(item => item.length > 1);
```
Anna-Tkachenko, line 21: Same here, just fix code style

Anna-Tkachenko (changes requested):

alexandra-protyanova (approved):

PR #1313, opened 2022-08-08
task js_style-to-object

Moblat890:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
-  // write your code here
+  const result = {};
+  const splitString = sourceString.split(';').map(elem => elem.split(':'));
+
+  const delEpmty = splitString.filter(elem => elem.length !== 1);
+
+  const delSpaces = delEpmty.map(elem => elem.map(item => item.trim()));
+
+  delSpaces.map(elem => {
+    result[elem[0]] = elem[1];
```
alexandra-protyanova, line 25: u can trim elements here

alexandra-protyanova (approved):

src/convertToObject.js
Moblat890: I don't understand how to implement it and why it can't be like I wrote

alexandra-protyanova: It is an option actually, I approved your solution, but it would make complexity of your solution easier

alexandra-protyanova (approved):

PR #1320, opened 2022-08-11
solution

AndreyJigijalkin:


src/convertToObject.js
```diff
@@ -1,18 +1,17 @@
...
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const entriesArray = sourceString
+    .split(';').map((a) => a.split(':')
+      .map((b) => b.trim())).filter((c) => c.length === 2);
```
VolodymyrChuyko, line 6:
```suggestion
    .split(';')
    .map((a) => a.split(':').map((b) => b.trim()))
    .filter((c) => c.length === 2);
```
I offer to format your code like this, to make it more understandable

src/convertToObject.js
```diff
@@ -1,18 +1,17 @@
...
+
+  const result = {};
+
+  for (let i = 0; i < entriesArray.length; i++) {
+    result[entriesArray[i][0]] = entriesArray[i][1];
+  }
```
VolodymyrChuyko, line 12: it is possible to use a method here instead of a loop

VolodymyrChuyko (approved):
looks good to me!

AlinaMatuschak (approved):
Awesome 👏

PR #1321, opened 2022-08-11
Solution

dmytro-legerda:


src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const sourceArray = sourceString.split(';');
+  const styleArray = sourceArray.map(x => x.split(':'));
+
+  const styleArray2 = styleArray.map(function(y) {
```
alexandra-protyanova, line 21: it is better to use arrow function here

alexandra-protyanova (changes requested):

marina-tilniak (approved):

PR #1331, opened 2022-08-15
solution

Githi54:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const changes = sourceString;
+
+  changes.split(';')
+    .filter(elem => elem.trim().length !== 0)
+    .map(elem => {
```
AlinaMatuschak, line 22: Bad way to use the map method. The map should use if you need to create `new array` with the results of calling a provided function on every element in the calling array. If you need to do something for each element on the array, but you don’t expect new array, should use forEach

AlinaMatuschak (changes requested):
Good start 👏

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  sourceString.split(';')
+    .filter(elem => elem.trim().length)
```
AlinaMatuschak, line 20: Checklist 

```suggestion
  sourceString
    .split(';')
    .filter(elem => elem.trim().length)
```

AlinaMatuschak, line 20: Element is not elem name 💩

AlinaMatuschak (changes requested):
Almost perfect 👍

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+ * and values are the values of related CSS
+ * properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
AlinaMatuschak, line 17: If the mentor leaves some comments, check all file, maybe you have similar issues.
All variable names should be with a logical name. `Result` doesn’t describe what exactly this variable should have on value

AlinaMatuschak (approved):
Awesome 🤩 . But be more attentive to mentor comments )

PR #1333, opened 2022-08-15
Solution

Andrii-Stiebalov:


Dmytro-Matsehora (approved):
well done! 👍

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+
+  const stylesObj = arrOfString.reduce((prev, str) => {
+    const arr = str.split(':');
+    const [property, value] = arr.map(elem => elem.trim());
+
+    return {
+      ...prev,
+      [property]: value,
+    };
+  }, {});
```
dellachaise, line 32: You could also try to do the same using `forEach`, it would be shorter.

dellachaise (approved):
Great job 👍 

With unusual `reduce` usage  🔥

PR #1336, opened 2022-08-16
solution

yungdena:


Anton-Liada (approved):
👍

src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const css = {};
```
dellachaise, line 17: Better rename to `cssObject` for example)

src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
+
+  sourceString.split(';')
+    .filter((el) => el.trim())
```
dellachaise, line 20: Point 4 from checklist.
```suggestion
  sourceString
    .split(';')
    .filter((el) => el.trim())
```

dellachaise, line 20: `el` it is not a good name for variable all over the code. Try `currentStyleElement`, `currentStyKey` below. Name should describe what is inside this vsriable.

dellachaise (changes requested):
Good solution 👍 

Just add a few little fixes according to checklist and comments 👇

dellachaise (approved):
Great job 👍

PR #1338, opened 2022-08-16
Solution

Jamal-Amoian:


src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssStyles = {};
+
+  const redactedString = sourceString
+    .split(';')
+    .filter(s => s.trim().length > 0)
```
sergii-nosachenko, line 13:
```suggestion
    .filter(s => s.trim().length)
```
If `s.trim().length === 0` it's already false value. So other cases are true.

src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
+    .join(':')
+    .split(':')
+    .map(s => s.trim());
+
+  for (let i = 0; i < redactedString.length; i += 2) {
+    cssStyles[redactedString[i]] = redactedString[i + 1];
+  }
```
```suggestion
    .forEach(prop => {
      const [ propName, propValue ] = prop.split(':');
      cssStyles[propName.trim()] = propValue.trim();
    });
```

Array has nice iteration method `forEach`. Use it here, cause you need to iterate through filtered lines.
It can optimize your code.

sergii-nosachenko (changes requested):
Code can be optimized.

sergii-nosachenko (approved):
Now it looks better, good ✌

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssStyles = {};
+
+  sourceString
+    .split(';')
+    .filter(s => s.trim().length)
+    .forEach(prop => {
+      const [ propName, propValue ] = prop.split(':');
```
AlinaMatuschak, line 15: 😍

AlinaMatuschak (changes requested):
Awesome. Just fix `prop` name

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssStyles = {};
+
+  sourceString
+    .split(';')
+    .filter(s => s.trim().length)
```
AlinaMatuschak, line 13: s isn’t better than prop 😂

AlinaMatuschak (changes requested):
If I leave some comments, check all code (maybe you have similar issues)

AlinaMatuschak (approved):
🚀🚀🚀

PR #1339, opened 2022-08-17
solution

artem-kodesnikov:


OlehMy (approved):
Good work)

src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sortedObject = {};
+
+  sourceString.split(';')
+    .filter((item) => item.trim())
```
AlinaMatuschak, line 20: Checklist
```suggestion
  sourceString
    .split(';')
    .filter((item) => item.trim())
```

AlinaMatuschak, line 20: item is not best name 🥲

AlinaMatuschak (changes requested):
Almost perfect 👍

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const sortedObject = {};
+
+  sourceString
+    .split(';')
+    .filter((splitedRule) => splitedRule.trim())
+    .map((splitedPropeties) => splitedPropeties.split(':'))
+    .forEach((properties) =>
+      (sortedObject[properties[0].trim()] = properties[1].trim()));
```
AlinaMatuschak, line 24: you can use reduce to create style object that you need

AlinaMatuschak (approved):
Nice, but will be awesome if you try to use reduce to practice )

PR #1340, opened 2022-08-17
Solution

Mejeezz:


oksana-logos-frontend (approved):
Great🔥

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectOfStyles = {};
+
+  sourceString.split(';')
+    .filter(element => element.trim())
```
AlinaMatuschak, line 20: Element is not best name 🥲

AlinaMatuschak (approved):
Good work 👏

src/convertToObject.js
Mejeezz: got it! i'll try to name vars more understandable)

PR #1341, opened 2022-08-17
Solution style to object

PastolNapas:


OleksiiPustovit (approved):
Good job 🔥

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const arrSourceString = sourceString.split(';');
+  const result = {};
```
dellachaise, line 19: Better create meaningful variables, for example `stylesObject`.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  arrSourceString.forEach(style => {
+    const item = style.split(':');
```
dellachaise, line 22: Here the same, after reading variable name you should understand what is inside. Here could be `currentStyleKey`.

dellachaise (changes requested):
Great solution 👍 

Please create meaningful variables 👇

src/convertToObject.js
dellachaise: Still actual to rename this variable too)

dellachaise (approved):

PR #1344, opened 2022-08-17
Solution

OlehMy:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+  const stylesArray = sourceString
+    .split(';').map(style => style.trim())
+    .filter(style => style)
```
artem-kodesnikov, line 21:
```suggestion
    .split(';')
    .map(style => style.trim())
    .filter(style => style)
```
I think this will be more readable if you put new method on the new line

artem-kodesnikov (approved):
good job, but need a few fixes)

dellachaise (approved):
Well done 👍

PR #1346, opened 2022-08-17
solution

anyakrainiuk:


YuriiPronin (approved):
Good job!

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objCss = {};
+
+  const stylesArr = sourceString.split(';');
+
+  stylesArr.forEach(str => {
```
AlinaMatuschak, line 21: What is str?
You can use `styleRule` or something like that. Name should describe what is it

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    const currenValue = str.split(':');
```
AlinaMatuschak, line 22:
```suggestion
    const splitedRule = str.split(':');
```

src/convertToObject.js
AlinaMatuschak, line 21: you can do it shorter, but it's your choice )

```suggestion
  sourceString
    .split(';')
    .forEach(str => {
```

AlinaMatuschak (changes requested):
Good work. Just fix naming )

src/convertToObject.js
anyakrainiuk: I did it for more understandable for me)

src/convertToObject.js
anyakrainiuk: changed!)

AlinaMatuschak (approved):
Awesome 👏

PR #1347, opened 2022-08-17
Done

andrii-dmitruk:


Anastasiia-Khmelovska (approved):

src/convertToObject.js
```diff
@@ -11,8 +12,31 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const array = sourceString.split(';');
+  const formattedArray = array.map(row => (row
+    .replace(/\n/g, '')
+    .trim()
+    .split(':')
+  ));
+
+  const formattedObject = formattedArray.reduce((prev, current) => {
```
AlinaMatuschak, line 24:
```suggestion
  const formattedObject = formattedArray.reduce((accumulate, current) => {
```
or
```suggestion
  return formattedArray.reduce((formattedObject, current) => {
```

AlinaMatuschak, line 24: reduce 😍

src/convertToObject.js
```diff
@@ -11,8 +12,31 @@
...
+    if (current[0].length !== 0 && current[1].length !== 0) {
```
AlinaMatuschak, line 25: 0 converts to false
```suggestion
    if (current[0].length && current[1].length) {
```

AlinaMatuschak (approved):
Good work )

PR #1348, opened 2022-08-17
solutionConvertStyles

YuriiPronin:


anyakrainiuk (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const firstArr = sourceString.split(/[;:]/);
+  const secondArr = firstArr.map(x => x.trim());
+  const result = {};
+
+  for (let i = 0; i < secondArr.length; i += 2) {
```
dellachaise, line 21: Try to use `forEach or reduce` instead of `for`.

Also you could do trim inside `forEach`.

dellachaise (changes requested):
Good job 👍 

But better do not use `for here.

danylo-onishchuk (approved):

PR #1352, opened 2022-08-17
Solution-preview

plemiannik:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceObject = {};
+
+  const sourceArr = sourceString
+    .replace(/\n/g, '')
+    .replace(/ +/, '')
+    .replace(/  +/g, '')
```
ShmelyovOleg, line 22: not sure, maybe you can replace it with one regex expression

ShmelyovOleg (approved):

src/convertToObject.js
plemiannik: Think so 2

dellachaise: Try to use 1 regex expression instead of 3. Or did `trim` inside if `forEach` instead of this regex expressions.

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+    .split(';')
+    .filter(a => a.length > 0)
+    .map(a => a.split(':'));
+
+  for (const i of sourceArr) {
```
dellachaise, line 27: Better use `forEach` instead if `for`.

dellachaise (changes requested):
Good job 👍 

Just add fixes according to comments 👇

dellachaise (approved):
Great job 👍 

Just add changes to variables names 👇

PR #1354, opened 2022-08-18
Solution

Ohiiko:


NataliyaAn (approved):
brevity smart solution. good job)

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObj = {};
+
+  sourceString.split(';')
+    .filter((item) => item.trim().length)
```
dellachaise, line 20:
```suggestion
  sourceString
    .split(';')
    .filter((item) => item.trim().length)
```

dellachaise (changes requested):
Great solution 👍 

Just fix a one little issue according to checklist 👇

dellachaise (approved):
Well done 👍

PR #1357, opened 2022-08-18
implemented function convertToObject

ShmelyovOleg:


AndreyJigijalkin (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,24 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesCSS = {};
+  const arr = sourceString.split(';').filter(item =>
+    item.trim()).map(item => item.split(':'));
+
+  arr.forEach(item => (stylesCSS[item[0].trim()] = item[1].trim()));
```
dellachaise, line 21: Add changes according to 2 and 4 points from [checklist](https://github.com/mate-academy/js_style-to-object/blob/master/checklist.md)

dellachaise (changes requested):
Good solution 👍 

But please read [checklist](https://github.com/mate-academy/js_style-to-object/blob/master/checklist.md) attentively 😉

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesCSS = {};
+  const arrayOfStyles = sourceString
+    .split(';')
+    .filter(item => item.trim())
+    .map(item => item.split(':'));
+
+  arrayOfStyles.forEach(item => (stylesCSS[item[0].trim()] = item[1].trim()));
```
dellachaise, line 23: Create meaningful variables, `item` here - is not a good option. Variable name should describe what inside it.  `stylesString` for example would be better.

dellachaise (approved):
Well done 👍 

Just please create more meaningful variables 👇

PR #1361, opened 2022-08-18
Solution

anna-kolmakova:


Alzay007 (approved):

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const cssStyleObject = {};
+
+  sourceString
+    .split(';')
+    .filter(value => value.trim())
+    .map(value => value.split(':'))
+    .forEach(value => (
+      cssStyleObject[value[0].trim()] = value[1].trim()));
```
AlinaMatuschak, line 17: you can use reduce to create style object that you need

AlinaMatuschak (approved):
Nice, but will be awesome if you try to use reduce to practice )

PR #1363, opened 2022-08-18
Solution

SergeyPistelev:


ibondrk (approved):

ibondrk:
Good job!

src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+
+  sourceString
+    .split(';')
+    .filter(string => string.trim())
+    .map(string => string.split(':'))
+    .forEach(string => (styleObject[string[0].trim()] = string[1].trim()));
```
dellachaise, line 22: Create meaningful variables, `string` here - is not a good option. Variable name should describe what inside it.  `styleString` for e[ample would be better.

dellachaise (changes requested):
Great solution 👍 

Just add changes to naming 👇

src/convertToObject.js
SergeyPistelev: Variable names changed.
Thanks for review)

dellachaise (approved):
Well done 👍

PR #1364, opened 2022-08-18
Version 0.1

David-Shahmatov:


wfurs0w (approved):
Cool 🔥

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
-}
+  const splitString = sourceString.split(';');
+  const stylesObject = {};
+
+  splitString.forEach(style => {
+    const item = style.split(':');
```
dellachaise, line 22: Create meaningful variables, `item` here - is not a good option. Variable name should describe what inside it.  `stylesString` for example would be better.

dellachaise (changes requested):
Great solution 👍 

But please create more meaningful variables 👇

dellachaise (approved):
Well done 👍

PR #1365, opened 2022-08-18
Solution

ghost:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    .map(item => {
+      cssStyles[item[0].trim()] = item[1].trim();
+    });
 
+  return cssStyles;
+}
 module.exports = convertToObject;
+// for(var key in input){
+//   output[key.replace(/\s+/,'_')] = input[key];
+// }
```
Alej4ndro1, line 31: You need to delete comments

Alej4ndro1 (approved):
Good but delete comments

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const cssStyles = {};
+
+  sourceString.split(';')
+    .filter(item => item.trim())
+    .map(item => item.split(':'))
+    .map(item => {
```
AndriiRut, line 22: item is the synonym of element, and makes the code less readable. use such names as style, properties, values, string and etc)

checklist(NAMING)

AndriiRut (changes requested):

AndriiRut (approved):
🚀

PR #1367, opened 2022-08-18
function to create Object of style string was created

ibondrk:
Function to create Objects is ready

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -1,18 +1,23 @@
 'use strict';
 
-/**
- * Implement convertToObject function:
- *
- * Function takes a string with styles (see an example in [stylesString.js](./stylesString.js))
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const sortedSourceArray = sourceString
+    .replace(/\s+/g, ' ')
+    .trim()
+    .split(';')
+    .map(splitProperty => splitProperty.split(':'))
+    .filter(cleanUp => cleanUp.length > 1);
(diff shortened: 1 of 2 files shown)


SergeyPistelev (approved):
Interesting solution!
Good job)

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const sortedSourceArray = sourceString
+    .replace(/\s+/g, ' ')
+    .trim()
+    .split(';')
+    .map(splitProperty => splitProperty.split(':'))
+    .filter(cleanUp => cleanUp.length > 1);
```
AlinaMatuschak, line 9: you can split the property and check if you have the property name and value on your reduce callback. There will be -2 loops

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
+  const createObject = (prev, [ property, value ]) => {
+    return {
+      ...prev,
+      [property.trim()]: value.trim(),
+    };
+  };
+
+  const sourceObject = sortedSourceArray.reduce(createObject, {});
+
+  return sourceObject;
```
AlinaMatuschak, line 20:
```suggestion
  return sortedSourceArray.reduce(createObject, {});
```

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
+    .filter(cleanUp => cleanUp.length > 1);
+
+  const createObject = (prev, [ property, value ]) => {
+    return {
+      ...prev,
+      [property.trim()]: value.trim(),
+    };
+  };
+
+  const sourceObject = sortedSourceArray.reduce(createObject, {});
```
AlinaMatuschak, line 18: reduce 😍

AlinaMatuschak (approved):
🚀

PR #1369, opened 2022-08-18
task solution code

rboiar:


Roman-Artiukh (approved):
change your code style at 20 row according checklist

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedStyles = {};
+
+  sourceString.split(';')
+    .filter(item => item.trim().length)
```
AlinaMatuschak, line 21: Checklist 

```suggestion
  sourceString
    .split(';')
    .filter(item => item.trim().length)
```

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    .forEach(item => {
+      const element = item.split(':');
```
AlinaMatuschak, line 23: What is an element? By name, I don’t understand that it's an array with splited rule

AlinaMatuschak (approved):
Good work ) But I believe that you fix my issues 👇

PR #1370, opened 2022-08-19
Solution

vadym-zahazei:


plemiannik:
u forgot about checklist birdy

plemiannik (approved):
👍

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resultObject = {};
+
+  sourceString.split(';')
+    .filter(property => property.trim())
```
dellachaise, line 20:
```suggestion
  sourceString
    .split(';')
    .filter(property => property.trim())
```

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+    .map(element => element.split(':'))
+    .map(element => {
```
dellachaise, line 22: Create meaningful variables, `item` - is not a good option. Variable name should describe what inside it.  `stylesString` for example would be better.

dellachaise (changes requested):
Great job 👍  

Just please create more meaningful variables 👇

danylo-onishchuk (approved):

PR #1371, opened 2022-08-19
convert to object

oksana-kyryienko:


roma-shafihullin (approved):
good solution 👍

src/convertToObject.js
```diff
@@ -12,7 +12,17 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  // write your code here'
+  const obj = {};
+
+  sourceString
+    .split(';')
+    .filter(string => string.trim())
+    .map(string => string.split(':'))
+    .forEach(string => (obj[string[0].trim()]
```
AndriiRut, line 22: is var `string` on the 22th line contain a string or some other type of data?

AndriiRut: Checklist # 2

AndriiRut (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+  const obj = {};
+
+  sourceString
+    .split(';')
+    .filter(item => item.trim())
+    .map(item => item.split(':'))
+    .forEach(item => (obj[item[0].trim()]
+    = item[1].trim()));
+
+  return obj;
```
AndriiRut, line 27:
```suggestion
  sourceString
    .split(';')
    .filter(string => string.trim()) // what callback in  the method filter should return? maybe map or forEach fit better?
    .map(string => string.split(':')) 
    .forEach(item => (obj[item[0].trim()] = item[1].trim()));

  return obj;
```

AndriiRut (changes requested):
Good! But try to use more understendable naming like 
`  sourceString
    .split(';')
    .filter(string => item.trim())
    .map(item => item.split(':'))
    .forEach(item => (obj[item[0].trim()]
    = item[1].trim()));`

AndriiRut (approved):
Oh! I see. Using of filter with trim will return false for empty string! Ok then! 
Good job. 🚀

PR #1372, opened 2022-08-19
Solution

Roman-Artiukh:


rboiar (approved):
A very clear and good solution to this task!

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssStyles = {};
+  let key = '';
+  let value = '';
+
+  sourceString.split(';')
+    .filter(style => style.trim())
+    .map(style => style.split(':'))
```
AlinaMatuschak, line 23: Checklist 
```suggestion
  sourceString
    .split(';')
    .filter(style => style.trim())
    .map(style => style.split(':'))
```

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    .forEach(style => {
```
AlinaMatuschak, line 24: you can use reduce.  It will be just one loop instead of 3 )

AlinaMatuschak (approved):
GJ 🔥

src/convertToObject.js
Roman-Artiukh: ok, thank you))

PR #1378, opened 2022-08-19
task complette

Sasha-Gordijchuk:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+
+  sourceString.split(';')
+    .map(style => style.trim())
+    .filter(style => style)
+    .map(style => style.split(':')
+      .map(property => property.trim()))
```
vladgusakovskiy, line 23: Just a small mistake) Pay attention to criteria num 4 in code style

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    .forEach(style => {
+      styleObject[style[0]] = style[1];
```
vladgusakovskiy, line 25: It could be possible to avoid using **forEach** method. You can transfer line 25 to your map (line 23). Just like this **styleObject[style[0].trim()] = style[1].trim();** But it's all up to you

vladgusakovskiy (approved):
Everything is fine! You just should correct that small mistake in code style)

src/convertToObject.js
danylo-onishchuk: agree

danylo-onishchuk (approved):

PR #1382, opened 2022-08-20
Style to Object done

Andrii-Dunets:


oksana-logos-frontend (approved):
this can be solved more easily without additional variables, you can use trim() with elements in your first filter, sow your first map would be useless...

danylo-onishchuk (approved):

PR #1387, opened 2022-08-21
js style to object

erikkaraman:


max-kravchenko (approved):
Great solution!

src/convertToObject.js
```diff
@@ -1,18 +1,18 @@
...
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  sourceString
+    .split(';')
+    .filter(cssString => cssString.trim().length)
```
AndriiRut, line 8: if you got empty string after `cssString.trim()` `Undefined` will convert in false anyway. You don't need to get the length of a string so the `0 length` converts into false.

AndriiRut (approved):
🔥

PR #1396, opened 2022-08-26
style to object done

sanyokmalyshev:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+    const key = elem
+      .slice(0, elem.indexOf(':'))
+      .trim()
+      .replace(/\s+/g, '');
+    const value = elem
+      .slice(elem.indexOf(':') + 1)
+      .trim();
+
+    cssObject[key] = value;
+  });
```
stanislav-topikha, line 33: `map` creates new array, in this case you don't need it, use forEach

stanislav-topikha (changes requested):
comment ⬇️

PR #1399, opened 2022-08-29
solution

vitalii-babiienko:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
-  // write your code here
+  const result = {};
+
+  sourceString
+    .split(';')
+    .filter(styleString => styleString.trim())
+    .forEach(styleString => {
+      const styleArray = styleString.split(':');
+
+      result[styleArray[0].trim()] = styleArray[1].trim();
```
stanislav-topikha, line 25:
```suggestion
      const styleProp = styleArray[0].trim();
      const styleValue = styleArray[1].trim()
      
      result[styleProp] = styleValue;
```

stanislav-topikha (changes requested):
let's increase readability a bit

stanislav-topikha (approved):

PR #1401, opened 2022-08-31
Solution

mariabramnik:


src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+
+  const resObj = sourceString.replace(/\s+/g, ' ')
+    .split(';').filter(function(entry) {
+      return /\S/.test(entry);
```
stanislav-topikha, line 16: cmon, don't just copy and paste
what is entry in this case, name it properly
use arrow function
or just rewrite in more simple and readeble way

Screenshot 2022-09-01 at 15 45 11

stanislav-topikha (changes requested):
comment ⬇️

src/convertToObject.js
```diff
@@ -3,16 +3,24 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
+
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
```
liliia-livandovskaya, line 12: remove comments

liliia-livandovskaya (approved):
👍

PR #1403, opened 2022-09-02
Solution

AMahalias:


src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const sourceArray = sourceString.split(';').filter(element => element.trim());
+  const cssProperty = sourceArray.map(item => item.split(':'));
+
+  cssProperty.map((property) =>
+    (result[property[0].trim()] = property[1].trim()));
```
stanislav-topikha, line 22: you don't use returned arr, use `forEach` or simple loop

stanislav-topikha (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
-  // write your code here
+  const result = {};
+  const sourceArray = sourceString.split(';').filter(element => element.trim())
+    .map(item => item.split(':'));
+
+  sourceArray.forEach((property) =>
+    (result[property[0].trim()] = property[1].trim()));
+
+  return result;
 }
```
SanyaBratashchuk, line 25: I guess, you can create resultObject using reduce

SanyaBratashchuk (changes requested):

qoa11a (approved):

PR #1409, opened 2022-09-08
Solution

illia-kots:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleArray1 = sourceString.split(';');
```
alexandra-protyanova, line 17: not a good idea to use numbers in variables

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+  const styleArray2 = styleArray1.map(el => el.split(':'));
+  let styleArray3 = [];
+  const styleArray4 = styleArray2.map(el => {
+    if (el.length > 1) {
+      styleArray3 = el.map(y => y.trim());
+    }
+
+    return styleArray3;
+  });
```
alexandra-protyanova, line 26: all these code can be moved to reduce, u only trim and checking length

alexandra-protyanova (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleArray = sourceString.split(';').map(el => el.split(':'));
+
+  const newObject = styleArray
+    .filter(el => el.length > 1)
+    .map(y => y.map(word => word.trim()))
```
SanyaBratashchuk, line 21: you can trim elements in reduce

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    .reduce((a, x) => {
+      a[x[0]] = x[1];
+
+      return a;
+    }, {});
+
+  return newObject;
```
SanyaBratashchuk, line 28:
```suggestion
  return styleArray
    .filter(el => el.length > 1)
    .map(y => y.map(word => word.trim()))
    .reduce((a, x) => {
      a[x[0]] = x[1];

      return a;
    }, {});;
```

src/convertToObject.js
SanyaBratashchuk, line 21: y is'nt enough informative name of the variable

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #1420, opened 2022-09-18
Solution

11Olya112:


src/convertToObject.js
```diff
@@ -3,16 +3,24 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
+ *
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
```
Bohdan-Maksymiuk, line 13:
```suggestion
  const stylesObject = {};
```
name variables more clear 
do not use obj, arr, etc. names

Bohdan-Maksymiuk (approved):

PR #1424, opened 2022-09-19
Solution

Shepeha-Andriy:


Veremchukk (commented):
Well done

naziamoff (changes requested):
take a look at the checklist again

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+ * CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+  const rulesTrimAndSplit = sourceString
```
naziamoff, line 18: not proper name of the variable. Perfect variable name describes what is stored inside - that makes your code clear.
Trim and split in the name don't give me any additional info because I see that you split the string and trim it.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    .split(';')
+    .map(elem => elem.trim().split(':'));
+
+  rulesTrimAndSplit.forEach(rule => {
+    if (rule[0] && rule[1]) {
+      styles[rule[0].trim()] = rule[1].trim();
+    }
+  });
```
naziamoff, line 26: 5. [CODE PRACTICE] - Avoid using loops in this task, use iteration methods instead.
ForEach is actually a loop, so think of a different way of solving it. My recommendation is to practice reduce method in this task.

naziamoff (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1426, opened 2022-09-19
solution

ihorandrianov:


src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+    .split(';')
+    .map(a => a.split(':').map(b => b.trim()));
+
+  const styles = {};
+
+  for (const style of stylesRaw) {
+    if (style.length === 2) {
+      styles[style[0]] = style[1];
+    }
+  }
```
Dikey945, line 24: try to use iteration methods, you can use reduce method in this case

Dikey945: > _ No description provided.
Good Job! 🔥

OleksiiBlezniuk, line 24: Let's use array methods here, reduce is best option here 🙂

OleksiiBlezniuk (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+    .map(a => a.split(':').map(b => b.trim()));
+
+  const styles = stylesRaw.reduce((acc, [style, value]) => (style !== ''
+    ? {
+      ...acc,
+      [style]: value,
+    }
+    : {
+      ...acc,
+    }), {});
```
OleksiiBlezniuk, line 25: It's strange, better to use if in this case

OleksiiBlezniuk (approved):

PR #1428, opened 2022-09-20
Solution

AntonVaida:


src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
- *
- * @return {object}
- */
+//  * @param {string} sourceString
+//  *
+//  * @return {object}
+//  *
 function convertToObject(sourceString) {
-  // write your code here
+  const StyleList = {};
```
xiimiik, line 8: `const styleList = {};`

xiimiik (approved):
everything is fine, but we name objects with cameCase starting with a lowercase letter
GJ 🔥

silchyn (approved):
Cool!

PR #1429, opened 2022-09-20
Solutin

miamarichka:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+  sourceString
+    .split(';')
+    .filter(style => style.trim())
+    .forEach(keyVsValue => {
+      const [key, value] = keyVsValue.split(':');
+
+      obj[key.trim()] = value.trim();
+    });
+
+  return obj;
```
SanyaBratashchuk, line 29: create an object using reduce instead of forEach

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #1430, opened 2022-09-20
done, but in protest

ghost:


src/convertToObject.js
```diff
@@ -1,18 +1,21 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
fatihsamur, line 4: I think mentors will not like this name. checklist #2

fatihsamur (approved):
nice and short as always 👍 approved

src/convertToObject.js
```diff
@@ -1,18 +1,21 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesAsObject = {};
+
+  sourceString
+    .split(';')
+    .map(item => item.split(':'))
+    .map(props => props
+      .map(item => item
+        .trim()))
```
VinnieJ-2k20, line 11:
```suggestion
      .map(item => item.trim()))
```

VinnieJ-2k20 (approved):
Good job 👍

PR #1433, opened 2022-09-20
Task done

NazikVr:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
-  // write your code here
+  const result = {};
+
+  const temporality = sourceString.split(`;`).map((a) => a.split(':'));
+
+  for (const ch of temporality) {
+    if (ch[1] !== undefined) {
+      result[ch[0].trim()] = ch[1].trim();
+    }
+  }
```
SanyaBratashchuk, line 25: create a result object using reduce method instead of `for` loop

SanyaBratashchuk (changes requested):

src/convertToObject.js
NazikVr: how i can use  if (ch[1] !== undefined) in reduce?

liliia-livandovskaya: Please, take a look at this example and try to rewrite your code according to it 

```
    .map(element => element.split(':')).filter(element => element.length === 2)
    .reduce((prev, current) => {
      prev[current[0].trim()] = current[1].trim();

      return prev;
    }, {});
```

liliia-livandovskaya (approved):

PR #1435, opened 2022-09-20
Sting has been converted to object

iam-datsenko:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+    .map(el => el.trim())
+    .filter(el => el !== '');
+
+  for (let i = 0; i < modifiedString.length; i++) {
+    const singleStyle = modifiedString[i].split(':');
+    const key = singleStyle[0].trim();
+    const value = singleStyle[1].trim();
+
+    cssObject[key] = value;
+  }
```
iPanda177, line 31: loop is a bad idea, use array iteration method instead

iPanda177 (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const modifiedString = sourceString
+    .split(';')
+    .map(el => el.trim())
+    .filter(el => el !== '')
+    .map(el => el.split(':'))
+    .reduce((elA, elB) => ({
+      ...elA, [elB[0].trim()]: elB[1].trim(),
+    }), {});
```
OleksiiBlezniuk, line 24:
```suggestion
    .reduce((acc, cur) => ({
      ...acc,
      [elB[0].trim()]: elB[1].trim(),
    }), {});
```

OleksiiBlezniuk: 1 - it's not previous and current elements like in sort method callback, it's accumulator and current element
2 - start new object property at new line

OleksiiBlezniuk (approved):

PR #1436, opened 2022-09-20
solution

artem-musii:


Yevheniya-Deryka (approved):
Cool decision! 
I know there is no such variant in the tests but I'm just curious, can something go wrong if we we'll have smth like this: "font-weight:     ;" somewhere in the initial string? 🤔 🙂

artem-musii:
> Cool decision! I know there is no such variant in the tests but I'm just curious, can something go wrong if we we'll have smth like this: "font-weight: ;" somewhere in the initial string? 🤔 🙂

I haven't thought of that, seems like it will go wrong

Yevheniya-Deryka:
> > Cool decision! I know there is no such variant in the tests but I'm just curious, can something go wrong if we we'll have smth like this: "font-weight: ;" somewhere in the initial string? 🤔 🙂
> 
> I haven't thought of that, seems like it will go wrong

but it's just my 'imaginary condition' so never mind)))

Anna-Tkachenko (approved):
Good job! 🔥

PR #1437, opened 2022-09-20
Solution

Anna-Borodaienko:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  let stylesArray = sourceString.split('\n');
+
+  stylesArray = stylesArray.filter(style => style.length > 3);
```
jojnajbude, line 21: You can split by this separator ';' and use trim() for delete all spaces and '\n'

jojnajbude (approved):

src/convertToObject.js
OleksiiBlezniuk: Strange condition, why do you need it? Why 3?

Agree with @jojnajbude, better to separate by `;` and use trim

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+  stylesArray.forEach(element => {
+    const newElement = element
+      .substring(0, element.length - 1)
+      .split(':');
+
+    const property = newElement[0].trim();
+    const propertyValue = newElement[1].trim();
+
+    stylesObject[property] = propertyValue;
+  });
```
OleksiiBlezniuk, line 32: forEach is like `for` loop 🙂  Let's learn how to use reduce

OleksiiBlezniuk (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = sourceString.split(';')
+    .map(element => element.trim()
+      .split(':')
+      .map(item => item.trim()))
+    .filter(style => style.length > 1)
+    .reduce((prev, [ property, propertyValue ]) => ({
```
OleksiiBlezniuk, line 22:
```suggestion
    .reduce((acc, [ property, propertyValue ]) => ({
```

OleksiiBlezniuk: Array desctructuring, awesome 🔥

OleksiiBlezniuk: But in case if array, there is no need in spaces between brackets

OleksiiBlezniuk (approved):
Good job!

PR #1438, opened 2022-09-20
Solution

vladimir-dublya:


src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
 function convertToObject(sourceString) {
   // write your code here
+  const styles = sourceString
+    .split(';')
+    .map(element => element.trim())
+    .filter(element => element !== '');
+
+  const correctStyles = {};
+
+  for (let i = 0; i < styles.length; i++) {
```
imaksweb, line 25: [CODE PRACTICE] - Avoid using loops in this task, use iteration methods instead.

src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
+    const keys = styles[i]
+      .slice(0, styles[i]
+        .indexOf(':')).trim();
+    const values = styles[i]
+      .slice(styles[i]
+        .indexOf(':') + 1).trim();
```
imaksweb, line 31: [CODE STYLE] - All chained methods called on the same level should have same indentation. This makes code much more readable.
```suggestion
    const keys = styles[i]
      .slice(0, styles[i]
        .indexOf(':'))
      .trim();
    const values = styles[i]
      .slice(styles[i]
        .indexOf(':') + 1)
      .trim();
```

imaksweb (changes requested):

nikitaovch99 (approved):
You could do this much simpler with just map and without find and reduce but I guess you do you.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
   // write your code here
+  const styledObj = sourceString.split(';')
+    .map(prop => {
+      return prop.trim().split(':').map(elem => elem.trim());
+    })
+    .filter(prop => prop.length === 2)
+    .reduce((acc, prop) => ({
+      ...acc,
+      [prop[0]]: prop[1],
+    }), {});
```
Andr1yK, line 26: u can merge those methods, and add this style to the object only if prop.length === 2 :)

Andr1yK (changes requested):

naziamoff (approved):
GJ

PR #1439, opened 2022-09-20
solution

myroslav-diiak:


src/convertToObject.js
```diff
@@ -1,18 +1,29 @@
...
+  const stringArray = sourceString
+    .split(';')
+    .map(element => element.trim())
+    .filter(element => element !== '');
+
+  const newObj = {};
+
+  for (let i = 0; i < stringArray.length; i++) {
+    const arrayProperty = stringArray[i].split(':');
+
```
Vladosik1337, line 20: I think you can use here 'map' instead 'for'

Vladosik1337 (approved):
Nice)

src/convertToObject.js
```diff
@@ -1,18 +1,29 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const stringArray = sourceString
```
OleksiiBlezniuk, line 11: Better not to use data types in variable names, use more specific names

src/convertToObject.js
```diff
@@ -1,18 +1,29 @@
...
+
+  const newObj = {};
+
+  for (let i = 0; i < stringArray.length; i++) {
+    const arrayProperty = stringArray[i].split(':');
+
+    if (arrayProperty[0]) {
+      newObj[arrayProperty[0].trim()] = arrayProperty[1].trim();
+    }
+  }
```
OleksiiBlezniuk, line 24: Use array methods (reduce is best option here)

OleksiiBlezniuk (changes requested):

src/convertToObject.js
```diff
@@ -1,18 +1,29 @@
...
+
+  const newObj = {};
+
+  properties.forEach((element) => {
+    const property = element.split(':');
+
+    if (property[0]) {
+      newObj[property[0].trim()] = property[1].trim();
+    }
+  });
```
OleksiiBlezniuk, line 24: forEach is just like loop 🙂 Let's learn how to use `reduce`

OleksiiBlezniuk (changes requested):

src/convertToObject.js
```diff
@@ -1,18 +1,29 @@
...
+
+  const newObj = {};
+
+  properties.reduce((prev, currentValue) => {
+    const property = currentValue.split(':');
+
+    newObj[property[0].trim()] = property[1].trim();
+
+    return currentValue;
+  }, 0);
```
OleksiiBlezniuk, line 24: And now you've just used reduce just like jorEach 😢

OleksiiBlezniuk (changes requested):

src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
+    .map(element => element.trim())
+    .filter(element => element !== '');
+
+  return properties.reduce((prev, currentValue) => {
+    const property = currentValue.split(':');
+
+    prev[property[0].trim()] = property[1].trim();
+
+    return { ...prev };
+  }, {});
```
OleksiiBlezniuk, line 22: Ok, but mutation of incoming function parameter is a bad practice (now we know that this object is used only here, but it's not always so obvious like in this case)

OleksiiBlezniuk (approved):

PR #1440, opened 2022-09-20
convertToObject function was implemented

Yevheniya-Deryka:


artem-musii (approved):
GJ

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+  const propsList = sourceString.split(';');
+  const propsObject = {};
+
+  propsList.forEach(property => {
+    const currentProp = property.split(':');
+
+    if (currentProp[0] && currentProp[1]) {
+      propsObject[currentProp[0].trim()] = currentProp[1].trim();
+    }
+  });
```
OleksiiBlezniuk, line 26: forEach is like `for` loop 🙂 Let's learn how to use reduce

OleksiiBlezniuk (changes requested):

src/convertToObject.js
Yevheniya-Deryka: done! is it better to use reduce here just to learn?) or there is some other reason?

OleksiiBlezniuk: Reduce is more logical here if we using tools that supposed for specific operation

Of course you can do everything using loop (or forEach), and it will works fine, but why, if we have tools for it 🙂 

For example here we have array and we want to make object from it (to get one instance from list of instance) - it's exactly why reduce is needed. (also using reduce we don't need any empty variable outside our method)

Use cases of forEach - is when you need to do some operation with element and you don't need to get result from it (just like replacement of for loop, it will be more often used in DOM, React, promises, fetch, etc. 🙂 ) 

And of course this case is good for learning of reduce usage (and understanding that reduce can be used in many cases, not just to calculate sum of numbers)

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const propsList = sourceString.split(';');
+
+  return propsList.reduce((propsObject, property) => {
+    const currentProp = property.split(':');
+
+    if (currentProp[0] && currentProp[1]) {
+      propsObject[currentProp[0].trim()] = currentProp[1].trim();
```
OleksiiBlezniuk, line 23: Ok, but better to avoid incoming object mutation (now we know, that it is not used anywhere outside this method, but still it's a bad practice)

OleksiiBlezniuk (approved):

src/convertToObject.js
Yevheniya-Deryka: You mean this object - propsObject ?

PR #1441, opened 2022-09-20
js_style-to-object

xiimiik:


AntonVaida (approved):
GJ

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+  const styles = {};
+  const properties = sourceString
+    .split(';')
+    .map(elem => elem.trim().split(':'));
+
+  for (const prop of properties) {
+    if (prop[0] && prop[1]) {
+      styles[prop[0].trim()] = prop[1].trim();
+    }
+  }
```
OleksiiBlezniuk, line 24: Let's use array methods for this purpose, (reduce is good option here)

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
10/10 👍

PR #1443, opened 2022-09-20
Solution

Rina07:


Oleksandra-Shevchenko (approved):
LGTM 💯

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+  const entries = sourseArr.map((el) => el.split(':'));
+
+  entries.forEach((entry) => {
+    const key = entry[0];
+    const value = entry[1];
+
+    if (value) {
+      obj[key.trim()] = value.trim();
+    }
+  });
```
OleksiiBlezniuk, line 30: Well, forEach is like `for ... of` loop 🙂  Let's learn how to use reduce

OleksiiBlezniuk (changes requested):

src/convertToObject.js
Rina07: have learnt)

OleksiiBlezniuk (approved):
Good job!

PR #1444, opened 2022-09-21
Solution

Oleksandra-Shevchenko:


Rina07 (approved):
LGTM :fire:

silchyn (changes requested):
You did not follow checklist rule #2, but you said the opposite. Why?

naziamoff (changes requested):
Checklist p.2, p.5

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+  const styles = sourceString
+    .split(';')
+    .map(style => style.split(':')
+      .map(styleEl => styleEl.trim()));
+
+  styles.forEach(style => {
+    if (style.length !== 0 && style[1] !== undefined) {
+      styleSet[style[0]] = style[1];
+    }
+  });
```
naziamoff, line 30: forEach is actually a loop. You should avoid using them in this task

naziamoff (changes requested):

naziamoff (approved):
🔥

PR #1446, opened 2022-09-21
convertToObject solution

vmamchur:


src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const correctStyles = {};
+  const stylesFormatting = sourceString
+    .split(';')
+    .map(element => element.split(':')
+      .map(char => char.trim()))
```
imaksweb, line 21: [CODE STYLE] - All chained methods called on the same level should have same indentation. This makes code much more readable.

Also I'm not sure that `char` is a descriptive name for this variable
```suggestion
    .map(element => element
      .split(':')
      .map(char => char.trim()))
```

imaksweb (approved):
GJ 🔥  Just left some suggestions 🙂

Anna-Tkachenko (approved):
Well Done! 🔥

PR #1454, opened 2022-09-21
Solution

Yeliseiev:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesSet = {};
+
+  const styles = sourceString.split(';');
+
+  styles.forEach(style => {
+    if (style.includes(':')) {
```
Maksym845, line 22: I don`t know if can we use forEach loop in this task.

Maksym845 (approved):
But at all a good solution.

src/convertToObject.js
Yeliseiev: It does not modifying the given string.

silchyn (approved):
Cool!

PR #1460, opened 2022-09-21
convertToObject

Arsen-Shpak:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const obj = {};
+  const arr = sourceString.split(';').map((item) =>
```
dmytristepanov, line 19: image

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+    item
+      .trim()
+      .split(':')
+      .filter((element) => element !== '')
+  );
+  // console.log(arr);
+
+  for (const name of arr) {
+    if (name[0] && name[1]) {
+      obj[name[0].trim()] = name[1].trim();
```
dmytristepanov, line 29: image

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+    }
+  }
+
+  return obj;
```
dmytristepanov, line 33: Naming

dmytristepanov (approved):
Good job 🔥 , look at the 2nd and 5th rules in the checklist

silchyn (changes requested):
Checklist rule no. 5

naziamoff (changes requested):
checklist p.2
variable name should describe **what is stored inside**, not its data type

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const rightElementsOfString = sourceString.split(';').map((property) =>
```
naziamoff, line 18: wordy naming gives us much more context than `object` or sth, so good job with it 🚀 
But if you can make it shorter - you should avoid unnecessary wordiness
```suggestion
  const validProperties = sourceString.split(';').map((property) =>
```

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+    property
+      .trim()
+      .split(':')
+      .filter((element) => element !== '')
+  );
+
+  const ourCssInObject = rightElementsOfString.reduce((currentObj, item) => {
```
naziamoff, line 25: same here
```suggestion
  const convertedCss = rightElementsOfString.reduce((currentObj, item) => {
```

naziamoff (approved):
Good job! But take a look at my comments, naming matters

PR #1463, opened 2022-09-21
solution was added

Maksym845:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayOfStyles = sourceString.replace(/\n/g, '')
```
Yeliseiev, line 17: You don't need replace, because you use trim() for every element in array. It replace '\n' anyway.

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+    .split(';')
+    .map(element => element.trim())
+    .filter(element => element !== '')
+    .map(line => {
+      const temp = line.split(':');
+
+      return [temp[0].trim(), temp[1].trim()];
+    });
+  const result = {};
```
Yeliseiev, line 26: 2. [CODE STYLE] - Avoid variable namings like `array , object arr, obj,` `result`. It will decrease code readability

Yeliseiev (changes requested):
Great job. It's cool that you remember the replace method. I forgot about it.
Rename result please)

src/convertToObject.js
Maksym845: Good point, thank you!

Yeliseiev (approved):
Nice!

silchyn (changes requested):
Avoid using data types in variable names

naziamoff (approved):

PR #1465, opened 2022-09-21
v1: solution

igor-1987:


AndriyZhabynets (approved):
Good job!

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+    .filter(item => item.length > 1);
+
+  const stylesObj = stylesArr.reduce((accumulator, property) => {
+    return {
+      ...accumulator,
+      [property[0]]: property[1],
+    };
+  }, {});
+
+  return stylesObj;
```
OleksiiBlezniuk, line 33:
```suggestion
  return stylesArr.reduce((accumulator, property) => ({
    ...accumulator,
    [property[0]]: property[1],
  }), {});
```

OleksiiBlezniuk: Don't forget that you can return value, without saving to variable 🙂

OleksiiBlezniuk (approved):
👍

PR #1466, opened 2022-09-21
implemented convertToObject function

SerhiiRomaniukWD:


src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedStyles = {};
+
+  sourceString.split(';')
```
glugovsky, line 19: Checklist point 4: all chained methods called on the same level should have same indentation
```suggestion
  sourceString
    .split(';')
```

glu3s (approved):
Nice work!

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedStyles = {};
+
+  sourceString
+    .split(';')
+    .filter(elem => elem.trim().length > 1)
+    .map(property => {
+      convertedStyles[property.split(':')[0].trim()]
```
naziamoff, line 23: Try to create a variable for `property.split(':')[0].trim()` to make your code more understandable. It is not really easy to understand what is happening there

naziamoff (changes requested):

naziamoff (approved):

PR #1468, opened 2022-09-21
Solution v01

anton-iskryk:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+
```
artem-nasadchuk, line 18: just a tip: don't use object, array, etc. in naming a variable. for example, I think just 'styles' is better.

artem-nasadchuk (approved):

src/convertToObject.js
Anna-Tkachenko: fix @artem-nasadchuk comment

Anna-Tkachenko (approved):

PR #1473, opened 2022-09-22
Solution

OleksandrGlulhovskiy:


ArturNagaycev (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+
+function callback(acc, prop) {
+  const styleName = prop[0].trim();
+  const styleValue = prop[1].trim();
+
+  return {
+    ...acc,
+    [styleName]: styleValue,
+  };
 }
```
OleksiiBlezniuk, line 33:
```suggestion
function callback(acc, [key, value]) {
  const styleName = key[0].trim();
  const styleValue = value[1].trim();

  return {
    ...acc,
    [styleName]: styleValue,
  };
}
```

OleksiiBlezniuk: You can use array destructuring here

OleksiiBlezniuk (approved):

src/convertToObject.js
ArturNagaycev: then we don't need to use a value in brackets(after key and value), right?

  const styleName = key.trim();
  const styleValue = value.trim();

PR #1475, opened 2022-09-22
Solution style to object

Ihor-1993:


NickBrodyuk:
nice

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  styles.forEach(style => {
+    if (style.includes(':')) {
+      const stuff = style.split(':');
+      const property = stuff[0].trim();
+      const value = stuff[1].trim();
+
+      stylesSet[property] = value;
+    }
+  });
```
Andr1yK, line 29: looks like as reduce

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  const styles = sourceString.split(';');
+
+  styles.forEach(style => {
+    if (style.includes(':')) {
+      const stuff = style.split(':');
+      const property = stuff[0].trim();
+      const value = stuff[1].trim();
+
+      stylesSet[property] = value;
```
Andr1yK, line 27:
```suggestion
      const stuff = style.trim().split(':');

      stylesSet[property] = value;
```

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesSet = {};
+
+  const styles = sourceString.split(';');
+
+  styles.forEach(style => {
+    if (style.includes(':')) {
+      const stuff = style.split(':');
```
Andr1yK, line 23: try to decrease complexity (don't use nested loops)

Andr1yK (changes requested):

PR #1477, opened 2022-09-22
Solution

baranovskazoya:


src/convertToObject.js
```diff
@@ -3,16 +3,24 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';')
+    .map(element => element.split(':')).filter(element => element.length === 2)
+    .reduce((previous, current) => {
+      previous[current[0].trim()] = current[1].trim();
```
imaksweb, line 20: All chained methods called on the same level should have same indentation.
```suggestion
  return sourceString
    .split(';')
    .map(element => element.split(':'))
    .filter(element => element.length === 2)
    .reduce((previous, current) => {
      previous[current[0].trim()] = current[1].trim();
```

imaksweb (approved):
Good job! Just one remark 🙂

src/convertToObject.js
```diff
@@ -3,16 +3,24 @@
...
+ *  properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';')
+    .map(element => element.split(':')).filter(element => element.length === 2)
```
OleksiiBlezniuk, line 18:
```suggestion
    .map(element => element.split(':'))
    .filter(element => element.length === 2)
```

PR #1480, opened 2022-09-23
Solution

andrew-satyr:


varkovroma (approved):

naziamoff (changes requested):
- checklist p.2
- let's practice reduce method in this task

src/convertToObject.js
```diff
@@ -1,18 +1,18 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  let classes = sourceString.split(';');
+  const resultObj = {};
```
Anna-Tkachenko, line 5: it's not a good variable name

Anna-Tkachenko (approved):

PR #1481, opened 2022-09-23
cod for style object

RusNeday:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  const splitArr = sourceString.split(';');
+
+  splitArr.sort((a, b) => a.localeCompare(b));
```
SanyaBratashchuk, line 21: why do you need sorting?

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  for (const key of splitArr) {
+    const i = key.split(':');
+
+    if (i[1] !== undefined) {
+      result[i[0].trim()] = (i[1].trim());
+    }
+  }
```
SanyaBratashchuk, line 29: Let's use array methods

SanyaBratashchuk (changes requested):

marina-tilniak (approved):

PR #1483, opened 2022-09-23
Solution

rstarovoit:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    .split(';')
+    .map(el => el.trim()
+      .split(':')
+      .map(i => i.trim()))
+    .filter(style => style.length > 1)
+    .reduce((data, [property, propertyValue]) => ({
+      ...data,
+      [property]: propertyValue,
+    }),
+    {});
```
grafit-off, line 27:
```suggestion
    .reduce((data, [property, propertyValue]) => ({
      ...data,
      [property]: propertyValue,
    }), {});
```

grafit-off (approved):
GJ

naziamoff (changes requested):
That's really great that you've practiced array methods like that and managed to do it with just one variable 🔥  
Now we should care of your code's readability.
- fix Mykola's comment, he is right
- think of a way to separate some logic, because of the methods chain is really long it is not really easy to understand what you are doing

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const classes = sourceString
```
naziamoff, line 17:
```suggestion
  const styles = sourceString
```

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    .split(';')
+    .map(el => el.trim()
+      .split(':')
+      .map(i => i.trim()))
+    .filter(style => style.length > 1);
+
+  const result = classes
```
naziamoff, line 24: `result` is not the best name for the variable
I would make it like 
```suggestion
  const convertedStyles = classes
```

naziamoff (approved):
Good job! But pay attention to the naming

naziamoff (approved):

PR #1491, opened 2022-09-25
Solution

oleksandr-korobko:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+  const sortedMess = sourceString
+    .split(';')
+    .map(a => a.split(':')
+      .map(b => b.trim()));
+
+  for (const element of sortedMess) {
+    if (element[0] && element[1]) {
+      resultStyles[element[0]] = element[1];
+    }
+  }
```
myroslav-diiak, line 27: Use array method reduce for this, don`t use for or foreach

myroslav-diiak (approved):

Denchepornyuk (approved):

PR #1492, opened 2022-09-25
Object to style Solution

VovaDzhuryk:


MAKSOti (approved):
GJ)
everything is fine but think about the name of the variable
[CODE KNOWLEDGE] - Don't need to make copy before each operation. Moreover, methods like filter or map don't mutate original array.

PR #1493, opened 2022-09-25
solution

poznianski:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+  return sourceString.split(';')
+    .map(property => property.trim().split(':'))
+    .map(property => property.map(part => part.trim()))
+    .filter(property => property.length > 1)
+    .reduce(callback, {});
+}
+
+function callback(accumulator, property) {
+  const styleName = property[0];
+  const styleValue = property[1];
```
Dikey945, line 26: More like advise, so u can use destructuring assignment here to avoid creating variables, 

`function callback(accumulator, [styleName, styleValue])`

Dikey945:
Good Job! 🔥

src/convertToObject.js
Denchepornyuk: you can, but for this code `callback(accumulator, [styleName, styleValue])` more complicated than this one `callback(accumulator, property)`
Destructuring you can use in this way
```
function callback(accumulator, property) {
  const [styleName, styleValue] = property;
```

Denchepornyuk (approved):

PR #1505, opened 2022-09-27
Solution

vaadyak:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const arrayStyles = sourceString.split(';')
+    .map(property => property.split(':'));
+
+  const deleteGap = arrayStyles
+    .map(items => items.map(element => element.trim()))
```
Andr1yK, line 22:
```suggestion
    .map(property => property.trim().split(':'));

  const deleteGap = arrayStyles
```

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    .filter(item => item.length > 1);
+
+  return deleteGap.reduce((item, style) => {
+    return {
+      ...item,
+      [style[0]]: style[1],
+    };
+  }, {});
```
Andr1yK, line 30:
```suggestion
  return deleteGap.reduce((item, [key, value]) => ({
      ...item,
      [key]: value,
    }), {});
```

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const arrayStyles = sourceString.split(';')
+    .map(property => property.split(':'));
+
+  const deleteGap = arrayStyles
```
Andr1yK, line 21: U can use reduce here

Andr1yK (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const arrayStyles = sourceString.split(';')
```
naziamoff, line 18: you should not mention data type in the name of the variable, it doesn't give any additional info

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    .map(property => property.split(':'));
+
+  const deleteGap = arrayStyles
```
naziamoff, line 21: `deleteGap` name should be changed. The name should describe **what is stored inside**, not the action you have done to it

naziamoff (approved):
GJ but I recommend you to fix naming here

PR #1507, opened 2022-09-27
solution

hajnaloltyan:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let styleArray = sourceString.trim().split(';');
+
+  styleArray = styleArray.map(item => item.split(':'));
+
+  styleArray = styleArray.map(item => item.map(element => element.trim()));
+
+  styleArray = styleArray.filter(item => item.length !== 1);
```
kamal-tufekcic, line 23: You can use const here, and use all the methods at once

kamal-tufekcic (approved):

src/convertToObject.js
VinnieJ-2k20: Yep, better to rewrite using chained array methods, since they all return a value.

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+
+  const styleObject = Object.fromEntries(styleArray);
```
VinnieJ-2k20, line 25: Nice 👍

VinnieJ-2k20 (changes requested):
Need to fix some code style.

ihor-jpeg (approved):

PR #1510, opened 2022-09-28
Solution

divineretribution1:


src/convertToObject.js
```diff
@@ -1,18 +1,28 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {}
+  const arr = sourceString.replace(/\s\s+/g, '')
+  .replace('\n', '')
+  .split(';')
+  .slice(0, -1)
+  .filter(item => !!item).map(string => {
```
SanyaBratashchuk, line 22:
```suggestion
  .filter(Boolean).map(string => { 
```

src/convertToObject.js
```diff
@@ -1,18 +1,28 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {}
+  const arr = sourceString.replace(/\s\s+/g, '')
+  .replace('\n', '')
+  .split(';')
+  .slice(0, -1)
+  .filter(item => !!item).map(string => { 
+    return obj[`${string.split(':')[0]}`] = string.split(':')[1].trim()
```
SanyaBratashchuk, line 23: use forEach or reduce to create an object

SanyaBratashchuk (changes requested):

src/convertToObject.js
```diff
@@ -1,18 +1,27 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {}
+  const arr = sourceString
+  .split(';')
+  .map(item => item.split(':').map(string => string.trim()))
+  .filter(item => item.length > 1);
+  arr.forEach(item => (obj[item[0]] = item[1]))
```
liliia-livandovskaya, line 22:
```suggestion
  const arr = sourceString
    .split(';')
    .map(item => item.split(':').map(string => string.trim()))
    .filter(item => item.length > 1);
    
  arr.forEach(item => (obj[item[0]] = item[1]))
```

liliia-livandovskaya (approved):

PR #1517, opened 2022-10-02
Solution

vallerii:


MAKSOti (approved):
GJ 
everything is fine but think about the name of the variable

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    .filter(element => element !== '')
+    .map(line => {
+      const data = line.split(':');
+
+      return [data[0].trim(), data[1].trim()];
+    });
+
+  stylesList.forEach(item => {
+    results[item[0]] = item[1];
+  });
```
OleksiiBlezniuk, line 29: ok, but better to use `reduce` for this purpose

OleksiiBlezniuk (approved):

PR #1523, opened 2022-10-05
function convertToObject was added

Olena-Shevchenko:


liliia-livandovskaya (changes requested):
GJ! Lets work on changes

PR #1528, opened 2022-10-13
Solution

anastasia-babko:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+  let result = sourceString.split(';');
+
+  result = result.map(str => str.trim()
+    .split(':')
+    .map(unit => unit.trim()));
+
+  result = result.filter(str => str.length > 1);
```
SanyaBratashchuk, line 24:
```suggestion
  const result = sourceString.split(';')
    .map(str => str.trim()
    .split(':')
    .map(unit => unit.trim()))
    .filter(str => str.length > 1);
```

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+
+  for (const item of result) {
+    stylesObject[item[0]] = item[1];
+  }
+
+  return stylesObject;
 }
```
SanyaBratashchuk, line 31: use reduce to create a result object

SanyaBratashchuk (changes requested):

src/convertToObject.js
anastasia-babko: done.

SanyaBratashchuk (approved):

PR #1530, opened 2022-10-14
convert

metinbicaksiz:


src/convertToObject.js
```diff
@@ -3,16 +3,44 @@
...
+  const obj = {};
+
+  for (let i = 0; i < noEmpty.length; i++) {
+    const splitItem = noEmpty[i].split(':');
+
+    const trimmedSplit = splitItem.map(element => {
+      return element.trim();
+    });
+
+    obj[trimmedSplit[0]] = trimmedSplit[1];
```
amitpatiljc, line 40: there is method for this type of operation where we can present `key` and `value` pair

amitpatiljc (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,43 @@
...
   // write your code here
+  const properties = sourceString
+    .split(';')
+    .map(element => {
+      if (typeof element === 'string') {
+        return element.trim();
+      }
+    })
+    .filter(element => {
+      return element !== '';
```
amitpatiljc, line 27: do the explicit return

amitpatiljc, line 27:
```suggestion
    .filter(element => element !== '');
```

amitpatiljc (changes requested):

amitpatiljc (approved):
🔥

PR #1535, opened 2022-10-17
Solution

alexxxstep:


src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const strArr = sourceString.split(';');
+
+  const obj = strArr.reduce((acc, i) => {
```
polosanya, line 19: Please use more informative names for your variables

polosanya (changes requested):
Nice work, but please give more informative names for your variables. It will make your code much cleaner

polosanya (approved):
Good job!

PR #1537, opened 2022-10-18
Solution

bokorzita:


src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+  const stylesArray = sourceString.split(';');
+
+  const prettyStylesArray
+    = stylesArray.map(
+      s => s.replace(/\s\s+/g, ' ')
+        .trim())
+      .filter(s => s);
+
+  prettyStylesArray.map(
+    s => s.replace(/\s+(:)/g, ':'));
```
amitpatiljc, line 28: instead of all this you can just `trim()` and `split(':')` 
and filter out blank spaces with `!s.includes('')`

amitpatiljc (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
+  const stylesArray = sourceString.split(';');
+
+  function fillObj(obj, e) {
+    obj[e[0]] = e[1];
+  }
+
+  stylesArray.forEach(
+    x => x.trim().length > 2
+      ? keyValueArray.push(x.split(':'))
+      : null);
```
jstmpelowycz, line 28: 1) no need to check if line length is greater than 2, checking for non-emptiness is enough
2) let's use `filter` method, you literally do the same job but in a twisted way

src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
+
+  for (let i = 0; i < keyValueArray.length; i++) {
+    keyValueArray[i][0] = keyValueArray[i][0].trim();
+    keyValueArray[i][1] = keyValueArray[i][1].trim();
+    fillObj(stylesObj, keyValueArray[i]);
+  }
```
jstmpelowycz, line 34: too complicated, there is a simpler way

jstmpelowycz (changes requested):
let's optimize your solution, so what the idea is
```
sourceString
  .split(';') // split by lines
  .filter(
    // remove empty lines (required trimming)
  ) 
  .forEach(
    // split to key and value
    // write trimmed key and value to object
  )
...
```

amitpatiljc (approved):
Good Job!

PR #1545, opened 2022-10-22
solution

Hanna-Zaitseva:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+  const sourceArray = sourceString.split(';');
+  const stylesObject = {};
+
+  for (let i = 0; i < sourceArray.length; i++) {
+    if (sourceArray[i].trim()) {
+      const style = sourceArray[i].split(':');
+
+      stylesObject[style[0].trim()] = style[1].trim();
+    };
+  }
```
tetiana-ivanchenko, line 26: Checklisrt:

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #1548, opened 2022-10-25
solution

MaxFedak:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * and values are the values otest.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const strings
+  = sourceString.replace(/\n/g, '')
```
denys-danyliuk, line 17:
```suggestion
 return sourceString
   .replace(/\n/g, '')
```

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    .split(';')
+    .filter(item => item.includes(':'))
+    .map(str => str.split(':')
+      .map(array => array.trim()))
+    .reduce((object, cur) => {
+      return {
+        ...object,
+        [cur[0]]: cur[1],
+      };
+    }, {});
```
denys-danyliuk, line 27: create callback and move it from function

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const strings
+  = sourceString.replace(/\n/g, '')
+    .split(';')
+    .filter(item => item.includes(':'))
+    .map(str => str.split(':')
+      .map(array => array.trim()))
```
denys-danyliuk, line 21:
```suggestion
    .map(str => str.trim().split(':')
```

denys-danyliuk (changes requested):

src/convertToObject.js
MaxFedak: it doesn't work
trimming this way doesnt help us getting rid of whitespaces before and after ':'. That's why i decided to do it after split

MaxFedak: Знімок екрана 2022-10-26 о 16 56 02

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
   // write your code here
+  return sourceString.replace(/\n/g, '')
+    .split(';')
+    .filter(item => item.includes(':'))
+    .map(str => str.split(':')
+      .map(array => array.trim()))
+    .reduce(createObj, {});
 }
 
+const createObj = (object, cur) => {
```
denys-danyliuk, line 24:
```suggestion
const createObj = (prev, cur) => {
```

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+  return {
+    ...object,
+    [cur[0]]: cur[1],
+  };
+};
```
denys-danyliuk, line 29:
```suggestion
const createObj = (object, cur) => {
    ...object, [cur[0]]: cur[1],
};
```

denys-danyliuk: actually you can trim your elements here instead of creating nested map

denys-danyliuk (approved):

PR #1555, opened 2022-10-29
solution

andrudenko:


src/convertToObject.js
```diff
@@ -1,18 +1,19 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const object = {};
+  const sourceArray = sourceString.split(';').filter(el => el.length > 5);
```
ivanchuk-yurii, line 5: why do you filter by length?

ivanchuk-yurii (changes requested):

src/convertToObject.js
andrudenko: in order to get rid from empty elements which could be created by separating with ';'. In file stylesString.js on lines 21 and 23.

ivanchuk-yurii: so first trim string and then check if it's not empty. 5 is "magic" value and wouldn't work if spaces will be more

ivanchuk-yurii (changes requested):

src/convertToObject.js
andrudenko: thx

src/convertToObject.js
```diff
@@ -1,18 +1,21 @@
...
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const object = {};
+  const sourceArray = sourceString.split(';')
+    .map(str => str.trim())
+    .filter(el => el.length > 0);
```
ivanchuk-yurii, line 7: can be just `el.length` in filter

ivanchuk-yurii (approved):

PR #1556, opened 2022-10-29
Convert to Object was created

iryna-lelyk:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+  const objWithStyles = arrWithStyles.reduce((prev, str) => {
+    const arrWithValues = str.split(':');
+    const key = arrWithValues[0].trim();
+    const value = arrWithValues[1].trim();
+
+    return {
+      ...prev,
+      [key]: value,
+    };
+  }, {});
```
taraselnik, line 31: cool way) I'll ask you in Slack how you got it)

taraselnik (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
-  // write your code here
+  const arrWithStyles = sourceString
+    .split(';')
+    .map(str => str.trim())
+    .filter(str => str.length > 1);
+
+  const objWithStyles = arrWithStyles.reduce((prev, str) => {
+    const arrWithValues = str.split(':');
+    const key = arrWithValues[0].trim();
+    const value = arrWithValues[1].trim();
```
denys-danyliuk, line 25: don't create this variables inside reduce method, otherwise on each call of this method they will create over and over again

src/convertToObject.js
denys-danyliuk, line 31: move this callback outside of this function, to avoid creating it on each call

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const arrWithStyles = sourceString
+    .split(';')
+    .map(str => str.trim())
+    .filter(str => str.length > 0);
+
+  const objWithStyles = arrWithStyles.map(str => str.split(':')
+    .map(word => word.trim()))
+    .reduce(createObj, {});
```
denys-danyliuk, line 24:
```suggestion
  return sourceString
    .split(';')
    .map(str => str.trim())
    .filter(str => str.length > 0);
    .map(str => str.split(':')
    .map(word => word.trim()))
    .reduce(createObj, {});
```

denys-danyliuk (approved):

PR #1557, opened 2022-10-29
Solution

taraselnik:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
-  // write your code here
+  const styleStorageObj = {};
+  const stylePropWithValues = sourceString.split('\n');
+  const stylePropWithValuesTrimmed = stylePropWithValues
+    .map(string => string
+      .slice(0, -1)
+      .trim()
+      .split(':')
+      .map(
+        entries => entries.trim()))
```
iryna-lelyk, line 26: If you're trimming each word in line 26, I guess you don't have to trim strings in line 24. But I don't know if it would work properly.

iryna-lelyk: And also you can put 25 and 26 lines in one line

iryna-lelyk (approved):
Interesting solution!

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleStorageObj = {};
```
mykola-koval, line 18: why do you need this variable?

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+  const stylePropWithValues = sourceString.split('\n');
```
mykola-koval, line 19: you can split by `;` and then by `:`

mykola-koval: i guess this will simplify your solution

mykola-koval (changes requested):

illiarerun (approved):

PR #1564, opened 2022-10-31
solution

IOneStarI:


Stepico (approved):
Seems good, so clear, I'll approve it, but I have some concerns that we weren't allowed to use forEach in this task, yet I'm not sure

mykola-koval (approved):

PR #1567, opened 2022-10-31
SOlition

bohdan-stepanets:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesWithSpace = sourceString
+    .split(';')
+    .map(style => style.trim())
+    .filter(style => style.length > 0)
+    .map(style => style.split(':'));
```
masha-girya, line 21: here you can create more suitable names for arguments. For example, not style but row. It will be easier to understand what you do in methods

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+
+  const stylesNoSpace = stylesWithSpace.map(style =>
+    style.map(key => key.trim()));
```
masha-girya, line 24: i think you could use this map method in previous chain: `.map(style => style.map(key => key.trim()))`

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+
+  const styles = stylesNoSpace.reduce((prev, style) => {
+    return {
+      ...prev,
+      [style[0]]: style[1],
+    };
+  },
+  {});
+
+  return styles;
```
masha-girya, line 34: and also add this to the chain

masha-girya (changes requested):
I checked you code (if it will be in one chain) on my comp, it works!))

src/convertToObject.js
```diff
@@ -13,23 +13,23 @@
...
-  {});
+function convertToObject(sourceString) {
+  const styles = sourceString
+    .split(';')
+    .map(row => row.trim())
+    .filter(row => row.length > 0)
+    .map(item => item.split(':'))
+    .map(item =>
+      item.map(key => key.trim()))
+    .reduce((prev, element) => callbackForReduce(prev, element), {});
```
masha-girya, line 32: reduce takes a callback as a parameter and start value. but you wrote that reduce takes (prev, elem)
`.reduce(callbackForReduce(prev, element), {});`

masha-girya (changes requested):
sorry, you have to change just one moment:)

masha-girya (approved):

illiarerun (approved):

PR #1568, opened 2022-10-31
style to object

Svitlana-Yudina:


src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrWithoutSemi = sourceString
+    .split('')
+    .filter(el => el !== ';');
```
mitattt, line 19: you could use just` split(";")` incstead

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+
+  const arrOfProp = arrWithoutSemi
+    .join('')
+    .split('\n')
```
mitattt, line 23: what does it mean `.split('\n')` here?

mitattt (approved):
gj, there are some mistakes here 👇

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrWithoutSemi = sourceString
```
mykola-koval, line 17: bad variable name

src/convertToObject.js
mykola-koval: agree

src/convertToObject.js
mykola-koval: you can firstly split by `;` and then by `:`

mykola-koval (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+
+function getProperty(prev, el) {
+  const index = el.indexOf(':');
+
+  return {
+    ...prev,
+    [el.slice(0, index).trim()]: el.slice(index + 1).trim(),
+
+  };
 }
```
mykola-koval, line 39: it's not clear to me why do you need this function?

mykola-koval (approved):

PR #1573, opened 2022-10-31
solution

andrii-bassarab:


src/convertToObject.js
```diff
@@ -1,18 +1,29 @@
 'use strict';
 
+const stylesString = require('./stylesString');
```
denys-danyliuk, line 3: why do you need this?

src/convertToObject.js
```diff
@@ -1,18 +1,29 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArr = stylesString.split(';');
```
denys-danyliuk, line 12:
```suggestion
  const stylesArr = sourceString.split(';');
```

src/convertToObject.js
```diff
@@ -1,18 +1,29 @@
...
+  const stylesItem = [];
+  const stylesObj = {};
+
+  stylesArr.forEach(string => {
+    stylesItem.push(string.trim().split(':'));
+  });
```
denys-danyliuk, line 18: don't create new array, let's use iteration methods with 'stylesArr' to do this

src/convertToObject.js
```diff
@@ -1,18 +1,29 @@
...
+
+  stylesItem.forEach(element => {
+    if (element[1] !== undefined) {
+      stylesObj[element[0].trim()] = element[1].trim();
+    }
+  });
```
denys-danyliuk, line 24: same here, let's use iteration methods

denys-danyliuk (changes requested):
GJ! Pay attention to the comments

mykola-koval (approved):

denys-danyliuk (approved):

PR #1575, opened 2022-10-31
complete

spiritkyiv:


Largoxoxo (approved):
Great solution!

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+ * and values are the values of related CSS
+ * properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const StylesObject = {};
```
denys-danyliuk, line 17:
```suggestion
```

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+
+  sourceString.split(';')
```
denys-danyliuk, line 19:
```suggestion
  return sourceString
    .split(';')
```

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    .map(style =>
+      style.trim()).filter(style => style)
+    .forEach(style => {
+      const styleValue = style.split(':');
+
+      StylesObject[styleValue[0].trim()] = styleValue[1].trim();
+    });
```
denys-danyliuk, line 26: don't use forEach to mutate anything, forEach is needed to iterate through array elements and use callback for every of them. Use reduce to build an object
Also move callback outside of function to avoid creating it everytime

denys-danyliuk (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+ * and values are the values of related CSS
+ * properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const StylesObject = sourceString.split(';')
```
illiarerun, line 17: it's not a class, it's a variable
```suggestion
  const stylesObject = sourceString.split(';')
```

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    .map(style =>
+      style.trim()).filter(style => style)
```
illiarerun, line 19: all methods in chain should have the same indentation
```suggestion
    .map(style => style.trim())
    .filter(style => style)
```

illiarerun (changes requested):

Olexandr6 (approved):

PR #1578, opened 2022-10-31
added solution

kryvets-vadym:


src/convertToObject.js
```diff
@@ -1,18 +1,31 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resultObject = {};
+
+  const sourceStringArray = sourceString
+    .replace(/\s+/g, ' ')
```
Oleksii-Siviolov, line 12: This is nice. I forgit about this expression, but it is really a good case to use it.

src/convertToObject.js
```diff
@@ -1,18 +1,31 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resultObject = {};
```
Oleksii-Siviolov, line 9: Not the best way to name your object, see #2 in Checklist. Nobody will understand, what is inside of this object.

Oleksii-Siviolov (changes requested):
Nicely done, I like your solution, but you should pay attention to the naming.

Oleksii-Siviolov (approved):
GJ! Keep it up!

mykola-koval (approved):
the logic could be simplified a little, but ok
check other students' work, make some student reviews :)

PR #1581, opened 2022-11-01
Solution

Vasyliy22:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+  return arr.reduce((acc, el) => {
+    const splitArr = el.split(':');
+
+    if (splitArr.length !== 2) {
+      return acc;
+    }
+
+    return {
+      ...acc,
+      [splitArr[0].trim()]: splitArr[1].trim(),
```
Anastasiia-Svintsova, line 28: It's better to move `[splitArr[0].trim()]` and `splitArr[1].trim()` to variables

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.split(';');
+
+  return arr.reduce((acc, el) => {
```
Anastasiia-Svintsova, line 19: What does acc mean? Give meaningful names to variables

Anastasiia-Svintsova (changes requested):

polosanya (approved):
Good job!

PR #1583, opened 2022-11-01
done

AndreyChernysh:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleArray = sourceString.split(';').map((item) => {
+    return item.trim();
```
Andriy-Shymkiv, line 18: That is what @illiarerun comented to my code, so you should do it the same way:

don't use map to mutate anything, map is used to transform elements and create new array using them. use reduce to create new object.

Olexandr6: +1

Olexandr6 (changes requested):
Check comment above

Olexandr6 (approved):

PR #1584, opened 2022-11-01
Task was Solved

Stepico:


IOneStarI (approved):
GJ, on my opinion little bit complicated)

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const originString = sourceString.replace(/\s\s+/g, ' ');
+  const styles = originString.split(';');
+  const stylesToObject = {};
+  const validStyles = styles
+    .filter(style => style
+      .split('')
+      .find(symbol => symbol === ':'));
```
illiarerun, line 23: would be much easier
```suggestion
  const validStyles = styles
    .filter(style => style.includes(':'));
```

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  validStyles.map((style) => {
+    const copyStyle = style.split(':');
+
+    stylesToObject[copyStyle[0].trim()] = copyStyle[1].trim();
+  });
```
illiarerun, line 29: don't use map to mutate anything, map is needed to transform array elements and create new one. use reduce to build an object

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const originString = sourceString.replace(/\s\s+/g, ' ');
+  const styles = originString.split(';');
+  const stylesToObject = {};
```
illiarerun, line 19:
```suggestion
```

illiarerun (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+  const styles = originString.split(';');
+  const validStyles = styles.filter(style => style.includes(':'));
+
+  const stylesToObject = validStyles.reduce((prev, style) => {
+    const copyStyle = style.split(':');
+
+    prev[copyStyle[0].trim()] = copyStyle[1].trim();
+
+    return prev;
+  }, {});
```
denys-danyliuk, line 27: move callback outside of function to avoid creating each time when func runs

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const originString = sourceString.replace(/\s\s+/g, ' ');
+  const styles = originString.split(';');
+  const validStyles = styles.filter(style => style.includes(':'));
+
+  const stylesToObject = validStyles.reduce((prev, style) => {
```
denys-danyliuk, line 21: there is no need to create this variables
```suggestion
  return sourceString
    .replace(/\s\s+/g, ' ')
    .split(';')
    .filter(style => style.includes(':'));
    .reduce((prev, style) => {
```

denys-danyliuk (changes requested):

PR #1585, opened 2022-11-01
Solution

Esceype:


src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .map(str => str.trim())
+    .filter(str => str.length > 0)
```
Andriy-Shymkiv, line 13: what did you trying to check here i don't get it ??

Andriy-Shymkiv (approved):

src/convertToObject.js
Esceype: The filter has been made to escape empty elements in the resulting array

Andriy-Shymkiv: then you can write like this `.filter(item => item)`

PR #1586, opened 2022-11-01
Convert to object

maksym-kryvolap:


src/convertToObject.js
```diff
@@ -3,16 +3,41 @@
...
+
+  for (let i = 0; i < newStr.length; i++) {
+    if (newStr[i] !== ' ') {
+      stringStyles += newStr[i];
+    } else if (newStr[i] === ' ' && newStr[i - 1] === ':') {
+      stringStyles += newStr[i];
+    } else if (newStr[i] === ' ' && newStr[i + 1] !== ' '
+      && newStr[i - 1] !== ' ') {
+      stringStyles += newStr[i];
+    }
```
Bohdan-Maksymiuk, line 29: avoid such nesting `else if` conditions
This code is hard to read and maintain

Bohdan-Maksymiuk (changes requested):
also take a look over array iteration methods, array `split()` and `trim()` methods

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
-  // write your code here
+  const newStr = sourceString.replace(/\n/g, '');
+  const objectStyles = {};
+
+  const arrayStyles = newStr.split(';').filter(part => part.length > 2);
+
+  for (let i = 0; i < arrayStyles.length; i++) {
+    const styles = arrayStyles[i].split(':');
+
+    objectStyles[styles[0].trim()] = styles[1].slice(1).trim();
```
TarasHoliuk, line 25:
```suggestion
    objectStyles[styles[0].trim()] = styles[1].trim();
```
If I understood correctly `slice(1)` is doing the same as `trim()` here, so it could be removed

TarasHoliuk (approved):
GJ! Approved with one comment (check it, please)

PR #1588, opened 2022-11-01
Style to object solution

shtht24:


MikhailKucherina (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssStyles = {};
+
+  sourceString.split(';')
+    .map(style => style.split(':').map(item => item.trim()))
+    .filter(style => style.length === 2)
+    .map(style => (cssStyles[style[0]] = style[1]));
```
illiarerun, line 22: formatting tip
```suggestion
  sourceString.split(';')
    .map(style => style.split(':').map(item => (
      item.trim()
     )))
    .filter(style => style.length === 2)
    .map(style => (cssStyles[style[0]] = style[1]));
```

illiarerun (approved):

src/convertToObject.js
illiarerun, line 22: either use reduce or forEach for that, don't use `map` for mutations

illiarerun (changes requested):

src/convertToObject.js
shtht24: ty for review, i've fixed it:)

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssStyles = {};
+
+  sourceString.split(';')
```
denys-danyliuk, line 19:
```suggestion
  return sourceString
    .split(';')
```

denys-danyliuk (approved):

PR #1589, opened 2022-11-01
convertToObject

sh1gatsu:


viktoriia-ovcharenko (approved):
nice job

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+    const key = entries[0];
+    const value = entries[1];
+
+    if (value !== undefined) {
+      const trimmedProperty = value.trim();
+      const trimmedClass = key.trim();
+
+      styles[trimmedClass] = trimmedProperty;
+    }
+  });
```
illiarerun, line 32: don't use map to mutate anything, map is used to transform elements and create new array using them. use reduce to create new object

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssItems = sourceString.split(';');
+
+  const styles = {};
```
illiarerun, line 19:
```suggestion
```

illiarerun (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssItems = sourceString
+    .split(';')
+    .filter(removedBlankLines => removedBlankLines.includes(':'))
+    .map(item => item.split(':'));
+
+  return cssItems.reduce((value, entries) => {
```
denys-danyliuk, line 22:
```suggestion
  return sourceString
    .split(';')
    .filter(removedBlankLines => removedBlankLines.includes(':'))
    .map(item => item.split(':'));
    .reduce((value, entries) => {
```

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    const trimmedProperty = entries[0].trim();
+    const trimmedClass = entries[1].trim();
+
+    return {
+      ...value,
+      [trimmedProperty]: trimmedClass,
```
denys-danyliuk, line 28:
```suggestion
      [trimmedProperty]: trimmedClass,
```
u can trim them here without creating extra variables on each call of callback
also move callback outside of function to avoid creating it every time

denys-danyliuk (changes requested):

denys-danyliuk (approved):

denys-danyliuk (approved):

PR #1590, opened 2022-11-01
add solution

AlexTsimba:


src/convertToObject.js
```diff
@@ -1,18 +1,27 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedObject = {};
```
denys-danyliuk, line 4:
```suggestion
```

src/convertToObject.js
```diff
@@ -1,18 +1,27 @@
...
+
+  const sourseArr = sourceString
+    .replace(/;/g, '')
+    .split('\n')
+    .filter(item => item.includes(':')
```
denys-danyliuk, line 9: this is redundant

src/convertToObject.js
```diff
@@ -1,18 +1,27 @@
...
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedObject = {};
+
+  const sourseArr = sourceString
```
denys-danyliuk, line 6:
```suggestion
 return sourceString
```

src/convertToObject.js
```diff
@@ -1,18 +1,27 @@
...
+    .replace(/;/g, '')
+    .split('\n')
+    .filter(item => item.includes(':')
+    );
+
+  const assignProperty = (param) => {
+    const property = param
+      .split(':')
+      .map(prop => prop.trim());
```
denys-danyliuk, line 15: move them from here

src/convertToObject.js
```diff
@@ -1,18 +1,27 @@
...
+
+    convertedObject[property[0]] = property[1];
+
```
denys-danyliuk, line 18: also use reduce for creating new object

denys-danyliuk (changes requested):

src/convertToObject.js
```diff
@@ -1,18 +1,27 @@
...
+    return convertedObject;
+  };
+
+  sourseArr.forEach(assignProperty);
```
denys-danyliuk, line 22: move this to all other methods

illiarerun: and better use `reduce` for it

PR #1591, opened 2022-11-01
js_style-to-object

MXMUK:


Oleksandr-Stryzhavchuk (approved):
I think, good solution)

mykola-koval (approved):

PR #1595, opened 2022-11-01
style to object

sssolomia:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    .map(str => str.split(':')
+      .map(word => word.trim()))
+    .reduce(createObj, {});
 }
 
+const createObj = (prev, current) => {
+  return {
+    ...prev, [current[0]]: current[1],
+  };
+};
```
Largoxoxo, line 30: I'm not sure but I think it's better to create object before function initialization. Let's see what mentor will tell us.

Largoxoxo (approved):

src/convertToObject.js
mykola-koval: why?

mykola-koval (approved):

src/convertToObject.js
Largoxoxo: > why?

Nvm, my bad. Don't know why I thought that was a callback.

PR #1596, opened 2022-11-01
Sol

TarasV123:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  const modifiedStr = sourceString.split(';');
+  const mod2 = modifiedStr.map(x => x.split(':'));
```
```suggestion
  const modifiedStr = sourceString.split(';')
    .map(x => x.split(':'));
```

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+
+  for (const array of mod2) {
+    if (array[1]) {
+      result[array[0].trim(0)] = array[1].trim();
```
tetiana-ivanchenko, line 24:
```suggestion
      result[array[0].trim()] = array[1].trim();
```

tetiana-ivanchenko (changes requested):

src/convertToObject.js
```diff
@@ -16,8 +16,8 @@
 function convertToObject(sourceString) {
   const result = {};
 
-  const modifiedStr = sourceString.split(';');
-  const mod2 = modifiedStr.map(x => x.split(':'));
+  const modifiedStr = sourceString.split(';')
+    .map(x => x.split(':'));
```
TarasV123, line 20: I dont understand what is this ?  What is before this (.map.....) ???

AlisaVlasova: it`s same as `const modifiedStr = sourceString.split(';').map(x => x.split(':'));`

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  const modifiedStr = sourceString.split(';')
+    .map(x => x.split(':'));
+
+  for (const array of mod2) {
```
AlisaVlasova, line 22:
```suggestion
  for (const array of modifiedStr) {
```

AlisaVlasova (changes requested):

vlad-demchuk (approved):

PR #1597, opened 2022-11-02
Solution

alpopkov:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const correctStyles = {};
+
+  sourceString
+    .split(';')
+    .map(element => element.trim())
+    .filter(element => element !== '')
+    .map(element => {
```
Andriy-Shymkiv, line 23: don't use map to mutate anything, map is used to transform elements and create new array using them. use reduce to create new object.

that's what @illiarerun wrote to me in the review , so if you don't fix it, they will reject 😢

Andriy-Shymkiv (changes requested):

PR #1599, opened 2022-11-02
Solution

karyna-salnykova:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const propertiesArray = sourceString.split(';');
+  let cssProperties = {};
+
+  for (const property of propertiesArray) {
+    const pairOfInternals = property.split(':');
```
mitattt, line 21: you should use methods here

mitattt (approved):
i`ll approve but i guess you will get reject from menthor 🤔

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+
+    if (pairOfInternals[0] && pairOfInternals[1]) {
+      const prop = pairOfInternals[0].trim();
+      const value = pairOfInternals[1].trim();
+      const objOfInternals = {
+        [prop]: value,
+      };
+
+      cssProperties = Object.assign(cssProperties, objOfInternals);
```
Andriy-Shymkiv, line 30: agree with @mitattt . You should use reduce to create a new object. I understand, it's brain damage, because i was in the same situation as you  😭

Andriy-Shymkiv (changes requested):
i

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+    .split(';')
+    .map(string => string.trim())
+    .filter(string => string.length > 0)
+    .map(string => string.split(':')
+      .map(word => word.trim()))
+    .reduce((prev, current) => {
+      return {
+        ...prev, [current[0]]: current[1],
+      };
+    }, {});
```
denys-danyliuk, line 27: move callback outside of function to avoid creating it on each call

denys-danyliuk (approved):

PR #1602, opened 2022-11-07
Solution

SerhiiVovchenko:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrSourceString = sourceString.split(';');
+  const keyVelue = arrSourceString.map(value =>
```
qoa11a, line 18:
```suggestion
  const keyValue = arrSourceString.map(value =>
```
Look for typos and fix them. Velue => value.

qoa11a (changes requested):

polosanya (approved):
Great!

PR #1608, opened 2022-11-09
Solution

webdevw:


polosanya (approved):
Interesting solution! Good job!

PR #1611, opened 2022-11-10
Solution

Irina-Kulish:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    .split(';')
+    .filter(item => item.includes(':'))
+    .map(keyVal => {
+      return keyVal
+        .split(':')
+        .map(val => val.trim());
+    })
+    .reduce((acc, elem) => ({
+      ...acc,
+      [elem[0]]: elem[1],
```
vlad-demchuk, line 27: try not to cut names of variables (`val, res, elem`)

vlad-demchuk (changes requested):

nazarmatsevych (approved):

PR #1623, opened 2022-11-19
Solution

Nazar-Brunarskyi:


official-artem (approved):
Looks some difficult but, good job

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .replace(/\n/g, '')
+    .split(`;`)
+    .map(style => style.trim())
+    .filter(style => style)
```
solaryasha, line 21:
```suggestion
    .filter(Boolean)
```

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    .map(style => style
+      .split(`:`)
+      .map(item => item.trim())
+    )
+    .reduce((acc, css) => {
```
solaryasha, line 26:
```suggestion
    .reduce((acc, [key, value]) => {
```

solaryasha: Let's use an array destructuring

solaryasha (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    .split(`;`)
+    .map(style => style.trim())
+    .filter(Boolean)
+    .map(style => style
+      .split(`:`)
+      .map(item => item.trim())
+    )
+    .reduce((acc, [key, value]) => {
+      return {
+        ...acc,
```
solaryasha, line 28: I guess you could try `Object.assign` and mutate the object for better performance. But it's not critical

solaryasha (approved):
👍

PR #1627, opened 2022-11-21
stylesToObject

Mykola-Busarhin:


src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
+  for (const line of keysAndMean) {
+    for (let i = 0; i < line.length; i++) {
+      const key = line[0];
+      const meaning = line[1];
+
+      if (line[i + 1] !== undefined) {
+        stylesObject[key.trim()] = meaning.trim();
+      }
+    }
+  }
```
stanislav-topikha, line 19: try to simplify this part

stanislav-topikha (changes requested):

src/convertToObject.js
Mykola-Busarhin: This is the shortest solution I`ve come up with. If there is a shorter one, please, give me a hint)

stanislav-topikha (approved):
GJ 👍🏻

PR #1628, opened 2022-11-21
Solution

Yevhenii-Stanchenko:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * and values are the values of related CSS properties(see an examplin[
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObj = {};
+  const stylesToArr = sourceString
```
vlad-demchuk, line 16: check names of variables. 
try not to cut and use something better than `el`

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    .split(';')
+    .map(el => el.trim())
+    .filter(el => el.length > 0);
+
+  stylesToArr.forEach((el) => {
+    const styleToArr = el.split(':');
+
+    const [key, value] = styleToArr;
+
+    styleObj[(key.trim())] = (value.trim());
```
vlad-demchuk, line 26: do you need parenthesis here?

vlad-demchuk (changes requested):

TarasHoliuk (approved):

PR #1631, opened 2022-11-24
Implementation

UmizDemud:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  let runningStr = '';
+
+  sourceString.split(/\r?\n/).forEach(el => {
```
mgrinko, line 20: please use `map` `filter` and `reduce` methods to get some practice with them
Currently you are just using `forEach` as a for loop

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+    runningStr = '';
+
+    if (el.length) {
```
mgrinko, line 23: this is `filter` method

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+      runningStr = el.split(':');
```
mgrinko, line 24: this is `map`

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+      if (runningStr.length < 2) {
+        return;
+      }
+
+      result[runningStr[0].trim()] = runningStr[1]
+        .substring(0, runningStr[1].length - 1).trim();
+    }
+  });
+
+  return result;
```
mgrinko, line 35: Just for practicing try to do all the code in one return
```js
return sourceString.split('\n')
  ...
```

mgrinko (changes requested):

meforoma (approved):
advanced 👍

PR #1634, opened 2022-11-26
style-to-object

mykolahuryn:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  const paramArray = sourceString.split(';');
+
+  for (const param of paramArray) {
+    const key = param.split(':')[0];
+    const val = param.split(':')[1];
```
polosanya, line 23:
```suggestion
    const value = param.split(':')[1];
```
It is better to use full, descriptive names, especially if it's about saving only 2-3 chars

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comment

PR #1639, opened 2022-11-28
solution

lefteris-L:


UmizDemud (approved):
I like it, separating by ';' is seems to be the right approach

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    const value = pair.slice(pair.indexOf(':') + 1).trim();
+
+    if (key.length && value.length) {
+      pairs[key] = value;
+    }
+  };
+
+  properties.map(getPairs);
+
+  return pairs;
```
naziamoff, line 31: please use `map` `filter` and `reduce` methods to get some practice with them

naziamoff (changes requested):

naziamoff (approved):
🔥

PR #1641, opened 2022-11-29
solution

ysrckr:


UmizDemud (approved):
Seems like the code is working, with additional checks. Can't see a negativity about efficiency angle on first sight

meforoma (approved):
you chose the complex / advanced path.
but it can be muuuch simpler.
Yet it does work.
And since there's no ultimate right solution - ✅

PR #1643, opened 2022-11-29
solution

nurdanokcu:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
-  // write your code here
+  const stylesArray = sourceString
+    .split(';')
+    .map(el => el.trim())
+    .filter(el => el.length > 0);
+
+  const stylesObj = {};
+
+  stylesArray.map(style => {
+    let eachPair = style.split(':');
```
UmizDemud, line 25: This can be done in one line but not a big deal for me, as it is more readable

UmizDemud (approved):
I ran some tests, to see if defining eachPair in the loop of the map function is a good idea.

let eachPair in loop
vs
const eachPair in loop
vs
let eachPair above loop

Your implementation did a great job regardless. Here are the tests:

image

src/convertToObject.js
nurdanokcu: I actually didn't think about doing it 
**_const eachPair = style.split(':')
      .map(pair => pair.trim());_**
this way. As I was doing the task in steps. 

I think it is better since it is one less variable declaration. 

Thank you Umut.

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString
+    .split(';')
+    .map(el => el.trim())
+    .filter(el => el.length > 0);
+
+  const stylesObj = {};
+
+  stylesArray.map(style => {
```
meforoma, line 24: good logic 👍 
i'd point out few things although:
- `.map` method returns a new array, so usually it's used when you actually need a new array, otherwise if you just iterate through an array and don't need a new one - better go with `.forEach` - it's an actual method purpose)

meforoma: and essentially i think you would benefit from `.reduce` method that would save you lines of code:
```
  return stylesArray.reduce((stylesObj, style) => {
    const [styleKey, styleValue] = style.split(':')
      .map(piece => piece.trim());

    stylesObj[styleKey] = styleValue;

		return stylesObj;
  }, {})
```

meforoma (approved):
really nice 👏 
please see few comments that would make code a bit better

meforoma (approved):
🎉

PR #1649, opened 2022-12-02
convertToObject function is implemented

Oleksii-Mishchenko:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    .map(propValPair => propValPair.split(':'))
+    .reduce((prev, curr) => {
+      const cssProperty = curr[0].trim();
+      const cssValue = curr[1].trim();
+
+      return {
+        ...prev,
+        [cssProperty]: cssValue,
+      };
+    }, {});
```
Anastasiia-Svintsova, line 29: In the future it is better to avoid such nesting and use variables

PR #1652, opened 2022-12-03
Develop

SalehBUD:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+  const stylesObject = {};
+
+  sourceString
+    .split(';')
+    .map((style) => style.trim())
+    .filter((style) => style.length > 0)
+    .forEach((style) => {
+      const styleArr = style.split(':');
+
+      stylesObject[styleArr[0].trim()] = styleArr[1].trim();
```
nurdanokcu, line 26:
```suggestion
      const [styleKey, styleValue] = style.split(':');

      stylesObject[styleKey.trim()] = styleValue.trim();
```

nurdanokcu: For better readability.

nurdanokcu (approved):
Your solution made it look like very easy task. 👏

src/convertToObject.js
meforoma: @SalehBUD let's apply @nurdanokcu's suggestion - it's a common good practice.

meforoma (changes requested):
please apply the comment

meforoma (approved):
nnnice 👍

PR #1654, opened 2022-12-03
DiDemkiv solution

DiDemkiv:


src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+  const sourceStringArr = sourceString.split(';');
+
+  for (const rule of sourceStringArr) {
```
batstolya, line 12: CL-5 [CODE PRACTICE] - Avoid using loops in this task, use iteration methods instead.
you can use for example the .map method

src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
+    if (!rule.includes(':')) {
+      continue;
+    }
```
batstolya, line 15: in this case you can use the filter method

batstolya (approved):

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+
+  sourceString.split(';')
+    .filter(rule => rule.includes(':'))
+    .forEach(rule => {
```
ivan-kadykalo, line 13: We can improve this algorithm and make this just in 1 iteration. Let's do it.

ivan-kadykalo (changes requested):
You are almost there 🙃

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+
+  sourceString.split(';')
+    .forEach(rule => {
```
solaryasha, line 12: Let's practice reduce for the object creation

solaryasha (changes requested):

solaryasha (approved):
Good job!

PR #1657, opened 2022-12-04
Solution

llllewvvaa:


Anastasia4sik (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+    .map(property => {
+      const splitted = property.trim().split(':');
+
+      return splitted.map(key => {
+        return key.replace(';', '').trim();
+      });
+    })
+    .forEach(property => {
+      formatedStyle[property[0]] = property[1];
+    });
```
ivan-kadykalo, line 30: You don't need 3 iterations to do this. Let's improve this algorithm and make the same just in one iteration

ivan-kadykalo (changes requested):
You are almost there 💪

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .reduce((accumulator, property) => {
+      const splittedString = property.split(':');
+      const propertyName = splittedString[0];
+      const propertyValue = splittedString[1];
```
solaryasha, line 20: Use destructuring for this case
```suggestion
      const [key, value] = property.split(':');
```

solaryasha (changes requested):

solaryasha (approved):
👍

PR #1661, opened 2022-12-04
Solution

illiashcherbak:


src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedStyles = sourceString
+    .split(';')
+    .map(item => item.trim())
+    .filter(item => item !== '')
+    .map(item => item.split(':').map(value => value.trim()))
+    .reduce((prev, curr) => ({
```
KrisPlanka, line 14: you should move second .map to the new line

KrisPlanka (approved):
Looks good

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedStyles = sourceString
+    .split(';')
+    .map(item => item.trim())
+    .filter(item => item !== '')
+    .map(item => item.split(':')
+      .map(value => value.trim()))
+    .reduce((prev, curr) => ({
+      ...prev,
```
ivan-kadykalo, line 16: Wow, 5 iterations. Not the best performance, let's optimize this algorithm and made it with only 1 iteration.

illiashcherbak: I was only able to remove one map method. I have no idea how I can do this in one iteration

ivan-kadykalo: @illiashcherbak imagine if you will have such a task, to optimize the productivity of some solution, what will you do? 
Saying "I have no idea" is not the best solution. Why didn't you come to Q&A if you don't know how to do it?

I will wait for you on my Q&A tomorrow at 13:30 and we will discuss how to resolve it in 1 iteration.

ivan-kadykalo (changes requested):
You need to learn how to solve tasks or ask if you don't know how.
Will wait for you on my Q&A tomorrow.

PR #1666, opened 2022-12-05
solution v1

Lun4trick:


Alireza-Mohammadhossein (commented):
Good job!
Nice code with createObjectProp

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+
+  separateElements.map(elem => {
+    if (elem.length > 1) {
+      createObjectProp(elem[0], elem[1], result);
+    }
+  });
+
+  function createObjectProp(key, value, obj) {
+    obj[key] = value;
+  }
```
meforoma, line 32: it's nice you're introducing helper function, yet it's too short to be honest
i'd rather do it inside the loop right away - it would be simply shorter
also better de-structure elem instead of using `elem[0] / elem[1]` - it will be more explanatory
```suggestion
  separateElements.map(([cssKey, cssValue]) => {
    if (elem.length > 1) {
      result[cssKey] = cssValue;
    }
  });
```

meforoma (changes requested):
please see the comment

src/convertToObject.js
Lun4trick: Ah yeah its true its too short to create a function for it, i love the helper functions maybe too much :D I fixed it, thanks for the suggestion! :)

meforoma (approved):
yepp 👏

PR #1667, opened 2022-12-05
Some sort of solution

4ebupel:
I miss working on my landind project((

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -3,16 +3,26 @@
 /**
  * Implement convertToObject function:
  *
- * Function takes a string with styles (see an example in [stylesString.js](./stylesString.js))
+ * Function takes a string with styles (see an example in
+ * [stylesString.js](./stylesString.js))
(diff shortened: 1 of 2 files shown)


anastasiia-roiek (approved):
nice!

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .filter(style => style.includes(':'))
+    .map(style => style.split(':'))
+    .reduce((key, style) => {
+      key[style[0].trim()] = style[1].trim();
```
ivan-kadykalo, line 22: You have used here 3 iterations, it is not the best solution from the optimization point. Let's improve the algorithm, and made the same in one iteration.

ivan-kadykalo (changes requested):
You are almost there 🙃

PR #1674, opened 2022-12-05
solution

kenvastolord:


src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
+    .split('\n')
+    .join('')
+    .trim()
+    .split(';')
+    .map(el => el.split(':'));
+
+  const result = {};
+
+  cleanedUpString.forEach(element => {
+    if (element.length > 1) {
```
Grayfox147, line 27:
```suggestion
   .filter(element=> element.includes(':'))
   and then =>
   .map(el => el.split(':'));
```
You could implement a filter iteration with a rule that will allow us to work only with elements that have ':' 
So basically you could implement this filter from the beginning and then you wouldnt need to worry about aplying the if statement for each element

src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
+      const [cssProperty, cssValue] = element;
+
+      const objectKey = cssProperty.trim();
+      const objectValue = cssValue.trim();
+
+      result[objectKey] = objectValue;
```
Grayfox147, line 33:
```suggestion
      const objectKey = cssProperty.trim();
      const objectValue = cssValue.trim();

      result[cssProperty.trim()] = cssValue.trim();
```

It would save you some coding just to implement it directly in the result

Grayfox147 (commented):
Overall amazing work, and really good and smart implementation!

meforoma (approved):
passable ✅

PR #1676, opened 2022-12-05
Solution

Sveta-Kryukova:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const newArray = sourceString.split(';')
```
TheOneThatKnowsNothing, line 17: You can just write return and don't create new variable, + wrong variable name since result of the func has type object

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+    .map((element) => element.split(':'))
+    .filter((word) => (word.length > 1))
+    .reduce((key, value) => {
```
TheOneThatKnowsNothing, line 20: You could use array  destructuring for parameters of callback for better readability

solaryasha (approved):
👍

PR #1677, opened 2022-12-05
Solution

kattentionn:


anastasiia-roiek (approved):
nice!

src/convertToObject.js
```diff
@@ -3,16 +3,21 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .map(item => item.split(':'))
+    .filter(item => item.length > 1)
+    .reduce((style, value) => {
+      style[value[0].trim()] = value[1].trim();
```
ivan-kadykalo, line 17: We can improve this algorithm and make it just in 1 iteration, instead of 3. Let's do it.

ivan-kadykalo (changes requested):
You are almost there 🙃

PR #1684, opened 2022-12-05
Solution

alyonkaradchuk:


src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
+  const properties = sourceString
+    .split(';')
+    .map(style => style.trim())
+    .filter(style => style.length > 0)
+    .map(style => style.split(':'))
+    .reduce((prev, style) => ({
+      ...prev,
+      [style[0].trim()]: style[1].trim(),
+    }), {});
+
```
viktoriia-kravchenko, line 18: We can use [key, value] destructuring instead of style here.

viktoriia-kravchenko (approved):
🔥

Ihor-Karpyn (approved):

PR #1693, opened 2022-12-06
Add solution

VladimirEdelmann:


src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .filter(value => value.includes(':'))
+    .map(value => value.split(':'))
+    .reduce((key, value) => {
+      key[value[0].trim()] = value[1].trim();
```
Nazar-Brunarskyi, line 18: change `value `to `[key, value]` it will make your code easier to read

Nazar-Brunarskyi (approved):

src/convertToObject.js
solaryasha, line 18: Let's use destructuring like these
```suggestion
    .reduce((styles, [key, value]) => {
      styles[key.trim()] = value.trim();
```

solaryasha (approved):

PR #1694, opened 2022-12-06
Solution

YanaEvaGonchar:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+  const stringToCSS = sourceString
+    .split(';')
+    .map(elem => elem.trim())
+    .filter(elem => elem !== '')
+    .map(elem => elem.split(':').map(property => property.trim()))
+    .reduce((prev, elem) => {
+      return {
+        ...prev, [elem[0]]: elem[1],
+      };
+    }, {});
```
anastasiia-bilska, line 26: Great! But I was advised to use only one iteration instead of several. Try to do the same

anastasiia-bilska (changes requested):

src/convertToObject.js
YanaEvaGonchar: Can you please explain what you mean? Because I don't really understand how to do it right now😞

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+    const [key, value] = elem.split(':');
+
+    if (value !== undefined) {
+      return {
+        ...prev,
+        [key.trim()]: value.trim(),
+      };
+    } else {
+      return prev;
+    }
```
valerii-derkach, line 29:
```suggestion
    } 
      return prev;
    
```

valerii-derkach (approved):

PR #1699, opened 2022-12-06
solution

korchovskyi:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  */
 function convertToObject(sourceString) {
   // write your code here
+  const settings = sourceString
+    .split(';')
+    .filter(style => style.includes(':'))
+    .map(style => style.split(':'))
+    .reduce((prev, style) => ({
+      ...prev,
+      [style[0].trim()]: style[1].trim(),
```
batstolya, line 25: You can use array destructuring, it's optional but looks better
```suggestion
    .reduce((prev, [key, value]) => ({
      ...prev,
      [key.trim()]: value.trim(),
```

korchovskyi: thanks, fixed

valerii-derkach (approved):

PR #1704, opened 2022-12-06
solution

mmahfuzi:


src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const styles = {};
+
+  sourceString.split(';').map(line =>
+    line.trim()).filter(line => line)
+    .map(lns => {
```
nurdanokcu, line 20: sourceString.split(';')
    .map(line =>line.trim())
    .filter(line => line)
    .map(lns => {

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const styles = {};
+
+  sourceString.split(';').map(line =>
+    line.trim()).filter(line => line)
```
nurdanokcu, line 19: I couldn't understand the purpose of the filter method here because it returns the same lines back.

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+    .map(lns => {
+      const [property, value] = lns.split(':');
```
nurdanokcu, line 21: I think **lns** is a bit unclear variable name. I liked the usage of destructuring.

src/convertToObject.js
nurdanokcu, line 20: After using the filter method there is a new array , the purpose of the task is taking each of the items from that  array and putting them into styles object. For this purpose .forEach method can be used instead of map. Because .map method returns a new array, so usually it's used when you actually need a new array .forEach method is great for when you need to do some action with each element of the array.

nurdanokcu (changes requested):
Nice and clear solution. 
I added some suggestions. Please see the comments.

src/convertToObject.js
mmahfuzi: I thought since we are not supposed to use loops in these part tasks, I used .map

src/convertToObject.js
nurdanokcu: Oh that is smart . I see now.

nurdanokcu (approved):
:rocket:

meforoma (approved):
🔥

PR #1705, opened 2022-12-07
Solution

Rijo-J:


grslckr (approved):

meforoma (changes requested):
review the suggested approach and let me know if it feels more optimal 🙂

Rijo-J:
> review the suggested approach and let me know if it feels more optimal 🙂

Definitely muuuch better! Idk why my brain works like that, I guess it goes with practice! thank you! 😁

Rijo-J:
> > review the suggested approach and let me know if it feels more optimal 🙂
> 
> Definitely muuuch better! Idk why my brain works like that, I guess it will get there with more practice! thank you! 😁

@meforoma :wave: Hi! I've rolled back the reduce callback since it was throwing some error and not working for some reason. Still tried to figure it out but I didn't want to loose 2 much time there.

meforoma (approved):
👍

PR #1709, opened 2022-12-07
solution51

dolnaro:


Grayfox147 (commented):
Really good implementation of methods! to fix your npm lint issues I would recommend to move the content of the line to the next one so you would not exceed the maximum comment line! 
Besides that everything is great!

meforoma (approved):
sweet 👍

PR #1712, opened 2022-12-07
completed

alaaashkar:


UmizDemud (approved):
👍  well done

naziamoff (changes requested):
well done, but let's practice reduce method here

naziamoff (approved):

PR #1714, opened 2022-12-07
converting CSS to js Object

Kavster92:


Rijo-J (approved):
Looks great imo! 😁 🚀 passing to mentors

naziamoff (changes requested):
well done, but let's practice reduce method here

naziamoff (approved):

PR #1715, opened 2022-12-07
Finished

jackelashkar:


src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  sourceString.split(';')
+    .filter(rule => rule.includes(':'))
```
CharAnagn, line 20: I like the filter method to exclude the empty strings

CharAnagn (approved):
nice good job 🔥

PR #1726, opened 2022-12-09
Solution

IrynaK-a:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+ * and values are the values of related CSS properties
+ *  (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
polosanya, line 17: `result` is not the best name. Let's try to use more meaningful names

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+  const cssProperties = sourceString.split(';');
+
+  cssProperties.forEach(prop => {
+    const array = prop.split(':');
```
polosanya, line 21: image

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    const propertyName = String(array[0]).trim();
+    const propertyValue = String(array[1]).trim();
+
+    result[propertyName] = String(propertyValue);
```
polosanya, line 25: Do you really need to specify that it's string everywhere?

polosanya (changes requested):
Almost perfect! Let's provide some changes

PR #1727, opened 2022-12-09
first commit

haduigon:


src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
+      )
+    )
+    .filter(string => string.length > 1)
+    .map(string => string.map(string2 => string2.trim()))
+    .reduce((a, x) => {
+      a[x[0]] = x[1];
+
+      return a;
+    }, {}
+    );
```
AlisaVlasova, line 35: try to simplify solution
split, map, filter, reduce will be enough

AlisaVlasova (changes requested):

src/convertToObject.js
haduigon: I have done my best ) please, check it out )

vlad-demchuk (approved):

PR #1732, opened 2022-12-12
Solution

s3rhiienko:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const finalObject = {};
+
+  sourceString.split(';')
+    .map(x => x.trim())
+    .filter(x => x.length > 1)
+    .map(x => x.split(':'))
```
polosanya, line 22: Let's use more informative names than just `x` everywhere

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const finalObject = {};
```
polosanya, line 17:
```suggestion
  const objectWithStyles = {};
```
IMHO, it is better to clarify what is stored in the object

polosanya (changes requested):
Almost done! Let's just improve your naming a bit)

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectWithStyles = {};
+
+  sourceString.split(';')
+    .map(x => x.trim())
+    .filter(style => style.length > 1)
+    .map(x => x.split(':'))
```
Anastasiia-Svintsova, line 22: Would be great to rename `x` everywhere

Anastasiia-Svintsova (approved):

PR #1733, opened 2022-12-13
Solution

RomaSheva1987:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const result = {};
+  const sourseArray = sourceString.split(';');
+  const WihtoutSpaces
```
tetiana-ivanchenko, line 20: variable name shouldn’t be capitalised

tetiana-ivanchenko (approved):

PR #1735, opened 2022-12-14
Initial commit

yaroslav-furhala:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayCount = sourceString.split(';').map(x => x.split(':'));
+
+  const finish = {};
+
+  arrayCount.forEach(a => {
```
DarkMistyRoom, line 21:
```suggestion
  arrayCount.forEach(pair => {
```

DarkMistyRoom: it will be better to understand what contains in each element

DarkMistyRoom (approved):

PR #1736, opened 2022-12-14
Solution

FallenMAD:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const words = sourceString.split(';');
+
+  const myObj = {};
```
Anastasiia-Svintsova, line 19: Better to use meaningful variables names
```suggestion
  const styleObject = {};
```

Anastasiia-Svintsova (approved):

PR #1770, opened 2023-01-02
Develop: Finished the function

sonyaxent:


src/convertToObject.js
```diff
@@ -1,18 +1,18 @@
...
-  // write your code here
-}
+  const keyValuePairs = sourceString.split(';')
+    .map(line => line.trim())
+    .filter(key => key.length > 1)
+    .map(solution => solution.split(':'));
+
+  const result = {};
+
+  keyValuePairs.map(([key, value]) => {
```
agniya-i, line 11: `map` is usually used to create new array, in this case, it would be better to change it to forEach or try to use reduce

agniya-i (changes requested):

SanyaBratashchuk (approved):

PR #1777, opened 2023-01-04
string was converted

StetsenkoViktoriia:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+
+  sourceString.split(';')
+    .map(element => element.trim())
+    .filter(element => element.length > 0)
+    .map(element => {
```
OlEzhA0, line 22: use for each or reduce and return reduced object at once

OlEzhA0 (changes requested):
Please check my comment

loralevitska (approved):
👍

PR #1781, opened 2023-01-05
Solution

tania-boiar:


Avichez (approved):
Wow, this is a shortest solution I've seen 🔥 
But readability of code a little bit difficult.

tania-boiar:
> Wow, this is a shortest solution I've seen fire But readability of code a little bit difficult.

Thanks! 🙌 
I still hope that this solution is understandable enough not to be rewritten.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString.split(';')
+    .reduce((prevStyle, style) => {
```
```suggestion
  const styles = sourceString
    .split(';')
    .reduce((prevStyle, style) => {
```

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+      if (style.split(':')[1]) {
+        return {
+          ...prevStyle,
+          [style.split(':')[0].trim()]: style.split(':')[1].trim(),
+        };
```
Anna-Tkachenko, line 23: you split `style` 3 times in one check
can we save the result of split in some variable and use it where we need it?

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #1785, opened 2023-01-06
Implement convertToObject function - completed

GarryRocksman:
Have a nice day !!!

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -3,16 +3,33 @@
 /**
  * Implement convertToObject function:
  *
- * Function takes a string with styles (see an example in [stylesString.js](./stylesString.js))
+ * Function takes a string with styles (see an example
+ * in [stylesString.js](./stylesString.js))
  * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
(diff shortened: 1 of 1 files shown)


oleh-holovnykh (approved):
Well done! :)

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedStyles = {};
+
+  const cleanStyleList = sourceString
+    .split(';')
+    .map(item => item
+      .replace(/\n/g, '')
+      .trim())
```
Anna-Tkachenko, line 23:
```suggestion
    .map(item => item
      .replace(/\n/g, '')
      .trim()
    )
```

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+    .filter(item => item !== '');
+
+  // filling a result Object with keys and values
+  cleanStyleList
+    .forEach(item => (
+      convertedStyles[item.split(':')[0].trim()]
+        = item.split(':')[1].trim()));
```
Anna-Tkachenko, line 30:
```suggestion
  cleanStyleList.forEach(item => (
    convertedStyles[item.split(':')[0].trim()]
      = item.split(':')[1].trim()
  ));
```

Anna-Tkachenko: will be better to go though items and split them initially, and then go through the splitted elements and set them to corresponding values

Anna-Tkachenko (approved):

PR #1792, opened 2023-01-06
solution

yaroslavna-horban:


nzrii-kysel (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+  const result = {};
+
+  arrFromString.map(item => item.trim())
+    .filter(item => item !== '')
+    .map(item => item.split(':'))
+    .forEach(item => {
+      const [key, value] = item;
+
+      result[key.trim()] = value.trim();
+    });
```
Denchepornyuk, line 25: instead of `forEach` and `result` variable you can use `reduce`

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1793, opened 2023-01-06
Solution

victorionov:


src/convertToObject.js
```diff
@@ -1,18 +1,21 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString.split(';').map(style =>
+    style.split(':').map(styleRule => styleRule.trim()));
```
sv-dubov, line 5:
```suggestion
  const stylesArray = sourceString
    .split(';')
    .map(style => style.split(':')
      .map(styleRule => styleRule.trim()));
```

sv-dubov (approved):
Looks good!

src/convertToObject.js
Denchepornyuk: agree, each method should be on the new line

Denchepornyuk (approved):

PR #1794, opened 2023-01-06
solution

Oleksandr-Maly:


sv-dubov (approved):
Good job! But maybe instead of forEach try to use reduce

alexandra-protyanova (approved):

PR #1796, opened 2023-01-07
solution

imariash86:


src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const stylesObj = {};
```
anna-inozemets, line 18: Maybe it is possible to find more suitable name of variables instead of using parts like "Obj", "Arr"

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+  const stylesArr = sourceString.split(';');
+  const clearStyles = stylesArr
```
anna-inozemets, line 20: I think you can unite this code and create only one variable

anna-inozemets (approved):
Pretty good)

src/convertToObject.js
alexandra-protyanova: Sometimes it is better to create more variables, to have clearer understanding what is going in code

src/convertToObject.js
alexandra-protyanova: agree

alexandra-protyanova (approved):

PR #1797, opened 2023-01-07
solution

vasyliev-anton:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+
+    const key = element[0].trim();
+    const value = String(element[1]).trim();
+
+    if (value === 'undefined') {
+      continue;
+    }
+
+    result[key] = value;
+  }
```
SanyaBratashchuk, line 32: use reduce or forEach

SanyaBratashchuk (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const result = {};
+  const stringArray = sourceString.split(';');
+  const keysAndValues = stringArray.map(element => element.split(':'));
+
+  const getProp = (str) => {
+    const key = str[0].trim();
+    const value = String(str[1]).trim();
```
DarkMistyRoom, line 23:
```suggestion
    const value = str[1].trim();
```

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+
+    result[key] = value;
+  };
+
+  keysAndValues.forEach(prop => prop.length > 1 ? getProp(prop) : result);
```
DarkMistyRoom, line 28: Why do you return the result here? what's the point?

DarkMistyRoom (changes requested):

agniya-i (approved):

PR #1803, opened 2023-01-08
Solution

SashaViun:


sv-dubov (approved):
Looks good! But maybe instead of forEach and result variable try to use reduce

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrFromString = sourceString.split(';');
+  const result = {};
+
+  arrFromString.map(item => item.trim())
```
alexandra-protyanova, line 18: i think it would work in the same way without first `map`

alexandra-protyanova (approved):

PR #1808, opened 2023-01-08
solution

vladyslav78292:


sv-dubov (approved):
Looks good!

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  sourceString
+    .split(';')
+    .map(style => style.trim())
+    .filter(style => style.includes(':'))
+    .forEach(style => {
+      const [key, value] = style.split(':');
+
+      stylesResult[key.trim()] = value.trim();
+    });
```
Denchepornyuk, line 27: use reduce instead of `stylesResult` variable and `forEach` method

Denchepornyuk (changes requested):

PR #1811, opened 2023-01-09
task was solved

ihor-soloviov:


Avichez (approved):
Wow, that's a great solution 👍
I don't know what about the naming of variables but in general all clear and understandable.

alexandra-protyanova (approved):

PR #1820, opened 2023-01-10
solution added

artem-sobolevskyi:


nkondd (approved):
Good job!

src/convertToObject.js
```diff
@@ -12,7 +12,17 @@
...
+    .filter((part) => part.includes(':'))
+    .map(part => {
+      const [key, value] = part.split(':');
+
+      return {
+        [key.trim()]: value.trim(),
+      };
+    });
+
+  return Object.assign({}, ...arrOfStylesObjects);
```
alexandra-protyanova, line 25: it isn't the best idea to use `Object.assign` with arrays

alexandra-protyanova (changes requested):

ArturErmolenko (approved):

PR #1821, opened 2023-01-11
Finish task

U-Ra-All:


src/convertToObject.js
```diff
@@ -1,18 +1,31 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const stylesObject = {};
+
+  sourceString.split(";").forEach((style) => {
+    if (style.trim().length > 1) {
+      stylesObject[style.slice(0, style.indexOf(":")).trim()] = style
+        .slice(style.indexOf(":") + 1)
```
OlEzhA0, line 23: why do we need slice and check with `indexOf`
you have string value, with `:` separator
```
const [key, value] = style.split(': ')
```

OlEzhA0: would be better if you try to rewrite this with `.reduce` method

OlEzhA0 (changes requested):
Hey! check please my comments

src/convertToObject.js
U-Ra-All: Thanks, done!

mykhalenych (approved):

PR #1824, opened 2023-01-14
Solution

AV1788:


src/convertToObject.js
```diff
@@ -3,16 +3,44 @@
...
+    }
+  }
+
+  const filtered = mass.filter(function(elem) {
+    if (elem !== '') {
+      return true;
+    } else {
+      return false;
+    }
+  });
```
TarasHoliuk, line 37:
```suggestion
  const filtered = mass.filter(Boolean);
```

`''` empty string will be transformed to `false`, because `Boolean('') === false` and vice versa

src/convertToObject.js
```diff
@@ -3,16 +3,44 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  const mass = [];
```
TarasHoliuk, line 18: bad names. They should be meaningful

src/convertToObject.js
```diff
@@ -3,16 +3,44 @@
...
+  const arrSplitFirstLevel = sourceString.split(';');
+
+  for (const words of arrSplitFirstLevel) {
+    const arrSplitSecondLevel = words.split(':');
+
+    for (const values of arrSplitSecondLevel) {
```
TarasHoliuk, line 24: avoid nested loops

TarasHoliuk (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,41 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectForStyles = {};
+  const array = [];
+  const arrSplitFirstLevel = sourceString.split(';');
+
+  for (const words of arrSplitFirstLevel) {
+    const arrSplitSecondLevel = words.split(':');
```
VitaliyBodnar, line 22: make sense to use destructuring here
```suggestion
    const [name, value] = words.split(':');
```

src/convertToObject.js
```diff
@@ -3,16 +3,41 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectForStyles = {};
+  const array = [];
```
VitaliyBodnar, line 18: it's bad practice to use 'general names' for variables, like a var, function, array, etc.
better to rename it list or other

src/convertToObject.js
```diff
@@ -3,16 +3,41 @@
...
+  for (const words of arrSplitFirstLevel) {
+    const arrSplitSecondLevel = words.split(':');
+
+    if (arrSplitSecondLevel[0] !== undefined
+      && arrSplitSecondLevel[1] !== undefined) {
+      const propertyName = arrSplitSecondLevel[0].trim();
+      const propertyValue = arrSplitSecondLevel[1].trim();
+
+      array.push(propertyName);
+      array.push(propertyValue);
```
VitaliyBodnar, line 30: method push can use multiple params 
```suggestion
      array.push(propertyName, propertyValue);
```

VitaliyBodnar (changes requested):
Almost done!

VitaliyBodnar (approved):
Good job!

PR #1831, opened 2023-01-18
styleToObject

Oleksandr-Maslo:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const stringsObject = {};
+  const subStrings
+    = sourceString.split(';').filter(
+      item => item.includes(':') && item.length > 1);
+
+  for (let i = 0; i < subStrings.length; i++) {
+    const key = subStrings[i].split(':')[0];
+    const property = subStrings[i].split(':')[1];
```
```suggestion
    const [key, property] = subStrings[i].split(':');
```

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stringsObject = {};
+  const subStrings
+    = sourceString.split(';').filter(
+      item => item.includes(':') && item.length > 1);
```
nazarmatsevych, line 20:
```suggestion
  const subStrings = sourceString
    .split(';')
    .filter(item => item.includes(':') && item.length > 1);
```

nazarmatsevych (changes requested):
Good work,
just small fixes in comments

nazarmatsevych (approved):

PR #1832, opened 2023-01-18
Solution

pushkarskiyrodion:


src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
+ * [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';')
+    .filter((element) => element.includes(':'))
```
Volodymyr-Melnyk-Fron10der, line 18: Good work! But if you use reduce - there is no need to use filter before, you can put this condition in reduce

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
+    .reduce((accum, curr) => {
+      const [key, value] = curr.split(':');
```
Volodymyr-Melnyk-Fron10der, line 20:
```
if (curr.includes(':')) {
    const [key, value] = curr.split(':');
}

 return accum;
```

Volodymyr-Melnyk-Fron10der (changes requested):
Make a little fix here

DarkMistyRoom (approved):

PR #1837, opened 2023-01-23
Solution

kirill-vypirovskyi:


src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
+    .split(';')
+    .forEach(element => arrayWithoutSpaces.push(element.trim()));
+
+  const clearArray = [];
+
+  arrayWithoutSpaces
+    .filter(elem => elem.length > 0)
+    .forEach(element => clearArray.push(element.trim().split(':')));
+
+  const result = {};
```
stanislavcodes, line 30: Better use different name for this variable to increase code readability

stanislavcodes (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayWithoutSpaces = [];
+
+  sourceString
+    .replace(/\n/g, '')
+    .split(';')
+    .forEach(element => arrayWithoutSpaces.push(element.trim()));
```
OleksiiBlezniuk, line 22: Why to push in empty array using `forEach` if result of split - is an array? =)

src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
+
+  const clearArray = [];
```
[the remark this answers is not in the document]
OleksiiBlezniuk: I think this array is redundant, why can't we use other methods on previous method result? It's okay to chaining methods

src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
+
+  arrayWithoutSpaces
+    .filter(elem => elem.length > 0)
+    .forEach(element => clearArray.push(element.trim().split(':')));
+
+  const result = {};
+
+  for (const item of clearArray) {
+    result[item[0].trim()] = item[1].trim();
+  }
```
OleksiiBlezniuk, line 34: Let's learn how to use array methods instead of loops (reduce is a great option here)

OleksiiBlezniuk (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resultingObject = {};
+
+  sourceString
+    .split(';')
+    .filter(element => element.trim())
+    .forEach(element => {
```
sergii-nosachenko, line 22: The `forEach` method is the regular loop. Use reduce here to accumulate props to object.

sergii-nosachenko (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .filter(element => element.trim())
+    .reduce((object, element) => {
```
```suggestion
    .reduce((styleAsObject, element) => {
```

sergii-nosachenko (approved):
Good work 👍

PR #1838, opened 2023-01-25
Task Done

Sergei-Nikolaienko:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+  const values = sourceString.split(';');
+  const resultObject = {};
+
+  values.forEach(value => {
+    const property = value.trim().split(':').map(element => element.trim());
+
+    if (!property.includes('')) {
+      resultObject[property[0]] = property[1];
+    }
+  });
```
Bohdan-Maksymiuk, line 27: try to use `reduce()` here

Bohdan-Maksymiuk (approved):
Good job 👍 
You may practice with reduce method here

PR #1845, opened 2023-01-27
js_style-to-object_v_01

DenysKolbasin:


src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const filterArr = sourceString.split(';').filter(value => value.length > 4);
```
ViktorHavryliuk, line 17: Why exactly greater than 4? If you want to filter out empty strings, better use filter(value => !!value.trim()), as it's more obvious

ViktorHavryliuk (changes requested):

src/convertToObject.js
DenysKolbasin: thanks

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const filterArr = sourceString.split(';').filter(value => !!value.trim());
+
+  return filterArr.reduce((prevElem, currentElem) => {
```
DarkMistyRoom, line 19:
```suggestion
  return filterArr.reduce((accumulator, currentElem) => {
```

DarkMistyRoom: it is better to understanding

DarkMistyRoom (approved):

PR #1846, opened 2023-01-28
done

ghost:


src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
+    item => item.includes(':') && item.length > 1
+  );
+
+  for (let i = 0; i < str.length; i++) {
+    const [key, value] = str[i].split(':');
+
+    result[key.trim()] = value.trim();
+  }
+
+  return result;
```
Volodymyr-Melnyk-Fron10der, line 24: There is faster way to do it
```suggestion
  return Object.fromEntries(str.map((line) => line.split(':')))
```

But if you split your items in reduce above - there will be only 
`return Object.fromEntries(str)`

src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const str = sourceString.split(';').filter(
+    item => item.includes(':') && item.length > 1
+  );
```
Volodymyr-Melnyk-Fron10der, line 16: Try to use `.reduce` to filter items + split (`split(':')`) them as you are doing bellow right here

src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const str = sourceString.split(';').filter(
```
Volodymyr-Melnyk-Fron10der, line 14:
```suggestion
  const styleLines/styles/properties = sourceString.split(';').filter(
```

Volodymyr-Melnyk-Fron10der (changes requested):
Great job! I suggest to make your code absolutely perfect:

ViktorHavryliuk (approved):

PR #1848, opened 2023-01-31
[SOLUTION]

grabynsky:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
```
Viktor-Kost, line 17: no needs to make this additional variable. Just return sourceString and after filter method chain reduce one

Viktor-Kost (changes requested):

Bohdan-Maksymiuk (approved):

PR #1850, opened 2023-01-31
Develop

vanyachyzh:


src/convertToObject.js
```diff
@@ -3,16 +3,40 @@
...
+    const key = item[0];
+    const value = item[1];
+
+    acc[key] = value;
+
+    return acc;
+  }, {});
+
+  // if we have some empty key we need to delete  it
+  delete object[''];
```
VitaliyBondarenko1982, line 37: Try to do it in reduce method. Or better make just one iteration and made everything in reduce instead map and reduce.

VitaliyBondarenko1982 (changes requested):

etojeDenys (approved):
LGTM

VitaliyBondarenko1982 (approved):
Good performance improvement!
You do not need to re-request review again, because this task is already approved 🙂

PR #1851, opened 2023-01-31
Solutions

OlyaNikora:


src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+ * (see an example in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString.split(';').map(style => style.split(':')
+    .map(property => property.trim())).filter(el => !''.includes(el));
```
etojeDenys, line 18: when we use chaining it is considered good practice to call the new method on a new line


``` 
   .map(style => style
            .split(':')
            .map(property => property.trim()))
   .filter(el => el.length > 1);
```

etojeDenys, line 18: and we can check here if array length bigger than one ( `.filter(el => el.length > 1);`) for better readability

etojeDenys (changes requested):

Bohdan-Maksymiuk (approved):

PR #1853, opened 2023-01-31
the task was solved

batiukova-nataliia:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stringCleared = sourceString.replace(/ /g, '').split('\n').join('');
+  const finalString = stringCleared
+    .replaceAll('1pxsolid#e8e8e8', '1px solid #e8e8e8')
+    .replaceAll('left!important', 'left !important')
+    .replaceAll('all.2sease-in-out', 'all .2s ease-in-out');
```
etojeDenys, line 22: the tests throw an error because they don't understand what replaceAll is (since this method was added very recently)

so let's replace it with a plain `replace` using the g (global) flag.

example: 
```
 .replace(/1pxsolid#e8e8e8/g, '1px solid #e8e8e8')
```

etojeDenys (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stringCleared = sourceString.replace(/ /g, '').split('\n').join('');
+  const finalString = stringCleared
+    .replace(/1pxsolid#e8e8e8/g, '1px solid #e8e8e8')
+    .replace(/left!important/g, 'left !important')
+    .replace(/all.2sease-in-out/g, 'all .2s ease-in-out');
```
Bohdan-Maksymiuk, line 22: you shouldn't rely on initial data in you logic
This is hardcode  - need to remove it

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+
+  const array = finalString.split(';');
```
Bohdan-Maksymiuk, line 24: use descriptive names, not array
name should describe what variable contains

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+
+  const filtered = array.filter(element => element !== '');
+  const object = filtered.reduce(function(target, key) {
```
Bohdan-Maksymiuk, line 27: same about names and use arrow functions as callbacks

Bohdan-Maksymiuk (changes requested):
you used complicated and hardcoded logic sometimes
read about `trim()` method and clean your code

etojeDenys (approved):
looks awesome

PR #1855, opened 2023-02-02
the function was created

YevheniiHura:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.split(';');
+
+  const obj = arr.reduce((accumulator, element) => {
+    const separator = element.indexOf(':');
+    const key = element.slice(0, separator).trim();
+    const value = element.slice(separator + 1).trim();
```
OlEzhA0, line 22: you can do like
```
const [key, value] = element.split(':').map(text => text.trim())
```

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+    accumulator[key] = value;
+
+    return accumulator;
+  }, {});
+
+  delete obj[''];
```
OlEzhA0, line 29: why do we need it? you can check your key before adding it to accumulator

OlEzhA0 (changes requested):
hey! please check my comments

src/convertToObject.js
YevheniiHura: done

src/convertToObject.js
YevheniiHura: done

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.split(';');
+
+  const obj = arr.reduce((accumulator, element) => {
+    // const separator = element.indexOf(':');
```
mykhalenych, line 20: delete comments

mykhalenych (changes requested):

src/convertToObject.js
YevheniiHura: done

PR #1856, opened 2023-02-02
● Solution

d-hubych:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+
+  splitedSource = splitedSource.filter(str => str.includes(':'));
+
+  splitedSource.forEach(style => {
+    const splitedStyle = style.split(':');
+    const key = `${splitedStyle[0].trim()}`;
+    const value = `${splitedStyle[1].trim()}`;
+
+    styles[key] = value;
+  });
```
IvanFesenko, line 28: Try to do it with reduce method

IvanFesenko (approved):
For practice try reduce

PR #1863, opened 2023-02-07
solution

YevaDrach:


tetiana-ivanchenko (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const obj = {};
+  const arrWithStyles = sourceString.split(';');
```
OleziO, line 21: It is better to name variables of these types - nameArr (in your case - stylesArr), this is considered a good practice.

OleziO (approved):

PR #1864, opened 2023-02-07
solution

Roman-Kulchytskyi:


src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
+  const newArr = sourceString.split(';')
+    .map(a => a.trim())
+    .filter(a => a.length > 1)
+    .map(a => a.split(':'));
+
+  return newArr.reduce((object, currentValue) => {
+    object[currentValue[0].trim()] = currentValue[1].trim();
+
+    return object;
+  }, {});
```
VitaliyBondarenko1982, line 24: Instead of map, filter, and again map try to do everything in reduce method just in one iteration.

VitaliyBondarenko1982 (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,24 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const newArr = sourceString.split(';');
+
+  return newArr.reduce((object, value) => {
+    if (value.split(':')[0].trim().length > 1) {
+      object[value.split(':')[0].trim()] = value.split(':')[1].trim();
+    }
```
```suggestion
  return newArr.reduce((object, value) => {
    const [key, value] = value.split(':');
    
    if (key.trim().length > 1) {
      object[key.trim()] = value.trim();
    }
```

tetiana-ivanchenko (changes requested):

Viktor-Kost (approved):

PR #1865, opened 2023-02-07
Solution

nadia-hope:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssStyles = {};
+
+  sourceString.split(';')
+    .map(eachString => eachString.trim().split(':'))
+    .map(eachArray => eachArray.map(eachString => eachString.trim()))
```
witflash, line 21: Adding this logic to the first `map` method is better. So you can avoid using the additional `map` method.

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+    .filter(eachArray => eachArray.length === 2)
+    .map(eachArray => {
```
witflash, line 23: It's really not a `map` but `forEach` method. Because it doesn't return a new array but does side effects only.

witflash (changes requested):
Your code is workable but you can improve it.
Pay attention to my comments, please.

polosanya (approved):
Good job!

PR #1866, opened 2023-02-08
develop

nazar-medushevskyi:


src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
+  const result = {};
+  const str = sourceString.split(';').filter(
+    item => item.includes(':') && item.length > 1
+  );
+
+  for (let i = 0; i < str.length; i++) {
+    const [key, value] = str[i].split(':');
+
+    result[key.trim()] = value.trim();
+  }
```
SanyaBratashchuk, line 22: Let's practice array iteration methods

SanyaBratashchuk (changes requested):

src/convertToObject.js
nazar-medushevskyi: go

PR #1867, opened 2023-02-08
Tests were completed successfully

read-my-code:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+  let tempArr = [];
+  const resultObj = {};
+
+  arrayOfProperties.forEach(item => {
+    tempArr = item.split(':');
+
+    if (tempArr.length === 2) {
+      resultObj[tempArr[0].trim()] = tempArr[1].trim();
+    }
+  });
```
Volodymyr-Melnyk-Fron10der, line 27: `tempArr ` array looks like not vital here, you can decompose your values right away like:

```suggestion
  const resultObj = {};

  arrayOfProperties.forEach(item => {
    const [key, value] = item.split(':');

    if (key.length && value.length) {
      resultObj[key] = value.trim();
    }
  });
```

Volodymyr-Melnyk-Fron10der (changes requested):
Great, but there is one comment

slavanosatskyi (approved):
Nice work, as an additional home task try to solve it with reduce function instead of forEach

PR #1869, opened 2023-02-08
Converter is done

oleksandr-plias:


src/convertToObject.js
```diff
@@ -1,18 +1,18 @@
...
+  const resultObj = {};
+  const createdArray = sourceString.split(';');
+
+  createdArray.map((element) => {
+    const [key, value] = element.split(':');
+
+    if (key && value) {
+      resultObj[key.trim()] = value.trim();
+    }
+  });
```
Anastasiia-Svintsova, line 13: The map() method should return some value. It doesn't fit here. Rewrite it by using forEach, please

Anastasiia-Svintsova (changes requested):
Good! Fix one comment, please

src/convertToObject.js
```diff
@@ -1,18 +1,18 @@
...
-}
+  const resultObj = {};
+  const createdArray = sourceString.split(';');
+
+  createdArray.forEach((element) => {
+    const [key, value] = element.split(':');
+
+    if (key && value) {
+      resultObj[key.trim()] = value.trim();
+    }
```
Volodymyr-Melnyk-Fron10der, line 12: This `if` is useless because string with spaces will be like truth value 
image
So you should trim it as well to check are there any symbols or not
```suggestion
    const trimmedKey = key.trim();
    const trimmedValue = value.trim();
    
    if (trimmedKey && trimmedValue ) {
      resultObj[trimmedKey] = trimmedValue ;
    }
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! I left one comment and as optional recommendation: 
This task is a great opportunity to use `reduce` method instead of using `forEach` loop and `resultObj`. If you want try to deal with it

ViktorHavryliuk (approved):

PR #1873, opened 2023-02-10
Solution

vlad-ptashnichenko:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const str = sourceString.trim();
+  const property = str.split(';');
+  const resultObject = {};
+
+  property.forEach(item => {
+    const element = item.split(':');
+    const key = element[0] ? element[0].trim() : element[0];
+    const value = element[1] ? element[1].trim() : element[1];
```
```suggestion
    const [key, value] = item.split(':');
```

Use destructuring and you can move trim to 27 line

nazarmatsevych (changes requested):

etojeDenys (approved):
LGTM

PR #1874, opened 2023-02-10
solution convert obj

Spec0s:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const splitSourceString = sourceString
+    .split(';')
+    .map(a => a.split(':'));
```
Volodymyr-Melnyk-Fron10der, line 20: Create meaningful names
```suggestion
    .map(keyValuePair => keyValuePair.split(':'));
```

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+
+  const sourceStringToObj = {};
+
+  for (const style of splitSourceString) {
```
Volodymyr-Melnyk-Fron10der, line 24: Use destructurization as much as possible
```suggestion
  for (const [propertyName, styleValue] of splitSourceString) {
```

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    if (style[0].toUpperCase() !== style[0].toLowerCase()) {
```
Volodymyr-Melnyk-Fron10der, line 25: Doesn`t code work without this condition? I struggle to find this condition useful here

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! Great work! Check my comments bellow pls

src/convertToObject.js
Spec0s: i am still very bad at naming thing but feel like improved a bit

src/convertToObject.js
Spec0s: oh man i was thinking about this  !  and forgot...

src/convertToObject.js
Spec0s: when i destructurized had to change as this was not working i didn't fully figure out why, but i guess had to do something with undefined at places i think

Volodymyr-Melnyk-Fron10der (changes requested):
Please make sure you have pushed your changes, as I don't see any changes here

Spec0s:
ok sorry maybe forgot

Volodymyr-Melnyk-Fron10der (approved):
Looks good✅

PR #1876, opened 2023-02-10
task done

tecHamster:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.split(';').map(item => item.split(':').map(elem => {
+    return elem.trim();
```
OleziO, line 18: If you have one string return - better to do it without brackets and word return, itst more compact, like:

`(elem) => elem.trim();`

OleziO (approved):

src/convertToObject.js
tecHamster: Line length will be over 80 symbols after this change. It is 80 right now

PR #1877, opened 2023-02-11
solution

marynatovtyn:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+
+  stylesArray.map(style => {
+    const [key, value] = style.split(':');
+
+    if (key && value) {
+      stylesObject[key.trim()] = value.trim();
+    }
+  });
+
+  return stylesObject;
```
SanyaBratashchuk, line 28: Do not use map when you don't need its result. Use forEach or reduce instead

SanyaBratashchuk (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+  const stylesArray = sourceString.split(';');
+
+  stylesArray.forEach(style => {
+    const [key, value] = style.split(':');
+
+    if (key && value) {
```
Volodymyr-Melnyk-Fron10der, line 23: the value or key like '   '  (spaces) will be received as true, so trim it as well

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, I left one comment and just optional recommendation: try to use a `reduce ` method here, that's exactly the case to use it

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    const separatorIndex = item.indexOf(':');
+    const key = item.slice(0, separatorIndex).trim();
+    const value = item.slice(separatorIndex + 1).trim();
+
+    prev[key] = value;
+
+    return prev;
+  }, {});
+
+  delete stylesObject[''];
```
DarkMistyRoom, line 29: what if there are some keys with empty values? better check it before adding them to the object

DarkMistyRoom (changes requested):

Viktor-Kost (approved):
Nice!)

PR #1880, opened 2023-02-12
Develop

RuslanLiapin:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+ (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  // eslint-disable-next-line max-len, no-unused-vars
+  let object = {};
```
IvanFesenko, line 18: Useless variable. You can return result without declaring variable

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+  };
+  const arrayOfProperties = sourceString.split(';').map(prop => {
+    return prop.trim();
+  }).filter(prop => {
+    return prop.length !== 0;
+  }).map(prop => prop.split(':'));
+
+  object = arrayOfProperties.reduce(callback, object);
+
+  return object;
```
IvanFesenko, line 33:
```suggestion
  return arrayOfProperties.reduce(callback, object);
```

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+    return {
+      ...prev,
+      [property[0].trim()]: property[1].trim(),
+    };
+  };
+  const arrayOfProperties = sourceString.split(';').map(prop => {
+    return prop.trim();
+  }).filter(prop => {
+    return prop.length !== 0;
+  }).map(prop => prop.split(':'));
```
IvanFesenko, line 29: Lets make some refactor
```suggestion
  const arrayOfProperties = sourceString.split(';')
    .map(prop =>  prop.trim())
    .filter(prop => prop.length !== 0)
    .map(prop => prop.split(':'));
```

IvanFesenko (changes requested):
Good job. Please check my comments

mykhalenych (approved):

PR #1883, opened 2023-02-13
Solution

taras-nakonechnyi:


src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const attrArray = sourceString.split(';').map(str => str.trim());
+  const result = attrArray.filter(el => el !== '').reduce((target, key) => {
+    const item = key.split(':');
+
+    target[item[0].trim()] = item[1].trim();
```
tetiana-ivanchenko, line 21:
```suggestion
    const [newKey, newValue] = key.split(':');

    target[newKey.trim()] = newValue.trim();
```

tetiana-ivanchenko (changes requested):

src/convertToObject.js
taras-nakonechnyi: I didn't know it was possible😁😉👍👍👍

sTorba24 (approved):
Approved!

PR #1884, opened 2023-02-14
solution

Alessandranitti15:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const formattedArray
+    = sourceString.split(';')
+      .map(line => line.trim())
+      .filter(line => line.length > 1);
```
Volodymyr-Melnyk-Fron10der, line 21: You can avoid extra map iteration by merging map & filter iterations like:
```suggestion
  const formattedArray = sourceString.split(';').filter(line => line.trim().length);
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! Almost perfect✅ Just a little change I would ask you about

Volodymyr-Melnyk-Fron10der (approved):
Good job👍

PR #1885, opened 2023-02-14
Task done

sergik94:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const styleAll = sourceString.split(';');
+  const styleTrue = styleAll.filter(style => {
+    return style.split(':')[0].trim().length > 0;
+  });
```
slavanosatskyi, line 21: Looks like a bit complicated logic here. You can try the approach below. It probably should do the trick as empty string is a falsy expression
```suggestion
  const validStyles = sourceString.split(';').map(style => style.trim()).filter(style => style);
```

slavanosatskyi (changes requested):

src/convertToObject.js
sergik94: Yes, I agree, your variant is a bit better

etojeDenys (approved):
LGTM

you can practice and rewrite it without using filter and only with one map (just at the end add `delete result[""];`)

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const styleTrue = sourceString
+    .split(';').map(style => style.trim()).filter(style => style);
+
+  styleTrue.map(style => {
```
etojeDenys, line 21: and here you can use forEach, because map is used to transform each element of an array, while forEach is used to run a function on each element without changing the array.

sergik94:
@etojeDenys , I've optimized code. Did you mean it when say "rewrite it without using filter and only with one map"?

mykhalenych (approved):

PR #1886, opened 2023-02-15
Initial commit

RomanHrytsiuk:


src/convertToObject.js
```diff
@@ -3,16 +3,60 @@
...
+
+//   const result = [].concat(...resultArray);
+
+//   for (let i = 0; i < result.length; i += 2) {
+//     if (result[i] === '') {
+//       continue;
+//     }
+//     obj[result[i]] = result[i + 1];
+//   }
+
```
OleziO, line 39: Delete all comments in the code.

src/convertToObject.js
```diff
@@ -3,16 +3,60 @@
...
+//   return obj;
+// }
+
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  const array = sourceString.split(';').map(item => item.split(':'));
+
+  const result = [].concat(...array);
```
OleziO, line 47: Better is to do this with the spread operator.
`const result = [...array];`

[More info](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax)

OleziO (changes requested):

src/convertToObject.js
RomanHrytsiuk: The code doesn't work when I do so

OleziO: image

OleziO: The same result

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+ * (see an example in [stylesString.js](./stylesString.js))
  * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
```
OleziO, line 15: Delete comment

OleziO (approved):

PR #1899, opened 2023-02-20
solution

toljan1:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  const stylesObj = {};
+
+  for (const part of arr) {
+    if (part[0] && part[1]) {
+      stylesObj[part[0].trim()] = part[1].trim();
+    }
+  }
+
+  return stylesObj;
```
VitaliyBondarenko1982, line 29: Good job, but suggesting to try make just one iteration through stylesArr with reduce method and make inside everything)

VitaliyBondarenko1982 (changes requested):

IvanFesenko (changes requested):
Refactor as was discussed in slack

etojeDenys (approved):
looks good

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArr = sourceString.split(';')
```
etojeDenys, line 17: but in the end, it`s the object, not the array

PR #1900, opened 2023-02-20
Develop

kateryna-babina:


timoxa2t (approved):
Good work

src/convertToObject.js
```diff
@@ -1,18 +1,18 @@
...
+  const styleDeclarations = sourceString.split(';');
+  const stylesObject = {};
+
+  styleDeclarations
+    .filter((style) => style.trim())
+    .forEach(style => {
+      const [key, value] = style.split(':');
+
+      stylesObject[key.trim()] = value.trim();
+    });
```
OleksiiBlezniuk, line 13: `forEach` - is same as loop, let's learn to use `reduce` method 😉

OleksiiBlezniuk (changes requested):

src/convertToObject.js
```diff
@@ -1,18 +1,18 @@
...
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleDeclarations = sourceString.split(';');
+  const stylesObject = styleDeclarations
+    .filter((style) => style.trim())
```
sergii-nosachenko, line 6: You don't need to have `styleDeclarations` and `stylesObject` variables.
Chain methods and just return value immediately.

`return arr.split().filter().reduce()`...

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1902, opened 2023-02-20
Solution

timoxa2t:


rashyd-hasratov (approved):
You are codewars star :1st_place_medal:

src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const oobjectified = sourceString
+    .split(';')
+    .map(prop => prop
+      .trim()
+      .split(':')
+      .map(item => item.trim()))
+    .filter(prop => prop.length > 1)
```
OleksiiBlezniuk, line 15:
```suggestion
    .filter(prop => prop)
```

OleksiiBlezniuk: You can just check element, empty string will return `false` when it's converted to boolean

OleksiiBlezniuk (approved):
Well done! But pay attention to comment

PR #1904, opened 2023-02-21
convertToObject is done

rashyd-hasratov:


milenadiamonds1 (approved):
good job!

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  // let's divide the whole rule into group of properties
```
OleksiiBlezniuk, line 17: Good code - is a code that is not requires comments, so let's remove annotations and write good code instead 😉

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+  // 1) let's divide each property into property-name and property-value
+  // 2) we don't save blank elements (they are not correct properties)
+  const propertiesSplitted = ruleSplitted
+    .map(property => property.split(':'))
+    .filter(property => (property.length === 2));
+
+  // let's remove side spaces (we need property name and value only)
+  const propertiesSplittedFormatted = propertiesSplitted.map(property => (
+    property.map(propertyPart => propertyPart.trim())
+  ));
```
OleksiiBlezniuk, line 29: But why should we save it to constants, if we can call methods in chain?

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+
+  const styles = {};
+
+  // each property-name is a key and property-value is the key's value
+  propertiesSplittedFormatted.forEach(property => {
+    styles[property[0]] = property[1];
+  });
```
OleksiiBlezniuk, line 36: `forEach` is same as loop, let's learn how to use `reduce` method 😉

OleksiiBlezniuk (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesSplitted = sourceString
```
sergii-nosachenko, line 17:
```suggestion
  return sourceString
```
You can immediately return a result.

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    .split(';')
+    .map(property => property.split(':'))
+    .filter(property => (property.length === 2))
+    .map(property => (
+      property.map(propertyPart => propertyPart.trim())
+    ))
+    .reduce((objectifiedStyles, property) => {
```
sergii-nosachenko, line 24:
```suggestion
    .reduce((objectifiedStyles, [key, value]) => {
```
Use destructurization.

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesSplitted = sourceString
+    .split(';')
+    .map(property => property.split(':'))
+    .filter(property => (property.length === 2))
+    .map(property => (
+      property.map(propertyPart => propertyPart.trim())
+    ))
```
sergii-nosachenko, line 23: This step can be omitted. Just use trim inside reduce.

sergii-nosachenko (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .map(property => property.split(':'))
+    .filter(property => (property.length === 2))
```
```suggestion
    .filter(property => property.length === 2)
```

sergii-nosachenko (approved):
Good work 👍

PR #1905, opened 2023-02-21
solution

milenadiamonds1:


vladyslav-mosor (approved):

vladyslav-mosor:
Good implementation of the task but I think it is better to rename your object variable.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splitted = sourceString.split('\n');
+  const obj = {};
+
+  for (const string of splitted) {
```
sergii-nosachenko, line 20: Checklist #5

sergii-nosachenko (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+
+  splittedString.forEach(string => {
+    const [prop, value] = string.split(':');
+
+    if (prop && value) {
+      objWithStyles[prop.trim()] = value.replace(/;/g, '').trim();
+    }
+  });
+
+  return objWithStyles;
```
sergii-nosachenko, line 28: We are learning how to use `map`, `filter`, and `reduce` methods. Let's use them instead of `forEach`. Also, don't forget about chaining. Checklist #4

sergii-nosachenko (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+ * and values are the values of related CSS
+ * properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splittedString = sourceString.split('\n');
```
sergii-nosachenko, line 17: Wrong separator. This string is also valid: `position: absolute; top: 10px;`. But your code will not work.

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+  const objWithStyles = {};
```
sergii-nosachenko, line 18: Useless variable

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+
+  splittedString.map(string => {
```
sergii-nosachenko, line 20: Instead of `map` use reduce and chain it to a splitted string array without creating a new variable. Inside `reduce` use the object as an accumulator.

sergii-nosachenko (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const objWithStyles = sourceString
+    .split(';')
+    .reduce((acc, string) => {
+      const [prop, value] = string.split(':');
+
+      if (prop && value) {
+        acc[prop.trim()] = value
+          .replace(/;/g, '')
```
sergii-nosachenko, line 24:
```suggestion
```
You don't have `;` in string.

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+ * and values are the values of related CSS
+ * properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objWithStyles = sourceString
```
sergii-nosachenko, line 17:
```suggestion
  return sourceString
```

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    .split(';')
+    .reduce((acc, string) => {
```
sergii-nosachenko, line 19: What is `acc`? What is `string`?

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1907, opened 2023-02-21
solution for styleToObject

vladyslav-mosor:


Yekatery (approved):
Well done!)

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArray = sourceString
+    .split(';\n')
+    .map(element => element.split(':'));
+
+  const stylesObject = {};
+
+  sourceArray.forEach(property => {
```
sergii-nosachenko, line 10: The `forEach` method is the same as using a loop. Use reduce instead.
Also, chain all methods. You don't need to have `sourceArray` and `stylesObject` constants at all.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good job 👍

PR #1908, opened 2023-02-22
Style to object

Anton-Tretiak:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+  const styles = sourceString
+    .split(';')
+    .map(style => style.split(':'));
+
+  styles.forEach(prop => {
```
sergii-nosachenko, line 22: The `forEach` method is the same as using a loop. Use reduce instead.
Also, chain all methods. You don't need to have `stylesObject` and `styles` constants at all.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    if (prop.length === 2) {
+      stylesObject[prop[0].trim()] = prop[1].trim();
```
sergii-nosachenko, line 24: Destructurize argument instead of having `prop[0], prop[1]`...

sergii-nosachenko (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .reduce((stylesObject, el) => {
```
sergii-nosachenko, line 19:
```suggestion
    .reduce((stylesObject, property) => {
```

sergii-nosachenko (approved):
Good work 👍
Commented

PR #1909, opened 2023-02-22
Solution

nazarovoleksiy:


gretoq (approved):
Good job, but you can use 'reduce' and write it after 'return' without defining another variables outside the iteration arrays method.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedStyles = {};
+
+  sourceString
+    .split(';')
+    .filter(item => item.trim())
+    .map(style => {
```
sergii-nosachenko, line 22: Use reduce to accumulate an object with props.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedStyles = {};
+
+  sourceString
```
sergii-nosachenko, line 19: You don't need a separate variable for object. You can immediately return result from the last method (reduce).

sergii-nosachenko (changes requested):

sergii-nosachenko:
@gretoq Your comment is the reason to reject, not to approve.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .filter(item => item.trim())
+    .reduce((styles, style) => {
```
Andr1yK, line 20: u can filter items in reduce)

Andr1yK (changes requested):

Andr1yK (approved):

PR #1910, opened 2023-02-22
solution

Yekatery:


sergey-lipikhin (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resultObject = {};
+
+  sourceString.split(';')
```
sergii-nosachenko, line 19:
```suggestion
  return sourceString
    .split(';')
```
You don't need a variable. It can be done inside the last reduce call.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1912, opened 2023-02-22
Develop js_style-to-object

sergey-lipikhin:


NataliaSavenets (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const properties = sourceString.split(';');
+  const propertyValuePair = properties
+    .map(property => property.split(':'))
+    .filter(pair => pair.length === 2)
+    .map(pair => (
+      { [pair[0].trim()]: pair[1].trim() }
+    ));
```
sergii-nosachenko, line 23: Use `reduce` instead. Accumulate props to an empty object.

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const properties = sourceString.split(';');
+  const propertyValuePair = properties
```
sergii-nosachenko, line 18:
```suggestion
  return sourceString
    .split(';')
```
Chain methods. Also, you don't need to have those variables. Just return the accumulated object.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 🔥

PR #1913, opened 2023-02-22
Solution for Convert to Object

ctk12:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const obj = {};
+  const res = sourceString.split(';')
```
polosanya, line 19: image

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    .map(data => data.split(':').map(val => val.trim()));
```
polosanya, line 20:
```suggestion
  const res = sourceString
    .split(';')
    .map(data => data
       .split(':')
       .map(val => val.trim())
     );
```
image

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+
+  res.map((ar) => {
+    if (ar.length !== 1) {
+      obj[ar[0]] = ar[1];
```
polosanya, line 24: Same problem with naming here

polosanya (changes requested):
Almost done!

qoa11a (approved):

PR #1920, opened 2023-02-22
Created a function which takes a string and returns styles CSS object

gretoq:


Vlados3611 (approved):
Good Job! ;)

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';')
+    .filter(style => style.trim().length > 0)
+    .map(style => style.trim())
+    .reduce((styles, style) => {
+      const nameOfStyle = style.split(':')[0].trim();
+      const valueOfStyle = style.split(':')[1].trim();
```
sergii-nosachenko, line 14: Why do you need to split the style twice? Better destructurize it at first call.

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
- *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';')
+    .filter(style => style.trim().length > 0)
+    .map(style => style.trim())
```
sergii-nosachenko, line 11: Useless

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';')
```
sergii-nosachenko, line 9:
```suggestion
  return sourceString
    .split(';')
```

sergii-nosachenko (changes requested):

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
+    .filter(style => style.trim().length > 0)
```
OleksiiBlezniuk, line 10:
```suggestion
    .filter(style => style.trim())
```

OleksiiBlezniuk: If this line is empty - it will return false too, so it's enough to just leave string trim result

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
+    .reduce((styles, style) => {
+      const [nameOfStyle, valueOfStyle] = style
+        .split(':')
+        .map(stringPart => stringPart.trim());
+
+      styles[nameOfStyle] = valueOfStyle;
+
+      return styles;
```
OleksiiBlezniuk, line 18:
```suggestion
      styles[nameOfStyle] = valueOfStyle;

      return {
        ...styles,
        [nameOfStyle]: valueOfStyle
      };
```

OleksiiBlezniuk: Not critical in this case, but still it's a bad idea to mutate incoming function param

OleksiiBlezniuk (approved):

PR #1956, opened 2023-02-25
'Solution'

nazarkharkevych:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+  });
+
+  const filteredStyles = trimedStyles.filter(property => property[0] !== '');
+
+  const result = filteredStyles.reduce((acc, property) => {
+    return {
+      ...acc,
+      [property[0]]: property[1],
+    };
+  }, {});
```
Volodymyr-Melnyk-Fron10der, line 32: you can save 2 useless loops by moving these operations inside your reducer. And now you have only 1 iteration insead of 3

```suggestion
  const result = filteredStyles.reduce((acc, [property, value]) => {
    const trimmedProperty = property.trim();
    const trimmedValue = value.trim();
    
    if (trimmedProperty.length && trimmedValue .length) {
      acc[trimmedProperty] = trimmedValue ;
    }
    
    return acc;
  }, {});
```

Volodymyr-Melnyk-Fron10der (changes requested):
Great job! I suggest to raise the performance of your program a bit

OleziO (approved):

PR #1963, opened 2023-02-27
Solution

Zavalishyn-Oleksandr:


src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
+    const arr = property.split(':');
+    let nameOfProperty = arr[0].split(' ').join();
+
+    do {
+      nameOfProperty = nameOfProperty.replace(',', '');
+    } while (nameOfProperty.includes(','));
+
+    do {
+      nameOfProperty = nameOfProperty.replace('\n', '');
+    } while (nameOfProperty.includes('\n'));
```
DarkMistyRoom, line 30: this logic is too complicated. you can just trim key and value and check if both of them exist after

DarkMistyRoom (changes requested):

Zavalishyn-Oleksandr:
Yeah this seems much less complicated, thanks)

SerhiiKirik (approved):
Good job 🔥

PR #1967, opened 2023-02-28
solution for stylesObject

kuzmi491:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
-  // write your code here
+  const arrayOfStyles = sourceString
+    .split(';')
+    .map(styles => styles.split(':'))
+    .filter(style => style.length === 2);
+
+  const objectOfStyles = arrayOfStyles
+    .reduce((acc, [styles, value]) => ({
+      ...acc, [styles.trim()]: value.trim(),
+    }), {});
```
illiarerun, line 25: fix formatting here
```suggestion
    .reduce((acc, [styles, value]) => ({
      ...acc,
      [styles.trim()]: value.trim(),
    }), {});
```

illiarerun (changes requested):

src/convertToObject.js
kuzmi491: done

illiarerun (approved):

PR #1970, opened 2023-03-02
Solution

kostiantin-hudzenko:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  const arrWithStyles = sourceString.split(';');
+
+  arrWithStyles.map((element) => {
+    const [key, value] = element.split(':');
```
IvanFesenko, line 21: it's better use forEach here, 
```suggestion
  arrWithStyles.forEach((element) => {
    const [key, value] = element.split(':');
```

IvanFesenko (approved):
Well done

PR #1973, opened 2023-03-03
style to object first try

RufatRM:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const arrayOfStrings = sourceString.split(';')
+    .filter(item => item && item !== '\n')
+    .map(item => item.replace('\n', ''));
```
Volodymyr-Melnyk-Fron10der, line 20: you can simplify it
```suggestion
    .filter(item => item && !item.includes('\n'))
```

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+
+  const stringToObject = arrayOfStrings.reduce((prev, element) => {
+    return {
+      ...prev,
+      [element.slice(0, element.indexOf(':')).trim()]:
+      (element.slice(element.indexOf(':') + 1).trim()),
+    };
+  }, {});
```
Volodymyr-Melnyk-Fron10der, line 28: make it more clear
```suggestion
  const stringToObject = arrayOfStrings.reduce((acc, element) => {
    const [property, value] = element.split(':');
    
    if (property.trim()) {
      acc[property.trim()] = value.trim(),
    }
    
    return acc;
  }, {});
```

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+
+  const key = '';
+
+  delete (stringToObject[key]);
+
```
Volodymyr-Melnyk-Fron10der, line 33: I don't see a reason to do it actually

Volodymyr-Melnyk-Fron10der (changes requested):
almost there

polosanya (approved):
Good job!

PR #1974, opened 2023-03-04
solution :)

katerynashylina:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+
+  splittedStrings.forEach(part => {
+    const [style, value] = part.split(':');
+
+    if (style && value) {
+      convertedObject[style.trim()] = value.trim();
+    }
+  });
+
+  return convertedObject;
```
SanyaBratashchuk, line 28: Try to use reduce to create resulting object

SanyaBratashchuk (approved):

PR #1977, opened 2023-03-06
Solution

fisherok3452:


etojeDenys (changes requested):
looks good, but you can try to rewrite it using `reduce()` it will improve readability

PR #1984, opened 2023-03-10
Solved

silcki:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+    .map(x => x.trim().replace('/"/', ''))
+    .filter(x => x.length > 0)
+    .map(x =>
+      x.split(':').map(y => y.trim())
+    )
+    .reduce(
+      (accumulator, [key, value]) => {
+        return {
+          ...accumulator,
+          [key]: value,
```
etojeDenys, line 29: can omit here

etojeDenys (approved):
looks awesome 🤩

src/convertToObject.js
silcki: What exactly can I omit here?

PR #1985, opened 2023-03-12
solution

zvir91:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splitBySem = sourceString.split(';');
+  const trimed = splitBySem.map(element => element.trim());
+  const splitByColon = trimed.map(element => element.split(':'));
```
Volodymyr-Melnyk-Fron10der, line 20: save one iteration
```suggestion
  const trimed = splitBySem.map(element => element.trim().split(':'));
```

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+
+  const trimAgain = splitByColon.map(element => element.map(el => el.trim()));
```
Volodymyr-Melnyk-Fron10der, line 22: and even two
`const trimed = splitBySem.map(element => element.trim().split(':').map(el => el.trim()));`

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+
+  const converted = {};
+
+  trimAgain.filter(el => el[0] !== '').forEach(el => {
+    converted[el[0]] = el[1];
+  });
```
Volodymyr-Melnyk-Fron10der, line 28: make it more clear
```suggestion
   trimAgain.forEach(([key, value]) => {
     if (key){
       converted[key] = value;
     }
  });
```

Volodymyr-Melnyk-Fron10der (changes requested):
Almost there, plz check my suggestions

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splitBySem = sourceString.split(';');
+  const trimed = splitBySem.map(
+    element => element.trim().split(':').map(el => el.trim()));
```
polosanya, line 20:
```suggestion
  const trimed = splitBySem.map(
    element => element
      .trim()
      .split(':')
      .map(el => el.trim())
  );
```
image

polosanya (changes requested):
Almost perfect!

polosanya (approved):

PR #1986, opened 2023-03-12
Comleted task

metja:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+  const result = {};
+  let arrCss = sourceString.split(';\n');
+
+  arrCss = arrCss.filter(properties => properties.includes(':'));
+
+  arrCss.forEach(properties => {
+    const [property, value] = properties.split(':');
+
+    result[property.trim()] = value.trim();
+  });
```
DarkMistyRoom, line 26: very good. but you should know that we can also use reduce method here to collect all data in the object

DarkMistyRoom (approved):

PR #1990, opened 2023-03-16
created the convertToObject function

prokivskaaa:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayOfStrings = sourceString.split(';');
+  const object = {};
```
roshirin, line 18: Checklist #2 suggests to use more specific variable names avoiding 'object', 'array' etc

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  for (const el of arrayOfStrings) {
+    const currentKey = el.slice(0, el.indexOf(':')).trim().replace(/_/g, '-');
+    const currentValue = el.slice(el.indexOf(':') + 1).trim();
```
roshirin, line 22: All chained methods called on the same level should have same indentation (Checklist #4)

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayOfStrings = sourceString.split(';');
+  const object = {};
+
+  for (const el of arrayOfStrings) {
```
roshirin, line 20: According to checklist #5 it would be better to use an iteration method here, although 'for' loop works fine in my opinion.

roshirin (approved):
Nice function, but there are some code style suggestions.

src/convertToObject.js
prokivskaaa: @roshirin Thank you! Done)

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayOfStrings = sourceString.split(';');
+  const objectWithStyles = {};
+
+  for (const el of arrayOfStrings) {
```
liliia-livandovskaya, line 20: checklist p.5  Avoid using loops in this task, use iteration methods instead.

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+    const currentKey = el
+      .slice(0, el.indexOf(':'))
+      .trim()
+      .replace(/_/g, '-');
+    const currentValue = el
```
liliia-livandovskaya, line 25:
```suggestion
      .replace(/_/g, '-');
      
    const currentValue = el
```
el is a bad variable name

prokivskaaa:
> Good job! Let's work on improvement

Thanks! Done)

liliia-livandovskaya (approved):
Great job!

PR #1996, opened 2023-03-20
Solution

igorkruz:


Esteban4769 (approved):
Nice work

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splitedString = sourceString
+    .split(';')
+    .map(el => el.split(':'))
```
liliia-livandovskaya, line 19: `el` is not a good name for the variable

liliia-livandovskaya (approved):

src/convertToObject.js
igorkruz: fixed )

PR #1998, opened 2023-03-21
made convertor from string to object

PavloReutskiy:


src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
+    .map(style => stylesNormalize(style))
+    .filter(element => element !== '');
+
+  const stylesObject = {};
+
+  stylesArray.forEach(property => {
+    const [value, key] = property.split(': ');
+
+    stylesObject[value] = key;
+  });
```
liliia-livandovskaya, line 35: you could also use `reduce` here

liliia-livandovskaya (approved):
Great job!

PavloReutskiy:
thx! try reduce for practice

PR #2004, opened 2023-03-22
solution

PashaMalyshkin:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const cssRulesObject = {};
+  const cssRulesArray = sourceString.split(';')
+    .map(cssItem => cssItem.slice(1).split(':'))
```
AnyaLyaLya, line 20: put first split on the next line 
image

AnyaLyaLya (approved):
GJ

src/convertToObject.js
liliia-livandovskaya: agree

liliia-livandovskaya (approved):
🔥

PR #2007, opened 2023-03-22
solution

Stambuliak:


src/convertToObject.js
```diff
@@ -12,7 +13,16 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const normalized = sourceString.split(':').map(x => x.trim()).join(': ');
+
+  normalized.split(';').map(x => x.trim()).filter(y => y !== '').forEach(z => {
+    const [key, value] = z.split(': ');
+
+    result[key] = value;
+  });
```
Kavon-Skela, line 23:
```suggestion
  const normalized = sourceString
    .split(':')
    .map(x => x.trim())
    .join(': ');

  normalized
    .split(';')
    .map(x => x.trim())
    .filter(y => y !== '')
    .forEach(z => {
      const [key, value] = z.split(': ');

    result[key] = value;
  });
```

Kavon-Skela (approved):
It is good overall. But I think. you need to work on your code style to make it more readable

src/convertToObject.js
```diff
@@ -12,7 +13,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
liliia-livandovskaya, line 16: checklist p.2 [CODE STYLE] - Avoid variable namings like array , object arr, obj, result. It will decrease code readability

src/convertToObject.js
```diff
@@ -12,7 +13,16 @@
...
+  const normalized = sourceString.split(':').map(x => x.trim()).join(': ');
+
+  normalized.split(';').map(x => x.trim()).filter(y => y !== '').forEach(z => {
+    const [key, value] = z.split(': ');
```
liliia-livandovskaya, line 20: checklist p.4  - All chained methods called on the same level should have same indentation. This makes code much more readable.

liliia-livandovskaya: Also, variable names such as `x, y, z` are bad

src/convertToObject.js
Stambuliak: +

src/convertToObject.js
Stambuliak: fixed

src/convertToObject.js
Stambuliak: fixed

liliia-livandovskaya (approved):

PR #2008, opened 2023-03-22
Solution

Kavon-Skela:


src/convertToObject.js
```diff
@@ -3,16 +3,57 @@
...
+//   .split(';')
+//   .filter(elem => elem && elem.length > 5)
+//   .map(elem => elem.trim())
+//   .map(elem => elem.split(':')
+//   .map(elem => elem.trim()))
+//   .reduce((acc, elem) => {
+//     acc[elem[0]] = elem[1];
+//     return acc;
+//   }, {})
+
```
VenomMB, line 58: For me that solution, more simply and clearly than first

VenomMB (approved):
Good job!!!!

src/convertToObject.js
```diff
@@ -3,16 +3,57 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const withoutSideSpaces = sourceString.trim();
+  const withoutNewLines = withoutSideSpaces
+    .split('\n')
+    .join(' ')
+    .split(';');
+
+  const withoutEmptyStrings = withoutNewLines
+    .filter(elem => elem && elem.length > 5)
```
liliia-livandovskaya, line 24: move `5` to the variable

liliia-livandovskaya (approved):
Great job! That is awesome that you add 2 solutions, both of them are good. But the first one is simpler to understand, so I would keep it.

PR #2010, opened 2023-03-22
Develop

vinnykalisa:


src/convertToObject.js
```diff
@@ -3,16 +3,40 @@
...
+
+  separateStyle.forEach(style => {
+    const formattedStyles = style.trim();
+    const key = formattedStyles
+      .slice(0, formattedStyles
+        .indexOf(':')).trim();
+    const value = formattedStyles
+      .slice(formattedStyles
+        .indexOf(':') + 1)
+      .trim();
```
liliia-livandovskaya, line 32:
```suggestion
    const formattedStyles = style.trim();
    
    const key = formattedStyles
      .slice(0, formattedStyles
      .indexOf(':'))
      .trim();
       
    const value = formattedStyles
      .slice(formattedStyles
      .indexOf(':') + 1)
      .trim();
```

vinnykalisa:
> Good job! Pay attention to the comments

what can I do if linter doesn't want me to set indentation as recommended?

PR #2013, opened 2023-03-23
Solution

NataliiaHen:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const stringArr = sourceString.split(';');
+  const data = {};
+
+  stringArr.forEach(elem => {
+    const source = elem.split(':');
```
nazarmatsevych, line 22:
```suggestion
    const [key, value]= elem.split(':');
```

you can use destructuring here

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    const key = source[0];
+    const value = source[1];
```
nazarmatsevych, line 24: And now you can remove this lines

nazarmatsevych (changes requested):

SerhiiKirik (approved):
LGTM 🔥

PR #2021, opened 2023-03-23
Develop

Nazarii98:


iHateAntichrist (approved):
everything is ok

src/convertToObject.js
```diff
@@ -1,18 +1,19 @@
...
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceObject = {};
+
+  sourceString
+    .split(';')
+    .map(element => element.trim())
```
liliia-livandovskaya, line 8: `element` is a bad name for the variable, it should have context

liliia-livandovskaya (approved):

PR #2029, opened 2023-03-23
task solved

M1k1ta:


Pozdnya (changes requested):
Pay attention to comments

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectStyles = {};
+
+  sourceString
+    .split(';')
+    .map(style => style.split(':').map(partStyle => partStyle.trim()))
```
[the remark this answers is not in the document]
M1k1ta: You will be right if it will be on the same level, but the second 'split' and 'map' there are in the first map so they have another level. If I don't right mentor will write about this.

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```
SerhiiKirik, line 19: In the future - Do not add extra files to PR
To add the file to a commit use git add src/.

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+    .map(style => (style[0] !== '')
+      ? (objectStyles[style[0]] = style[1])
+      : style);
```
SerhiiKirik, line 24:
```suggestion
    .map(style => {
      const [key, value] = style;
      
      if (key !== '') {
        objectStyles[key] = value;
      } 
```

SerhiiKirik (approved):
Great 🔥

PR #2031, opened 2023-03-23
solution

Evgenb000:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  let objectStyles = sourceString.split(';');
+
+  objectStyles = objectStyles
+    .map(word => word.split(':'))
+    .reduce((obj, item) => {
+      if (item.length === 2) {
```
maksym-didyk, line 23: maybe, you can do this solution an another way (without if)

maksym-didyk (approved):
Good job!
Maybe, you can do your solution without if?

src/convertToObject.js
liliia-livandovskaya: agree, it would be better to avoid such hardcoding

liliia-livandovskaya (approved):

PR #2032, opened 2023-03-23
Solution

vksazonov:


Anna-Kuzmenko (approved):
GJ!

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString.split(';');
+
+  const stylesPairs = stylesArray.map(style => style.split(':')
+    .map(word => word.trim())
+    .filter(space => space !== ''))
+    .filter(isEmpty => isEmpty.length > 0);
```
jstmpelowycz, line 23:
```suggestion
  const stylesPairs = stylesArray
    .map(style => (
      style
        .split(':')
        .map(word => word.trim())
    )
    .filter(space => space !== ''))
    .filter(isEmpty => isEmpty.length > 0);
```

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  const stylesObject = stylesPairs.reduce((obj, [key, value]) => {
+    obj[key] = value;
+
+    return obj;
+  }, {});
```
jstmpelowycz, line 29: or like this, just a suggestion
```suggestion
  const stylesObject = stylesPairs.reduce((obj, [key, value]) => ({
    ...obj,
    [key]: value,
  }), {});
```

jstmpelowycz (approved):

PR #2033, opened 2023-03-23
solution

IvanTkachenko3301:


src/convertToObject.js
```diff
@@ -1,18 +1,32 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleCssObject = {};
+  const fromStringToArray = sourceString
+    .split(';')
+    .map(elem => elem.split(':'))
+    .filter(elem => elem.length !== 1)
+    .map(elem => elem.map(elem2 => elem2.trim()));
```
Olekk17, line 20: It's been often said that we should not name variables like 'elem', maybe u need to replace it with something with more context

src/convertToObject.js
```diff
@@ -1,18 +1,32 @@
...
+
+  fromStringToArray.forEach(item => {
+    const key = item[0];
+    const property = item[1];
+
+    styleCssObject[key] = property;
+  });
```
Olekk17, line 27: Learn about Object.fromEntries method:
https://www.programiz.com/javascript/library/object/fromEntries

Olekk17 (approved):

src/convertToObject.js
liliia-livandovskaya: Yeap, `elem` is a bad name for the variable

liliia-livandovskaya (changes requested):
Great job! Pay attention to the comments

PR #2034, opened 2023-03-23
add solution

Pozdnya:


TanyaS133 (approved):

src/convertToObject.js
```diff
@@ -11,8 +11,30 @@
  *
  * @return {object}
  */
+/* eslint-enable */
```
SerhiiKirik, line 14:
```suggestion
```

SerhiiKirik: Do You have some errors with this text?

src/convertToObject.js
```diff
@@ -11,8 +11,30 @@
...
+      .trim())
+    .filter(property => property !== '');
+
+  return cssProperties.reduce((prev, item) => {
+    const key = item
+      .slice(0, item.indexOf(':'))
+      .trim();
+    const value = item
+      .slice(item.indexOf(':') + 1)
+      .trim();
```
SerhiiKirik, line 31:
```suggestion
  return cssProperties.reduce((prev, item) => {
  const indexOfDots = item.indexOf(':');
    const key = item
      .slice(0, indexOfDots)
      .trim();
    const value = item
      .slice(indexOfDots + 1)
      .trim();
```

SerhiiKirik (approved):
Great 🚀

src/convertToObject.js
Pozdnya: No, it's only for comments. Linter has shown an error.

src/convertToObject.js
Pozdnya: Done.

PR #2042, opened 2023-03-24
Solution

TanyaS133:


src/convertToObject.js
```diff
@@ -3,16 +3,40 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const listOfStyles = {};
+
+  const normalizeStyles = style => (
+    style
+      .split(':')
+      .map(x => x.trim())
+      .join(': ')
```
jstmpelowycz, line 23: I think, you don't need to join it back, you'll split it up regardlessly

jstmpelowycz (approved):

PR #2052, opened 2023-03-26
convert string to object

ofenchuk:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let styles = sourceString.split(';');
+
+  styles = styles.map(line => line.split(':'));
+
+  styles = styles.filter(line => line.length === 2);
```
VinnieJ-2k20, line 21: Use chained methods instead of reassigning to let variable.

VinnieJ-2k20 (changes requested):
Almost done. One small change left.

SerhiiKirik (approved):
Great 🔥

PR #2057, opened 2023-03-28
Solution

Zibi95:


Radoslaw-Czerniawski (changes requested):
Remove those comments left by Mate Academy, they are breaking the test pipeline. Your function works as it should as I logged some parts of its implementation to be sure, however to be certain I want it to pass our tests.

DorotaLeniecDev (approved):
Looking good :)

PR #2059, opened 2023-03-29
solution

DanielSorokowski:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+
+  const css = {};
+
+  properties.forEach(propertie => {
+    const key = propertie.split(':')[0];
+    const value = propertie.split(':')[1];
+
+    if (key && value) {
+      const keyClear = propertie.split(':')[0].replace(/\s/g, '');
+      const valueClear = propertie.split(':')[1].replace(/^\s+|\s+$/g, '');
```
solaryasha, line 27: Just use replace method on `key` and `value`

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+
+      css[`${keyClear}`] = `${valueClear}`;
```
solaryasha, line 29:
```suggestion
      css[keyClear] = valueClear;
```

solaryasha (changes requested):

PR #2060, opened 2023-03-29
solution

popeye88:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString.split('\n').filter(Boolean);
+  const obj = {};
+
+  for (const style of styles) {
```
Let's use iteration method instead

polosanya (changes requested):
Almost done! Please remake your solution without using loops

PR #2062, opened 2023-03-30
commit

bohdanmayor:


src/convertToObject.js
```diff
@@ -1,18 +1,21 @@
...
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
   // write your code here
+  const cssParams = {};
+  // write your code here
+  const splitByNewLine = sourceString.split(';');
+
+  for (let i = 0; i < splitByNewLine.length; i++) {
```
VitaliyBondarenko1982, line 9: good job.
let's replace for loop on forEach or better on reduce with initial value {}

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #2064, opened 2023-03-30
develop

nata-prog123:


src/convertToObject.js
```diff
@@ -1,18 +1,36 @@
...
+  const objectResult = {};
+  const additional = sourceString.split(';');
+
+  for (const key of additional) {
+    title = key.trim();
+
+    if (title !== '') {
+      num = title.indexOf(':');
+      property = (title.slice(num + 1, title.length)).trim();
+      title = (title.slice(0, num)).trim();
```
tetiana-ivanchenko, line 28: if you need to use some variables only in this scope, no need to create and define it in the upper scope

tetiana-ivanchenko (changes requested):

mykhalenych (approved):

PR #2065, opened 2023-03-31
Solution

katiailchenko:


package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```
ivadimko, line 19: Nice job updating the version! 🎉 However, if there were any breaking changes between versions, make sure to test the project fully and update any necessary code.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString.split(';').map(str => str.trim());
+  const result = {};
+
+  styles.forEach(string => {
+    const property = string.split(':');
```
ivadimko, line 21: Consider using more descriptive variable names like 'propertyPair' instead of 'property'.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+
+    if (property[0] && property[1]) {
```
ivadimko, line 23: nit: Instead of checking 'property[0]' and 'property[1]', use array destructuring and check for 'key' and 'value' directly. For example, 'const [key, value] = property;'

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
 /**
  * Implement convertToObject function:
  *
- * Function takes a string with styles (see an example in [stylesString.js](./stylesString.js))
+ * Function takes a string with styles (see an example in
+ * [stylesString.js](./stylesString.js))
  * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
```
ivadimko, line 8: nit: Typo in 'exampl' should be 'example'. 😸

ivadimko (approved):

PR #2068, opened 2023-04-01
Solution

Vasyl-Zhyliakov:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const properties = sourceString
+    .split('\n')
```
Volodymyr-Melnyk-Fron10der, line 18: it's better to split by `;`

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    .filter(prop => prop.length > 5);
```
Volodymyr-Melnyk-Fron10der, line 19: let's trim the value and just check lendth  `prop.trim().length`

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  const propList = {};
+
+  for (const char of properties) {
+    const index1 = char.indexOf(':');
+    const index2 = char.indexOf(';');
+
+    propList[char.slice(0, index1).trim()]
+      = char.slice(index1 + 1, index2).trim();
```
Volodymyr-Melnyk-Fron10der, line 28: slit every property by `:` so 0 element is a key and 1st is a value

Volodymyr-Melnyk-Fron10der (changes requested):

nazarmatsevych (approved):

PR #2074, opened 2023-04-05
Solution

yurii-shkrobut-m:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const array = sourceString.split(';');
+  const object = {};
+
+  array.forEach(element => {
```
DarkMistyRoom, line 20: image

DarkMistyRoom (changes requested):
Almost there. Use more descriptive names

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayProperties = sourceString.split(';');
+  const objectProperties = {};
+
+  arrayProperties.forEach(element => {
+    const keyValue = element.split(':');
+
+    if (keyValue.length === 2) {
+      objectProperties[keyValue[0].trim()] = keyValue[1].trim();
```
SerhiiKirik, line 24: use destructuring here
```suggestion
    const [key, value] = element.split(':');

    if (keyValue.length === 2) {
      objectProperties[key.trim()] = value.trim();
```

SerhiiKirik (changes requested):
Well done 🚀 
Let's make some improvements 🧑‍💻

SerhiiKirik (approved):
LGTM 🔥

PR #2079, opened 2023-04-06
solution

ghostargg:


YBM3 (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+    .split(';')
+    .map(property => {
+      return property
+        .split(':')
+        .map(element => element.trim())
+        .filter(element => element !== '');
+    })
+    .filter(empty => empty.length)
+    .sort(([a], [b]) => a.localeCompare(b))
+    .reduce((styles, currentProperty) => {
```
d-wasilewski, line 28: Try [array destructuring](https://www.freecodecamp.org/news/array-destructuring-in-es6-30e398f21d10/) here, you will avoid using mysterious values like property[0]

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+  return sourceString
+    .split(';')
+    .map(property => {
+      return property
+        .split(':')
+        .map(element => element.trim())
+        .filter(element => element !== '');
+    })
+    .filter(empty => empty.length)
+    .sort(([a], [b]) => a.localeCompare(b))
```
d-wasilewski, line 27: it's not necessary

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  */
 function convertToObject(sourceString) {
   // write your code here
+  return sourceString
+    .split(';')
+    .map(property => {
+      return property
+        .split(':')
+        .map(element => element.trim())
+        .filter(element => element !== '');
```
d-wasilewski, line 24: The filter(Boolean) method is commonly used in JavaScript to remove falsy values from an array. When filter method is called with Boolean as the condition, the filter method will remove any falsy values (such as your "") from the array.

To better visualize this example, `filter(Boolean)` is equal to `filter(element => Boolean(element))` and Boolean("") evaluates to false in your case
It's more like a trick, but it's good to know it :D

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2083, opened 2023-04-06
solution v1

bjeziorek:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+    lineValue = lineValue.trim();
+
+    return [lineName, lineValue];
+  });
+
+  const style = {};
+
+  lines.forEach(styleLine => (style[styleLine[0]] = styleLine[1]));
+
+  delete style._____toRemove;
```
MikiOOOO, line 31: interesting approach :) !

MikiOOOO (approved):

src/convertToObject.js
d-wasilewski: It is, but try doing it without using the `delete` keyword 🤔. Try using `filter` method to filter out the bad lines

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2084, opened 2023-04-06
1st commit

ihor-halkin:


witflash (approved):
This solution is good if you don't have rules with commas (for example: `transition-property: all, height, color;`).
So if there are no such cases in tests you can let it as it.
But you can think about alternative solutions if you want.

PR #2085, opened 2023-04-06
First solution

marshalek37:


src/convertToObject.js
```diff
@@ -1,18 +1,28 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const lineSplitted = sourceString.split('\n');
```
danielpopek94, line 4: you dont need to split lines

src/convertToObject.js
```diff
@@ -1,18 +1,28 @@
...
+  const stylesObject = lineSplitted.reduce((start, line) => {
+    const lineTrimmed = line.trim();
+
+    if (lineTrimmed === '' || lineTrimmed.startsWith(';')) {
```
danielpopek94, line 8: I recommend to use split with ';' before reduce()

src/convertToObject.js
```diff
@@ -1,18 +1,28 @@
...
+      return start;
+    }
+
+    const [propertyName, propertyValue] = lineTrimmed
+      .split(':')
+      .map(cssProperty => cssProperty.trim())
+      .map(propertyValueFixed => propertyValueFixed
```
danielpopek94, line 15: one trim here is enough, when you declare [propertyName, propertyValue]

src/convertToObject.js
```diff
@@ -1,18 +1,28 @@
...
+        .trim()
+        .replace(/\s*;\s*$/, ''));
+
+    return {
+      ...start,
+      [propertyName]: propertyValue,
+    };
```
danielpopek94, line 22: looks good :D

danielpopek94 (changes requested):

PR #2086, opened 2023-04-06
solution

sylcym:


natalia-klonowska (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+  const sourceObject = {};
+
+  const sourceArray = sourceString
+    .split(';')
+    .map(item => item.split(':').map(el => el.trim()))
+    .filter(el => el.length > 1);
+
+  for (const subArray of sourceArray) {
+    sourceObject[subArray[0]] = subArray[1];
+  }
```
d-wasilewski, line 26: You can use array destructuring in the for loop to make the code more concise, easier to read and to avoid unnamed variables like array[0]. It wi Here is how to do it - https://www.freecodecamp.org/news/array-destructuring-in-es6-30e398f21d10/

d-wasilewski (changes requested):

src/convertToObject.js
sylcym: done

d-wasilewski (approved):

PR #2089, opened 2023-04-06
solution

natalia-klonowska:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+  const convertedString = {};
+
+  for (const style of styles) {
+    const [key, value] = style
+      .split(':')
+      .map((item) => item.trim());
+
+    convertedString[key] = value
+      .split('')
+      .filter(char => char !== ';').join('').trim();
```
d-wasilewski, line 31: try splitting initially by `;` instead of `\n` so you will avoid this replacement

d-wasilewski (changes requested):

PR #2090, opened 2023-04-07
add solution

PufAnimalsDev:


src/convertToObject.js
```diff
@@ -13,6 +13,18 @@
...
   // write your code here
+  const cssStyles = {};
+
+  const cssPropertiesAndValues = sourceString
+    .split(';')
+    .map(propertyOrValue => propertyOrValue.split(':'))
+    .filter(propertyOrValue => propertyOrValue.length === 2);
+
+  for (const propertyOrValue of cssPropertiesAndValues) {
+    cssStyles[propertyOrValue[0].trim()] = propertyOrValue[1].trim();
```
d-wasilewski, line 24: Try [array destructuring](https://www.freecodecamp.org/news/array-destructuring-in-es6-30e398f21d10/) here, you will avoid using mysterious values like property[0]

d-wasilewski (changes requested):

kasiakubisiak (approved):

PR #2097, opened 2023-04-10
Solution

AnastassiaH:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const res = {};
+
+  sourceString.split(';')
+    .map(item => item.split(':').map(el => el.trim()))
+    .filter(item => item[0].length)
```
IvanFesenko, line 22: That might be an error, and better return boolean instead of number

IvanFesenko (changes requested):

Viktor-Kost (approved):

PR #2098, opened 2023-04-10
solve

ctgdprzyb:


Dka888 (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let cssArray = sourceString.replace(/(\r\n|\n|\r)/gm, '').split(';');
+
+  for (let i = 0; i < cssArray.length; i++) {
+    cssArray[i] = cssArray[i].trim();
+  };
```
d-wasilewski, line 21: use map for such operations

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+
+  cssArray = cssArray
+    .map(el => el.split(':'))
+    .filter(el => el.includes('') !== true);
```
d-wasilewski, line 25: .filter(Boolean) - [article](https://michaeluloth.com/filter-boolean)

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+  cssArray.forEach((el) => {
+    el[0] = el[0].trim();
+    el[1] = el[1].trim();
+  });
+
+  const cssObj = {};
+
+  cssArray.map((el) => {
+    cssObj[el[0]] = el[1];
+  });
```
d-wasilewski, line 36: do it in 1 loop, try to use [array destructuring](https://www.freecodecamp.org/news/array-destructuring-in-es6-30e398f21d10/) to avoid using things like el[0]

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+    .filter(el => el.includes('') !== true);
+
+  cssArray.forEach((el) => {
+    el[0] = el[0].trim();
+    el[1] = el[1].trim();
+  });
+
+  const cssObj = {};
+
+  cssArray.map((el) => {
```
d-wasilewski, line 34: don't use map if you don't want to return a new array

d-wasilewski (changes requested):

PR #2099, opened 2023-04-10
first version

Vladimir-Zadorozhnyi:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString.split(';');
+
+  const result = {};
```

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+
+  for (let i = 0; i < styles.length; i++) {
```
polosanya, line 22: image

polosanya (changes requested):
Almost done! Please remake your solution according to the checklist

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
-  // write your code here
+  const styles = sourceString.split(';');
+
+  const total = {};
+
+  styles.forEach(style => {
+    const parts = style.split(':');
+
+    if (parts.length === 2) {
+      const key = parts[0].trim();
```
polosanya, line 26: Consider using destructuring to simplify the code: `const [key, value] = parts.map(part => part.trim());`

polosanya (approved):
Good job!

PR #2102, opened 2023-04-11
solution

Amsko6:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectFromCss = {};
+
+  const cssLines = sourceString.split(';').map(line => {
+    const lineArray = line.split(':');
+    const lineKey = lineArray[0];
+    const lineValue = lineArray[1];
```
d-wasilewski, line 22: [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)
```suggestion
    const [lineKey, lineValue] = line.split(':');
```

d-wasilewski (changes requested):

PR #2103, opened 2023-04-11
solution

Ozwiena:


SuaBy514 (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splitExpressions = sourceString
+    .split(';')
+    .join('$')
+    .split(':')
+    .join('$')
+    .split('$');
```
d-wasilewski, line 22: what's the reason for doing that with `$`?

src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
+
+  const trimmedExpressions = splitExpressions.map(x => x.trim());
+
+  const codeExpressions = trimmedExpressions.filter(x => x.length > 0);
```
d-wasilewski, line 26: you can just chain these functions

src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
+
+  const entries = codeExpressions.reduce((rows, key, index) => (index % 2 === 0
+    ? rows.push([key])
+    : rows[rows.length - 1].push(key)) && rows, []);
```
d-wasilewski, line 30: it's not very clear what it does, maybe just try to create an empty object and then assign the key and values to it like `sourceObject[key] = value;` in a loop

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splitExpressions = sourceString
+    .split(';')
+    .join(',')
+    .split(':')
+    .join(',')
+    .split(',');
```
d-wasilewski, line 22: isn't it redundant?

d-wasilewski (changes requested):

src/convertToObject.js
Ozwiena: I think it is not redundant - when I remove this code, it does not work as expected

PR #2106, opened 2023-04-11
Solution

SuaBy514:


d-wasilewski (changes requested):
please refactor your code to clear white spaces with trim() method, it will save you a lot of effort

kasiakubisiak (approved):

d-wasilewski (approved):

PR #2111, opened 2023-04-12
solution js_style-to-object

stnayn:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObj = {};
+  let newStr = sourceString.replace(/[\r\n]+/g, '').split(';');
+
+  newStr = newStr.map((str) => str.trim().split(':'));
```
tetiana-ivanchenko, line 21: your variable name "newStr" but you put array into it

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  newStr.forEach((str) => {
```
tetiana-ivanchenko, line 23:
```suggestion
  newStr.forEach(([key, value]) => {
```

tetiana-ivanchenko (changes requested):

AlisaVlasova (approved):

PR #2116, opened 2023-04-14
solution

mariuszgit:


src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesTab = sourceString.split(';').filter(el => el.trim() !== '');
```
DorotaLeniecDev, line 10: According to the checklist:
_All chained methods called on the same level should have same indentation. This makes code much more readable._
```suggestion
  const stylesTab = sourceString
  .split(';')
  .filter(el => el.trim() !== '');
```

Also I would name this variable **stylesArr** instead of **stylesTab** (as array is more intuitive)
It would be also good to find better and more descriptive name for **el**

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
+  const resultObj = {};
+
+  stylesTab.forEach(style => {
+    const [property, value] = style.split(':').map(el => el.trim());
```
DorotaLeniecDev, line 14: Same here fix chained methods alignment, and check if you can use more descriptive name for **el**

DorotaLeniecDev (changes requested):
Very good solution, make some small fixes to meet checklist criteria

DorotaLeniecDev (approved):
Good job :)

PR #2123, opened 2023-04-19
Solution

Ir-ra:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
polosanya, line 17: Let's use more meaningful names
image

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+  const splitedString = sourceString.replace(/\n/g, '').split(';');
```
polosanya, line 18:
```suggestion
  const splitedString = sourceString
    .replace(/\n/g, '')
    .split(';');
```
image

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  for (let i = 0; i < splitedString.length; i++) {
```

polosanya (changes requested):
Almost done! Please redo the solution without using loops

sTorba24 (approved):
Cool!

PR #2127, opened 2023-04-22
Solution

vlad-liakh:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+
+  sourceString.split(';').forEach(element => {
+    const str = element.split(':');
+
+    if (str[0] !== undefined && str[1] !== undefined) {
+      obj[str[0].trim()] = str[1].trim();
```
etojeDenys, line 23: Use destructuring as mentioned earlier. Replace `obj[str[0].trim()] = str[1].trim();` with `obj[property.trim()] = value.trim();` 🧹

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+
+  sourceString.split(';').forEach(element => {
+    const str = element.split(':');
+
+    if (str[0] !== undefined && str[1] !== undefined) {
```
etojeDenys, line 22: Instead of checking for `undefined`, use the truthy check for both `property` and `value`. Replace `if (str[0] !== undefined && str[1] !== undefined)` with `if (property && value)` 🌟

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+
+  sourceString.split(';').forEach(element => {
```
etojeDenys, line 19: Use destructuring to make the code cleaner and more readable. Replace `const str = element.split(':');` with `const [property, value] = element.split(':');` 🧹

etojeDenys (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    const str = element.split(':');
+    const [porerty, value] = str;
```
ivanchuk-yurii, line 21:
```suggestion
    const [porerty, value] = element.split(':');
```

ivanchuk-yurii (approved):

PR #2132, opened 2023-04-26
chore: add solution

NikitaNevv:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssObject = {};
+  const makeArray = sourceString.replace(/(\n|\r)/gm, '')
+    .split(':').map(item => item.trim()).join(':').split(';')
+    .map(item => item.trim()).filter(item => item !== '');
```
Volodymyr-Melnyk-Fron10der, line 21: Instead of using multiple .map() and .filter() calls, you can use a single .reduce() function to create the cssObject.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssObject = {};
+  const makeArray = sourceString.replace(/(\n|\r)/gm, '')
+    .split(':').map(item => item.trim()).join(':').split(';')
```
Volodymyr-Melnyk-Fron10der, line 20: It's not necessary to split by ':' and join back with ':' again. You can split by ';' directly after removing newlines.

Volodymyr-Melnyk-Fron10der (changes requested):
--

qoa11a (approved):
Well done! 👍

PR #2138, opened 2023-04-28
Convert_style_string_to_object

rialleons:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
 /**
  * Implement convertToObject function:
  *
- * Function takes a string with styles (see an example in [stylesString.js](./stylesString.js))
+ * Function takes a string with styles
+ * (see an example in [stylesString.js](./stylesString.js))
  * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
+ * and values are the values of related CSS properties
+ * (see an example in [test file](./convertToObject.test.js))
```
OleziO, line 10: Remove all comments from the code.

OleziO (changes requested):

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
- *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const lines = sourceString.split('\n');
+  const filledLines = lines.filter(line => line.match(/[a-z]/));
+  const semicolonRemove = filledLines.map(line => line.slice(0, -1));
```
witflash, line 11: You could split by `;` I think, but your solution is also ok

witflash (approved):
Good job 👍

src/convertToObject.js
rialleons: Thank you for the tip) It is one full iteration shorter your way.
```suggestion
  const lines = sourceString.split(';');
  const filledLines = lines.filter(line => line.match(/[a-z]/));
```

PR #2140, opened 2023-04-29
solution

Absolution26:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resultObjStyles = {};
+  const arrKeyValue = sourceString.split(';').map(el => el.trim());
+
+  for (let i = 0; i < arrKeyValue.length; i++) {
```
agniya-i, line 20: you can also try to use `reduce` method here

agniya-i (approved):
I approved, but pay attention to comment below

PR #2143, opened 2023-04-30
converToObject solution

IzzetYusufov:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+    if (elem[1] === undefined) {
+      return { ...prev };
+    } else {
+      const firstElem = elem[0].trim();
+      const secondElem = elem[1].trim();
+
+      return {
+        ...prev,
+        [firstElem]: secondElem,
+      };
```
nazarmatsevych, line 30:
```suggestion
    const [ key, value ] = formArr;
    
    if (key && value) {
      prev[key.trim()] = value.trim();
    }
    
    return prev;
```

We can destructure array and make solution cleaner and simpler

nazarmatsevych (approved):
Good work,
left some comments to improve your solution

PR #2146, opened 2023-05-02
solution

generatedman:


src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString.split(';');
+  const stylesObject = {};
```
ruslanpashkov, line 16: you don't need to specify what data types are stored in your variables, especially if it's arrays/objects

javascript is a dynamically typed language and we need the flexibility to change data types at runtime, so let's think about other names for our variables

ruslanpashkov (changes requested):

33Orange (approved):
Code working.

For self improve: 
Try to do same task but using reduce.

PR #2148, opened 2023-05-03
1.0.0 initial solution

domkab:
!

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -1,18 +1,20 @@
 'use strict';
 
-/**
- * Implement convertToObject function:
- *
- * Function takes a string with styles (see an example in [stylesString.js](./stylesString.js))
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .map(style => style.trim())
+    .filter(style => style !== '');
+
+  const styleObject = {};
+
+  styles.forEach(style => {
+    const [key, value] = style.split(':').map(str => str.trim());
+
+    styleObject[key] = value;
+  });
+
+  return styleObject;
 }
 
 module.exports = convertToObject;
--- a/src/stylesString.js
+++ b/src/stylesString.js
@@ -1,13 +1,26 @@
+/* eslint-disable max-len */
 'use strict';
 
+/**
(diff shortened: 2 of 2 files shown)


src/stylesString.js
```diff
@@ -1,13 +1,26 @@
...
+ * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
+ *
+ * @param {string} sourceString
+ *
+ * @return {object}
+ */
+
 const stylesString = `
-  background-color:      #fff;
+  background-color: #fff;
```
Anastasiia-Svintsova, line 17: Don't change this file

src/convertToObject.js
```diff
@@ -1,18 +1,20 @@
...
+    .filter(style => style !== '');
+
+  const styleObject = {};
+
+  styles.forEach(style => {
+    const [key, value] = style.split(':').map(str => str.trim());
+
+    styleObject[key] = value;
+  });
+
```
Anastasiia-Svintsova, line 16: Use the reduce method instead

Anastasiia-Svintsova (changes requested):

src/convertToObject.js
```diff
@@ -1,18 +1,20 @@
...
+  const styleObject = styles.reduce((acc, style) => {
+    const [key, value] = style.split(':').map(str => str.trim());
+
+    acc[key] = value;
+
+    return acc;
+  }, {});
+
+  return styleObject;
```
anstsot, line 17: you can proceed with a chain of methods and avoid creating additional variables
```suggestion
    .filter(style => style !== '')
    .reduce((acc, style) => {
      const [key, value] = style.split(':').map(str => str.trim());

      acc[key] = value;

      return acc;
    }, {});

  return styles;
```

src/convertToObject.js
```diff
@@ -1,18 +1,20 @@
...
+    .split(';')
+    .map(style => style.trim())
+    .filter(style => style !== '');
+
+  const styleObject = styles.reduce((acc, style) => {
+    const [key, value] = style.split(':').map(str => str.trim());
+
+    acc[key] = value;
+
+    return acc;
```
anstsot, line 14: try to avoid variable modifying. you can use spread operator
```suggestion
    return { ...acc, [key]: value };
```

anstsot (changes requested):
It’s almost done! Let’s improve a little

anstsot (approved):
I'm approving but the previous comment regarding spread operator isn't fixed

domkab:
Oh sorry, i tried to fix it but it didnt work at first because i didnt specify object as initial value in reduce, will add that to return

PR #2158, opened 2023-05-06
MyPainObject

Oksana-Maruniak:


TarasHoliuk (approved):
The solution is laconic and works well. But next time you can be more careful with the variable names (they should be more descriptive).

PR #2162, opened 2023-05-08
add convertToObject solution

DidenkoDenis306:


ruslanpashkov (approved):
for self improve:
try to do the same task but use reduce

Anna-Tkachenko (approved):
⭐️

PR #2164, opened 2023-05-09
add task solution

maksym-nemera:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+
+    if (key.trim().length !== 0) {
+      return {
+        ...newObject,
+        [key.toString().trim()]: [value].toString().trim(),
+      };
+    }
+
+    return { ...newObject };
+  };
```
ruslanpashkov, line 30: do we need to create a separate callback variable, when we can pass it directly in our `reduce`?

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
-  // write your code here
+  const eachString = sourceString.split(';');
+
+  const callback = (newObject, stringStyle) => {
+    const [key, value] = stringStyle.split(':');
+
+    if (key.trim().length !== 0) {
+      return {
+        ...newObject,
+        [key.toString().trim()]: [value].toString().trim(),
```
ruslanpashkov, line 25: do we need to convert our properties to the string, when `typeof sourceString === 'string'`?

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const eachString = sourceString.split(';');
+
+  const callback = (newObject, stringStyle) => {
+    const [key, value] = stringStyle.split(':');
+
+    if (key.trim().length !== 0) {
```
ruslanpashkov, line 22: we can use one of these options here

```suggestion
    if (key) {
```

```suggestion
    if (value) {
```

src/convertToObject.js
ruslanpashkov, line 25: you don't need an array literal for value here

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+      return {
+        ...newObject,
+        [key.toString().trim()]: [value].toString().trim(),
+      };
+    }
+
+    return { ...newObject };
+  };
+
+  const result = eachString.reduce(callback, {});
```
ruslanpashkov, line 32: let's give a clearer name, to specify what data will be stored in our variable

ruslanpashkov (changes requested):

naziamoff (approved):
🔥

PR #2165, opened 2023-05-10
Solution

andrijdudar:


sTorba24 (changes requested):
Yes, i like second variant too. So use it

tetiana-ivanchenko (approved):

tetiana-ivanchenko (approved):

PR #2167, opened 2023-05-10
task solved

zisuan-od:


src/convertToObject.js
```diff
@@ -12,7 +12,12 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssObject = {};
+  const newString = sourceString.split(';\n').filter(a => a.includes(':'));
```
slavanosatskyi, line 16: it is always a good idea to choose meaningful names. It will make your code more readable. 'a' or 'b' is not the best  option

slavanosatskyi (approved):
Pay attention to your variable names.

PR #2170, opened 2023-05-10
Solution

denys2:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const makeAnArray = sourceString.replace(/(\n|\r)/gm, '').split(';');
+  const cssObject = {};
+
+  makeAnArray.map((item) => {
```
agniya-i, line 21:
```suggestion
  makeAnArray.forEach((item) => {
```

better to use `forEach` to just  iterate through an array, 
`map` is usually used to create a new array

agniya-i (approved):

PR #2174, opened 2023-05-12
Solution

BogdanParubchenk01:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const valueStyles = sourceString.split(';')
+    .map(style => style.split(':')
+      .map(data => data.trim()))
+    .filter(style => style.length > 1);
```
VitaliyBondarenko1982, line 20: instead of two map and filter iterations consider to make everything in reduce method
```suggestion
  const valueStyles = sourceString.split(';')
```

VitaliyBondarenko1982 (changes requested):

anstsot (approved):
Great job, approved!

PR #2177, opened 2023-05-13
Solution

vitalii-fedusov:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArray = sourceString.split(';');
+  const filtered = sourceArray.filter(item => item.includes(':'));
+
+  const arr = filtered.map((item) => {
+    const [key, value] = item.split(':');
+
+    return [key.trim(), value.trim()];
+  });
```
VitaliyBondarenko1982, line 24: consider to avoid fileter and map iterations and make everything in reduce

VitaliyBondarenko1982 (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
-  // write your code here
+  const sourceArray = sourceString.split(';');
+  const result = sourceArray.reduce((prev, styles) => {
+    if (!styles.includes(':')) {
+      return prev;
+    }
+
+    const [key, value] = styles.split(':');
+
+    if (key) {
```
etojeDenys, line 25: now this check is redundant I think

etojeDenys (changes requested):

Viktor-Kost (approved):

PR #2180, opened 2023-05-15
solution

skinnyjazz:


naziamoff (changes requested):
looks good, but let's now try `.reduce` method here

dinaminahi (approved):

naziamoff (approved):

PR #2182, opened 2023-05-16
solution

HannieA:


anstsot (approved):
Good job, approving.
But instead of creating additional `properiesObject` variable you can use `.reduce` method

PR #2183, opened 2023-05-17
Solution

yuliia-nudyk:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  for (let i = 0; i < sourceArray.length; i++) {
+    const sourceDesc = sourceArray[i].split(':');
+
+    if (sourceDesc.length === 2) {
+      const [key, value] = sourceDesc;
+
+      createdObj[key.trim()] = value.trim();
+    }
+  }
```
anstsot, line 29: try to no use the `for` loop and create separate `createdObj` variable.
you can use the chain of methods like `map`, `reduce` etc.

anstsot (changes requested):
There is a fix needed to improve your code.

Anastasiia-Svintsova (approved):

PR #2186, opened 2023-05-17
solution

artem-ivashchenko:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .map(string => string.trim())
+    .filter(string => string)
```
witflash, line 20: You can use `.filter(Boolean)` approach to filter empty strings

witflash (approved):
Well done 👍

PR #2188, opened 2023-05-17
add solution

niaYaro:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedFromString = {};
+  const stringToArray = sourceString
```
33Orange, line 18: It's recommended to use more descriptive variable names. Consider renaming 'convertedFromString' to 'styleObject'.

33Orange (changes requested):
Well done ! 🔥 

A few moments with readability

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+  const styleProperties = sourceString
+    .replace(/;/g, '').split('\n')
+    .filter(i => i.includes(':')).map(i => i.split(':')
+      .map(j => j.trim()).join(': '));
```
naziamoff, line 21: better to move each of the method call to a new line, because the way you have it now is totally not readable.
```suggestion
  const styleProperties = sourceString
    .replace(/;/g, '')
    .split('\n')
    .filter(i => i.includes(':'))
    .map(i => i.split(':')
      .map(j => j.trim())
      .join(': '));
```

naziamoff (changes requested):
great job 🔥 
Let's practice some reduce method here

naziamoff (approved):

PR #2189, opened 2023-05-17
add convertToObject function

AleksandrTsivilin:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let properties = sourceString.split(';');
```
33Orange, line 17: Split sourceString by '; ' instead of ';' to avoid extra whitespaces.

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+
+  properties = properties.slice(0, properties.length - 1);
+
+  properties = properties
+    .map(property => property.trim())
+    .filter(property => property.length !== 0);
+
+  properties = properties.map(property => {
+    return property.split(':');
+  });
```
33Orange, line 27: Array methods `slice()`, `map()`, `filter()` return new array. You can chain this functions instead of reassigning `properties` multiple times.

33Orange (changes requested):
Good job!

A few moments with readability.

src/convertToObject.js
AlexKherson: It doesn't work. There isn't space between the first and second lines; Maybe I'm not right. 
`background-color:      #fff;
-webkit-border-radius: 5px;` 

Here is an example.
`["
  background-color:      #fff", "
-webkit-border-radius: 5px", "
  border-radius     : 5px", "
  border: 1px solid #e8e8e8", "
  
"]
["
  background-color:      #fff;
-webkit-border-radius: 5px;
  border-radius     : 5px;
  border: 1px solid #e8e8e8;
  
"]`

naziamoff (changes requested):
looks good 👍 
But now let's practice `.reduce` method in this task

silchyn (approved):

PR #2200, opened 2023-05-18
convertToObject add solution

YeMyr:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+  const splitIt = trimIt.split(';');
+
+  const removeSpaces = splitIt
+    .map(prop => prop.split(':').map(item => item.trim()));
+
+  const removeEmpty = removeSpaces.filter(item => item[0] !== '');
+
+  removeEmpty.forEach((item) => {
+    result[item[0]] = item[1];
+  });
```
naziamoff, line 30: those methods return the new array, so instead of storing each result to a separate variable, you can just use a chain call instead. So let's do it

naziamoff (changes requested):

src/convertToObject.js
YeMyr: chain is done

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
-  // write your code here
+  const result = {};
+
+  sourceString.trim()
+    .split(';')
+    .map(prop => prop.split(':').map(item => item.trim()))
+    .filter(item => item[0] !== '')
+    .forEach((item) => {
+      result[item[0]] = item[1];
+    });
```
33Orange, line 25: Try to use destructuring assignment here.
This will improve the readability of the code.

```suggestion
    .forEach(([key, value]) => {
      result[key] = item[value];
    });
```

33Orange (approved):
Well done!


Pay attention to the comment

src/convertToObject.js
YeMyr: 👌made it, looks much better))) thanks!

naziamoff (approved):

PR #2207, opened 2023-05-18
Solution

annholovko:


naziamoff (approved):
awesome 🔥 
Also could be done just inside a single reduce method

PR #2209, opened 2023-05-18
Solution

ivankhrsh:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splitStr = new Set(sourceString.split(';'));
+  const result = {};
```
naziamoff, line 18: result - 🙅‍♀️ 🙅‍♀️ 🙅‍♀️ 
Let's think of a more meaningful name

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  splitStr.forEach(function(param) {
```
naziamoff, line 20: lets remove function declaration from here - it can work just fine without it

naziamoff (changes requested):
not bad, there are some small suggestions though

DidenkoDenis306 (approved):
Well done!

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  new Set(sourceString.split(';')).forEach(function(param) {
```
dinaminahi, line 19: Let's use arrow function as a callback to `forEach`
`function(param) { `  =>   `param => {`

dinaminahi (changes requested):
Almost there, need to fix one moment

dinaminahi (approved):
Good job!

naziamoff (approved):

PR #2210, opened 2023-05-19
add task solution

KatyaZaremska:


src/convertToObject.js
```diff
@@ -1,18 +1,18 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const arraySourcesString = sourceString.split(';');
+  const objectOfValues = {};
```
naziamoff, line 5: you should not add the data type to the name of the variable as this does not give us any additional info

naziamoff (approved):
good job 👍 

I would, however, suggest you to practice some reduce method here

PR #2213, opened 2023-05-19
add semi-solution

ulositska:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+  const properties = sourceString
+    .split(';')
+    .map(property => property.split(':'))
+    .filter(property => (property.length > 1));
+
+  const stylesObject = [...properties].reduce((obj, currentProperty) => {
+    const key = currentProperty[0].trim();
+    let value = currentProperty[1].trim();
+
+    value = value.split(':').join('');
```
dinaminahi, line 26: You already splited by `:` in line 19, no need to do that again, because no `:` are left

dinaminahi (changes requested):
Good job!
But there is one moment to fix

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+  const stylesObject = [...properties].reduce((obj, currentProperty) => {
+    const key = currentProperty[0].trim();
+    const value = currentProperty[1].trim();
+
+    return {
+      ...obj, [key]: value,
+    };
+  }, {});
+
+  return stylesObject;
```
silchyn, line 31:
```suggestion
  return [...properties].reduce((obj, currentProperty) => {
    const key = currentProperty[0].trim();
    const value = currentProperty[1].trim();

    return {
      ...obj,
      [key]: value,
    };
  }, {});
```

silchyn (approved):

PR #2215, opened 2023-05-19
solution

kirilshorin:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const objectOfStyles = {};
+  const arrayOfStyles = splitAndTrim(sourceString, ';')
+    .filter(style => style.length > 0);
+  const splitedStyles = arrayOfStyles.map(style => splitAndTrim(style, ':'));
+
+  splitedStyles.map(function makeObj(style) {
+    objectOfStyles[style[0]] = style[1];
+  });
```
33Orange, line 24: You don't need to use `map` here.
Map return new array. You can use `forEach` instead `map` in this case.

Use arrow functions for callbacks inside methods.

33Orange (changes requested):
Some comments:
- Use `forEach` instead of map when not returning a new array
- Use arrow functions for callbacks.

Good idea create a separated function with methods that repeat.

33Orange (changes requested):
1. Check my previous comment about arrow functions in callbacks.
2. Let's change logic a little bit and remove redundant iterations

dinaminahi (changes requested):
Please, review @33Orange comments

PR #2218, opened 2023-05-20
style in obj

diana-tuz:


src/convertToObject.js
```diff
@@ -1,18 +1,32 @@
...
+*
+* @return {object}
+*/
+function convertToObject(styles) {
+  const srt = styles.split(';');
+
+  const result = {};
+
+  for (let i = 0; i < srt.length; i++) {
+    const parts = srt[i].split(':');
```
nazarmatsevych, line 22:
```suggestion
    const [ key, value ]= srt[i].split(':');
```

Destructure array

src/convertToObject.js
```diff
@@ -1,18 +1,32 @@
...
+
+    if (parts[0] && parts[1]) {
```
```suggestion
    if (key && value) {
```

nazarmatsevych (approved):
Good work!

PR #2219, opened 2023-05-21
Solution

MaksimKad:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  sourceString
+    .split(';')
+    .map(string => string.trim())
+    .filter(symbol => symbol !== '')
```
33Orange, line 22: Consider using a more descriptive name for the 'symbol' variable, such as 'trimmedString'.

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    .forEach(property => {
+      const properties = property.split(':');
+      const key = properties[0];
+      const value = properties[1];
```
33Orange, line 26: You can use destructuring assignment instead. 

```
```

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  sourceString
+    .split(';')
+    .map(string => string.trim())
+    .filter(symbol => symbol !== '')
+    .forEach(property => {
```
33Orange, line 23: Change this `forEach` to `reduce` and create object as `initialValue` in reduce.
This object will be function result.

33Orange (changes requested):
Well done ! 🔥 

A few moments to fix:
- Use more descriptive names for params.
- Use destructuring assignment.

Let's change the `forEach` method to `reduce` with an empty object in the `initialValue`.

PR #2220, opened 2023-05-21
made task js style-to-object

vladyslav-lapshyn:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+    .split(';')
+    .map(string => string.trim())
+    .filter(string => string);
+
+  const stylesObject = {};
+
+  stylesArray.forEach(string => {
+    const property = string.split(':');
+    const propertyName = property[0].trim();
+    const propertyValue = property[1].trim();
```
33Orange, line 27: Use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) instead.

```
const [key, value] = string.split(':');
```

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString
+    .split(';')
+    .map(string => string.trim())
+    .filter(string => string);
+
+  const stylesObject = {};
+
+  stylesArray.forEach(string => {
```
33Orange, line 24: Use method `reduce` with empty object as `initialValue` instead.

33Orange (changes requested):
Well done ! 🔥 

A few moments: 
- Use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)

Let's try to change `forEach` to `reduce` with empty object as `initialValue`.

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+
+  const stylesObject = stylesArray.reduce((acc, string) => {
+    const [propertyName, propertyValue] = string
+      .split(':')
+      .map(stringPart => stringPart.trim());
+
+    return {
+      ...acc,
+      [propertyName]: propertyValue,
+    };
```
33Orange, line 30: **This piece of code return `NEW object` and overwrite the old one every iteration. !**

```suggestion
    const [propertyName, propertyValue] = string.split(':')

    acc[key.trim()] = value.trim();
   
    return acc;
```

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString
+    .split(';')
+    .map(string => string.trim())
+    .filter(string => string);
+
+  const stylesObject = stylesArray.reduce((acc, string) => {
```
33Orange, line 22: Why would you want to create a variable for a value that you will return right away?
```suggestion
  return stylesArray.reduce((acc, string) => {
```

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+      .split(':')
+      .map(stringPart => stringPart.trim());
+
+    return {
+      ...acc,
+      [propertyName]: propertyValue,
+    };
+  }, {});
+
+  return stylesObject;
```
33Orange, line 33:
```suggestion

```

33Orange (changes requested):
Almost done !

33Orange (approved):
Good job. 

Only you forgot to return the object right away, without creating a variable. But this is not a critical error.

PR #2221, opened 2023-05-21
solution

snuber-w:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    .map(element => element.split(':')
+      .map(item => item.trim()))
+    .filter(item => item.length > 1)
+    .reduce((prev, [key, value]) =>
+      ({
+        ...prev,
+        [key]: value,
+      }), {});
+
+  return result;
```
silchyn, line 29: that is not a CodeWars task! the solution is not readable at all

silchyn (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+    .filter(item => item.length > 1)
+    // filter from empty values (could have appeared after splitting)
+    .reduce((prev, [key, value]) =>
+    // we create an object where for each property we write its key and value
+      ({
+        ...prev,
+        [key]: value,
+      }), {});
+
+  return result;
```
silchyn, line 32: separate variables with descriptive names would bring just as much sense as comments

silchyn (changes requested):

src/convertToObject.js
snuber-w: hard to come up with variable names

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesSource = sourceString;
```
silchyn, line 17:
```suggestion
```

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+  const styleProperties = styleArray.map(element => element.split(':')
+    .map(item => item.trim())
+  );
+  const stylesFiltered = styleProperties.filter(item => item.length > 1);
+  const styles = stylesFiltered.reduce((prev, [key, value]) =>
+    ({
+      ...prev,
+      [key]: value,
+    }), {}
+  );
```
silchyn, line 29:
```suggestion
  const stylesWithoutLineBreaks = stylesSource.replace(/\n/gi, '');
  
  const styleArray = stylesWithoutLineBreaks.split(';');
  
  const styleProperties = styleArray.map(element => element.split(':')
    .map(item => item.trim())
  );
  
  const stylesFiltered = styleProperties.filter(item => item.length > 1);
  
  const styles = stylesFiltered.reduce((prev, [key, value]) =>
    ({
      ...prev,
      [key]: value,
    }), {}
  );
```

silchyn (approved):

PR #2223, opened 2023-05-21
sol

thienlaoo:


src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+ * and values are the values of related CSS propertiesjs))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splittedValues = sourceString.split('\n').map(string =>
+    string.replace(';', '').split(':'));
```
33Orange, line 16: You can initially split string by `;`. And you won't have to replace `;`

33Orange (changes requested):
1. Checklist p4: `[CODE STYLE] - All chained methods called on the same level should have same indentation. This makes code much more readable.`
2. Do you need all this variables? `splittedValues`, `trimmedValues`, `filteredValues` 
3. Do you need al this logic?
You can split `sourceString` and use 1 `forEach`, where you will be `split` values by `:` separator and use this `values.trim()` to fill your `convertedToObject` object.

PR #2225, opened 2023-05-22
solution

SofiaLiubchenko:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+
+  const propertyLines = sourceString.trim().split(';');
```
33Orange, line 19: Use `split(';')` first of all.
After that you can `trim` all your items in array.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  for (const line of propertyLines) {
+    const [key, value] = line.split(':').map(prop => prop.trim());
+
+    if (key && value) {
+      styles[key] = value;
+    }
+  }
```
33Orange, line 27: Let's try another way to solve this.
You can use `reduce()` with empty object as `initialValue` to create result object.

And don't forget about how to write chained methods. (Checklist p4)

```
const cssStyles = styles
 .map(style => style)
 .slice(0, 5)
```

33Orange (changes requested):
Almost done

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+      .map(prop => prop.trim());
+
+    if (key && value) {
+      result[key] = value;
+    }
+
+    return result;
+  }, {});
+
+  return styles;
```
33Orange, line 33: I think that it is not necessary to create a variable for the value that is immediately returned.

```suggestion
  return  styles = propertyLines.reduce((result, line) => {
    const [key, value] = line.split(':')

    if (key && value) {
      result[key.trim()] = value.trim();
    }

    return result;
  }, {});
```

33Orange (approved):
Well done

PR #2229, opened 2023-05-23
solution

maksym-kostetskyi:


loralevitska (changes requested):
Almost done) Let's improve

src/convertToObject.js
```diff
@@ -1,18 +1,21 @@
...
-  // write your code here
+  const result = {};
+  const sourceStringTrimmed = sourceString.trim();
+  const stringSplit = sourceStringTrimmed.split(';');
+
+  for (const entry of stringSplit) {
+    if (entry.includes(':')) {
+      const index = entry.indexOf(':');
+      const name = entry.slice(0, index).trim();
+      const value = entry.slice(index + 1, entry.length).trim();
```
DarkMistyRoom, line 12: `const splitedEntry = entry.split(':');
const name = splitedEntry[0].trim();
...`

DarkMistyRoom: It looks simpler, isn't it?)

src/convertToObject.js
```diff
@@ -1,18 +1,21 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const sourceStringTrimmed = sourceString.trim();
```
DarkMistyRoom, line 5: no need to trim string here because these spaces will be trimmed in 11-12 lines

DarkMistyRoom (approved):

PR #2238, opened 2023-05-27
solution

Gosyav:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArray = sourceString.split(';');
+
+  let trimmedWord;
+  let trimmedArray;
+  let result = {};
+
+  sourceArray.map(word => {
```
VitaliyBondarenko1982, line 23: it is better to make with reduce method

VitaliyBondarenko1982 (approved):
well done

IvanFesenko (approved):

PR #2240, opened 2023-05-29
add solution

dimarogkov:


src/convertToObject.js
```diff
@@ -1,18 +1,31 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const arr = sourceString.split(";");
```
Volodymyr-Melnyk-Fron10der, line 19: Consider using a more descriptive variable name instead of 'arr', e.g. 'stylePairs'.

src/convertToObject.js
```diff
@@ -1,18 +1,31 @@
...
+
+  return arr
+    .map((style) => style.trim().split(":"))
+    .filter(([property, value]) => property && value)
```
Volodymyr-Melnyk-Fron10der, line 23: you can do those stuff in the reduce method, so do it plz

Volodymyr-Melnyk-Fron10der (changes requested):

Volodymyr-Melnyk-Fron10der (changes requested):
-

OleziO (changes requested):

IvanFesenko (approved):
Well done

PR #2243, opened 2023-05-31
Add styles converter

AmadeuszAndroid:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString.split(';');
+  const stylesObject = {};
```
Radoslaw-Czerniawski, line 18: This empty object is redundant, you can pass it as starting value in the reduce and immediately return reduced val:
```js
 return stylesArray.reduce(
    (styles, style) => {
      const styleParts = style.split(':');
      const key = styleParts[0].trim();

      if (key !== '') {
        styles[key] = styleParts[1].trim();
      }

      return styles;
    }
    , {});
```

Radoslaw-Czerniawski: Also name of the styleParts could be more meaningful, `const propertyAndValue = style.split(':');`

Radoslaw-Czerniawski: I personally wouldn't declare any variables here and just return
```js
return sourceString.split(';').reduce(
    (styles, style) => {
      const styleParts = style.split(':');
      const key = styleParts[0].trim();

      if (key !== '') {
        styles[key] = styleParts[1].trim();
      }

      return styles;
    }
    , {});
```

Radoslaw-Czerniawski: And finally I would always return a new object with previous styles spread and new optional prop
```js
return {
  ...styles,
  ...(prop !== "" && { [prop]: propertyAndValue[1].trim() }),
};
```

so finally the function could look like this:
image

Radoslaw-Czerniawski (changes requested):
Everything is fine apart from the unnecessary declaration of `const stylesObject = {};` rather than passing empty object in reduce as second argument and just returning the result of reduce operation. Adjust according to the first comment, rest are more of a refactoring process I would personally implement here.

Generally in functional programming you want to avoid any side effects in this case mutation of outside dependencies -> modifying `stylesObject` in the process of reduce method.

src/convertToObject.js
AmadeuszAndroid: I shorten code.

darokrk (approved):
Nicely done 💪🏽

PR #2250, opened 2023-06-09
convertToObject Solution

ZupaFly:


src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
lerastarynets, line 10: Pls don't use consts with such names (see checklist point 2)

src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
+
+  const sourceStringSplit = sourceString.split(';');
+
+  for (const key of sourceStringSplit) {
```
lerastarynets, line 14: Pls avoid loops in this task (see checklist point 5). How about starting with `.map` and adding a few other chain methods to achieve the same?))

lerastarynets (changes requested):
I like the way you achieved this task :) however let's make it more consistent with the checklist

src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
```
lerastarynets, line 10: Pls see checklist point 2
[CODE STYLE] - Avoid variable namings like `array` , `object`, `arr`, `obj`, `result`. It will decrease code readability

lerastarynets (changes requested):
Almost perfect, but pls pay a little bit more attention to checklists, you'll need this skill while working with checklists for your real tasks on the future job)

src/convertToObject.js
ZupaFly: My bad, that was so common to use obj naming that i didnt pay atterntion to that part in checklist. Will improve it

lerastarynets (approved):
Well done :)

PR #2258, opened 2023-06-13
Solution

MykolaLopushenko:


IvanFesenko (approved):
Good job. For practice try resolve task with reduce method

PR #2262, opened 2023-06-13
add convert object

dubyuliia:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const pairs = sourceString.split(';').map(pair => pair.trim());
+  const expectedObj = pairs.reduce((item, pair) => {
+    if (pair) {
```
Viktor-Kost, line 19: The 'if (pair)' condition is not necessary if you filter out empty pairs.

Viktor-Kost (approved):
GJ!

PR #2270, opened 2023-06-15
solution

cebanoleksandr:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString
+    .split(';')
+    .map(style => style.trim())
+    .map(style => style.split(':'))
+    .map(style => style.map(st => st.trim()))
+    .map(style => style.filter(Boolean))
+    .filter(style => style.length === 2);
```
tetiana-ivanchenko, line 23: try to reduce count of methods and avoid nesting

tetiana-ivanchenko (changes requested):

PR #2280, opened 2023-06-17
solution

GGsmou:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+ * and values are the values of related CSS properties (see an exampl in
+ * [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styeObject = {};
```
IraShtynda, line 17: There is a grammatical error

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+  const stylesArray = sourceString
+    .split(';')
+    .filter((style) => style.trim() !== '');
+
+  stylesArray.forEach((style) => {
+    const [key, value] = style.split(':');
+
+    styeObject[key.trim()] = value.trim();
```
IraShtynda, line 25: You use the trim() function 3 times, you can move this functionality into a separate function

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styeObject = {};
+  const stylesArray = sourceString
+    .split(';')
+    .filter((style) => style.trim() !== '');
+
+  stylesArray.forEach((style) => {
+    const [key, value] = style.split(':');
```
IraShtynda, line 23: It might be better to use more readable names than key and value

IraShtynda (approved):
Great job! Please provide a few suggestions for the review.

src/convertToObject.js
GGsmou: Can't agree with this one
Why you need to write a separate function if you can use one method instead

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styeObject = {};
+  const stylesArray = sourceString
+    .split(';')
+    .filter((style) => style.trim() !== '');
```
```suggestion
    .filter((style) => style.trim());
```

tetiana-ivanchenko (approved):

PR #2289, opened 2023-06-19
Solution

valet2102:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+ *  (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+
+  const STYLE_OBJECT = {};
```
FrontKid, line 19: You cannot call ur vars with SCREAMING_SNAKE when the variable is evaluated during the program
(https://learn.javascript.ru/task/uppercast-constant)

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+    .map(item => item
+      .split(':')
+      .map(element => element
+        .trim()));
+
+  for (const [key, value] of STYLE_ARRAY) {
+    if (key !== undefined && value !== undefined) {
+      STYLE_OBJECT[key] = value;
+    }
+  }
```
FrontKid, line 31: you won't have '':undefined if u trim ur item before so u can try to use .reduce to create an object

FrontKid (changes requested):

PR #2294, opened 2023-06-19
Solution

ArtemBehter:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+  const arraySourceString = sourceString.split(';');
+
+  return arraySourceString.reduce((stylesObject, styles) => {
+    if (!styles.length) {
+      return stylesObject;
+    }
+
+    const [key, value] = styles.split(':').map((item) => item.trim());
+
+    if (!value) {
```
darokrk, line 27: You can try use here ternary operator instead of if:

```suggestion
    return !value
  ? stylesObject
  : {
      ...stylesObject,
      [key]: value,
    };
```

darokrk (approved):
Approved, good job 🥇 left one suggestion, you can fix it if you want to but it is not necessary 😃

PR #2297, opened 2023-06-20
Added task solution

ruslan-ihnatenko:


src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const finalResult = {};
+  const separator = /\s*(?:;|$)\s*/;
+
+  const tempResult1 = sourceString.split(separator)
+    .filter(element => element.length > 0);
+
+  for (const a of tempResult1) {
```
Anastasiia-Svintsova, line 15: Use the reduce method instead to iterate over the split elements and accumulate them into the result object.

src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
+    const tempElement = a.replace(/(\r\n|\n|\r)/gm, '').trim();
+    const indexOfDD = tempElement.indexOf(':');
```
Anastasiia-Svintsova, line 17: Just use the split method instead. It'll simplify the code. 
Split each non-empty element into key and value using the ':' delimiter, and trim whitespace
```
 const [key, value] = ...

Anastasiia-Svintsova (changes requested):

sTorba24 (approved):
Well done!

PR #2300, opened 2023-06-20
Solution

yehor-kor:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const lines = sourceString.split(';');
+  const convertedObject = {};
+
+  lines.forEach((item) => {
```
Anastasiia-Svintsova, line 20: It's better to use reduce method here to achieve the desired outcome in a more concise and efficient manner

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    const line = item.split(':');
+    const key = line[0].trim();
+    const value = line[1];
```
Anastasiia-Svintsova, line 23: Use destructuring instead 
```suggestion
    const [key, value] = item.split(':');
```

Anastasiia-Svintsova (changes requested):

sTorba24 (approved):
Well done!

PR #2303, opened 2023-06-20
Solution

melnyksergey7:


NadiaKovalchuk (changes requested):
really cool and understandable, but I think you need to rename your constants (s)

YegorKorenkov (approved):
Looks good!

PR #2308, opened 2023-06-20
Solution

eleonora-demikh:


AndriiMelnyk94 (approved):
Ok, good.

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .map(style => style.trim());
+
+  const result = {};
+
+  for (let i = 0; i < styles.length; i++) {
```
tetiana-ivanchenko, line 23: you can use reduce instead

tetiana-ivanchenko (approved):

PR #2312, opened 2023-06-20
Solution

NadiaKovalchuk:


src/convertToObject.js
```diff
@@ -11,8 +12,19 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceObj = sourceString.split(';')
+    .map(source => source.trim())
```
maksymPtashnyk, line 18: Checklist - 4 [CODE STYLE] - All chained methods called on the same level should have same indentation. This makes code much more readable.
(move `.split(';')` to a new line)

maksymPtashnyk (approved):
Good job, but check the comment)

tetiana-ivanchenko (changes requested):

src/convertToObject.js
ManoylovAC, line 18: It will be better to place all array operations on separate lines 
```suggestion
  const sourceObj = sourceString
    .split(';')
    .map(source => source.trim())
```

PR #2315, opened 2023-06-20
Solution

perfectAcedia:


Valentyna08 (approved):
good job!

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sortedArray = sourceString.split(';')
+    .map(element => trimStr(element))
+    .filter((word) => word !== '')
+    .reduce((styles, stringPart) => {
+      const trimmed = stringPart.split(':').map((part) => trimStr(part));
```
YegorKorenkov, line 21: Please destruct **trimmed** for better readability
```suggestion
      const [key, value]= stringPart.split(':').map((part) => trimStr(part));
```

YegorKorenkov (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sortedArray = sourceString.split(';')
```
ManoylovAC, line 17: Please be careful with names. You are not sorting this array, so the name should not be confusing. Also in result you have not array but object.

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+    .map(element => trimStr(element))
```
ManoylovAC, line 18: If you use several array functions in chain it will be better to place all of them on separate lines
```suggestion
  const sortedArray = sourceString
    .split(';')
    .map(element => trimStr(element))
```

src/convertToObject.js
ManoylovAC, line 17: since you have no additional variables and the function explains the whole point, you can immediately return the result
```suggestion
  return sourceString.split(';')
```

ManoylovAC (changes requested):
Good work but please be careful with clear naming - this is important part of coding

ManoylovAC (approved):
Good 👍

PR #2324, opened 2023-06-20
added task solution

vlopatko:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
-}
+  const resultObjectOfRules = {};
+  const arrOfRules = sourceString.split(';');
+  const arrOfFormatedRules = arrOfRules
+    .map(rule => rule
+      .split(':')
+      .map(item => item.replace('\n', '').trim()))
+    .filter(el => el.length > 1);
+
+  arrOfFormatedRules.forEach(rule => (
```
tetiana-ivanchenko, line 25:
```suggestion
  arrOfFormatedRules.forEach(([key, value]) => (
```

tetiana-ivanchenko (changes requested):

ivanbilotserkivskiy (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const resultObjectOfRules = {};
+  const arrOfRules = sourceString.split(';');
+  const arrOfFormatedRules = arrOfRules
+    .map(rule => rule
```
ManoylovAC, line 20: you can combine it if you want
```suggestion
  const arrOfFormatedRules = sourceString
    .split(';')
    .map(rule => rule
```

ManoylovAC (approved):
Good work 👍 
Check tips I left for you for the future

src/convertToObject.js
vlopatko: Oh, really... maybe I don't see this 😅 
Thanks for tips 🙏🏻

PR #2334, opened 2023-06-22
converToObject

zipic:


loralevitska (changes requested):
Almost perfect) Let's improve for more practice. Try to rewrite you solution with [split](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split) and [filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) array methods

SanyaBratashchuk (approved):

PR #2336, opened 2023-06-23
Solution

BeeGooDee:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  let pairs = sourceString.split(';');
+
+  pairs = pairs.map((x) => x.trim()).filter((x) => x.length > 0);
```
etojeDenys, line 21: Instead of using 'let' for the 'pairs' variable, you can chain the 'split()', 'map()', and 'filter()' methods directly and use 'const'.

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+
+  const toCss = function(x) {
```
etojeDenys, line 23: Consider using an arrow function for the 'toCss' function for consistency with the rest of the code.

etojeDenys (changes requested):

etojeDenys (approved):

PR #2338, opened 2023-06-24
Solution

Oleg-Lukanov:


src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
+    .map(char => char.trim())
+    .filter(char => char !== '');
+
+  const three = two.map(char => char.split(':'));
+
+  const threeTrim = three.map(char => {
+    return char.map(item => item.trim());
+  });
+
+  const four = threeTrim.reduce((prev, item) => {
```
alinatkachukkkk, line 29: Instead of using 'prev' and 'item' as parameter names in the reduce function, consider using more descriptive names like 'accumulator' and 'currentValue' or 'property' and 'value'.

src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
   // write your code here
+  const one = sourceString.split(';');
+  const two = one
+    .map(char => char.trim())
+    .filter(char => char !== '');
+
+  const three = two.map(char => char.split(':'));
+
+  const threeTrim = three.map(char => {
+    return char.map(item => item.trim());
```
alinatkachukkkk, line 26: Instead of using 'map' to trim the elements of each sub-array, consider using 'map' within the 'split' method on line 23. This will eliminate the need for the 'threeTrim' variable and make the code more concise.

src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const one = sourceString.split(';');
+  const two = one
+    .map(char => char.trim())
```
alinatkachukkkk, line 20: The 'trim' function is called on every style string twice, once on line 20 and once on line 26. To improve the efficiency of the code, consider removing one of these trim calls.

src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
+ * and values are the values of related
+ * CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const one = sourceString.split(';');
```
alinatkachukkkk, line 18: The variable names 'one', 'two', 'three', 'threeTrim', and 'four' are not descriptive. Consider using more meaningful names that describe the content or purpose of the variables.

src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
+  const two = one
```
alinatkachukkkk, line 19: Instead of using the 'char' variable name in the map and filter functions, consider using a more descriptive name like 'style' or 'property'.

PR #2342, opened 2023-06-27
solution

707Andrew:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArr = sourceString.split(';')
+    .map(style => style.trim())
+    .filter(style => style !== '');
+
+  const sourceObject = {};
```
SerhiiKirik, line 21: you can also try to do this with reduce

SerhiiKirik (approved):
Nice 🔥

PR #2344, opened 2023-06-28
convert To Object

NikolaienkoN:


SerhiiKirik (changes requested):
Well done 🔥 
But your algorithm is too complex, try to do it with 1 loop 👩‍💻

TarasHoliuk (approved):

PR #2346, opened 2023-07-01
add task solution

vetal-hovenko:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+    .split('\n')
+    .filter((value) => value.length > 3)
+    .map((value) => value.replace(/;/g, ''))
+    .map(value => value.trim())
+    .map((value) => value.split(':'))
+    .sort((value1, value2) =>
+      value1.toString().localeCompare(value2.toString()));
+
+  const objectOfCssValues = styleSheetsArray.reduce(
+    (obj, value) => {
```
volodymyr-soltys97, line 27: You can use destructuring directly in the parameter of your reduce function to make it cleaner. Like this: '(obj, [key, val]) => {...}'

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleSheetsArray = sourceString
+    .split('\n')
+    .filter((value) => value.length > 3)
+    .map((value) => value.replace(/;/g, ''))
+    .map(value => value.trim())
+    .map((value) => value.split(':'))
+    .sort((value1, value2) =>
```
volodymyr-soltys97, line 23: There's no need to sort the array, CSS rules order doesn't matter in an object representation.

volodymyr-soltys97 (approved):

PR #2347, opened 2023-07-01
add task style-to-object

P-Nazar:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    const [property, value] = style.split(':');
+
+    if (property && value) {
+      const propertyName = property.trim();
+      const propertyValue = value.trim();
+
+      result[propertyName] = propertyValue;
+    }
+  });
+
```
etojeDenys, line 30: we can rewrite it with `reduce`. it will improve readability

etojeDenys (approved):

PR #2349, opened 2023-07-03
Solution

ShamievDavid:


src/convertToObject.js
```diff
@@ -3,16 +3,44 @@
...
+  const stylesArray = sourceString.split('\n');
+  const trimmed = [];
+
+  stylesArray.forEach((style) => {
+    trimmed.push(style.trim().replace(';', ''));
+  });
+
+  const cleared = [];
+
+  trimmed.forEach((e) => {
```
mykhalenych, line 27: You are using a forEach loop to create a new array. A more efficient and readable way to do this would be using the map function.

src/convertToObject.js
```diff
@@ -3,16 +3,44 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString.split('\n');
```
mykhalenych, line 18: You are splitting the string by new line character. However, CSS properties may not always be on a new line. It would be more reliable to split by ';' as it is the character that separates CSS properties.

src/convertToObject.js
```diff
@@ -3,16 +3,44 @@
...
+  const trimmed = [];
+
+  stylesArray.forEach((style) => {
+    trimmed.push(style.trim().replace(';', ''));
```
mykhalenych, line 22: You are replacing only the first occurrence of ';' which might lead to incorrect results if a property value contains ';' within it. Instead of replace, you should use replaceAll.

src/convertToObject.js
```diff
@@ -3,16 +3,44 @@
...
+  });
+
+  const cleared = [];
+
+  trimmed.forEach((e) => {
+    if (e) {
+      cleared.push(e.split(':'));
```
mykhalenych, line 29: Instead of checking the truthiness of 'e', it would be better to specifically check if 'e' is not an empty string. This will avoid skipping properties which have falsy values like 0.

src/convertToObject.js
```diff
@@ -3,16 +3,44 @@
...
+    }
+  });
+
+  const final = cleared.reduce((res, [key, value]) => {
```
mykhalenych, line 33: In case value is undefined due to missing ':' in the input string, the function value.trim() will throw an error. You should add a check to ensure 'value' is not undefined before calling trim() on it.

mykhalenych (changes requested):

etojeDenys (approved):
lgtm

PR #2353, opened 2023-07-05
Solution

Luk2asz:


src/convertToObject.js
```diff
@@ -3,16 +3,54 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splitStrings = sourceString.split(';');
+
+  const splitLongStrings = splitStrings.filter(x => x.length > 4);
+
+  const stringsSplitedColon = [];
```
Radoslaw-Czerniawski, line 21: 1. Here you could use a [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) instead (get rid of the forEach):
```js
const propertiesAndKeysArr = filteredSplitStrings.flatMap(x => x.split(':'));
```

src/convertToObject.js
```diff
@@ -3,16 +3,54 @@
...
+
+  splitLongStrings.forEach(x => {
+    const arrayCopy = x.split(':');
+
+    stringsSplitedColon.push(arrayCopy[0]);
+    stringsSplitedColon.push(arrayCopy[1]);
+  });
+
+  const stringsWithoutSpaces = stringsSplitedColon.map(x => {
```
Radoslaw-Czerniawski, line 30: 2. Then this whole thing could be replace by [trim](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim):
```js
const trimmedPropertiesAndValuesArr = propertiesAndKeysArr.map(x => x.trim());
```

src/convertToObject.js
```diff
@@ -3,16 +3,54 @@
...
+    while (y.charAt(j - 1) === ' ') {
+      j--;
+    }
+
+    return y.slice(0, j);
+  });
+
+  const result = {};
+
+  for (let i = 0; i < stringsWithoutSpaces.length - 1; i += 2) {
```
Radoslaw-Czerniawski, line 49: 3. Now let's simplify this function some more going back to the step 1, we could use a regular map
```js
const propertiesAndKeysArr = trimmedPropertiesAndKeysArr.map(x => x.split(':'));
```
now we get this array of two arrays with two elements`[key, value]`
image

Radoslaw-Czerniawski: We run on this the map with trim, but we have to remember that there are two elements
```js
const trimmedPropertiesAndValuesArr = propertiesAndKeysArr.map(x => [x[0].trim(), x[1].trim()]);
```
now we get something like this:
image
now we just need to map this to an object

src/convertToObject.js
```diff
@@ -3,16 +3,54 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splitStrings = sourceString.split(';');
+
+  const splitLongStrings = splitStrings.filter(x => x.length > 4);
```
Radoslaw-Czerniawski, line 19: name could be changed to `filteredSplitStrings`

src/convertToObject.js
Radoslaw-Czerniawski: This will work and make the function return exact same result

src/convertToObject.js
Radoslaw-Czerniawski: However the syntax may be simplified a bit by destructuring each array in functions parameter:
```js
  const trimmedPropertiesAndValuesArr = propertiesAndKeysArr.map(([key,value]) => [key.trim(), value.trim()]);
```

Radoslaw-Czerniawski: And now instead of using the for loop, we could use [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) to reduce this array into one object and immediately return it:
```js
return trimmedPropertiesAndValuesArr.reduce(
  (accumulatedObject, [key, value]) => ({ ...accumulatedObject, [key]: value }),
  {}
);
```

Radoslaw-Czerniawski: Overall function now will look like this:
image

Radoslaw-Czerniawski: This is somewhat optimal, however still not enough, we could actually take care of this in single reduce:
image

Radoslaw-Czerniawski: And go even further, use the arrow function's implicit return and do this without functions body or declaring any additional variables in it:
image

Radoslaw-Czerniawski (changes requested):
Follow this refactor step by step, console.logging it's effects, understand what is happening and then implement

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+ * exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertToObject = sourceString => 
+  sourceString.split(';').reduce((acc, nextEL) {
```
DorotaLeniecDev, line 18: Bad formatting, callback arrow function needs =>
```suggestion
  sourceString.split(';').reduce((acc, nextEL) => {
```

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+ * and values are the values of related CSS properties (see an
+ * exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertToObject = sourceString =>
```
DorotaLeniecDev, line 17: Your function is already declared on line 14

DorotaLeniecDev (changes requested):
Fix code formatting and function declaration

DorotaLeniecDev (approved):
Good job

PR #2354, opened 2023-07-05
add task solution

NikitaOrliak:


src/convertToObject.js
```diff
@@ -3,16 +3,41 @@
...
+  const splittedTrim = splittedKeysAndValues
+    .filter((arr) => arr.length > 1)
+    .map((keyAndValue) =>
+      [
+        keyAndValue[0].trim(),
+        keyAndValue[1].trim(),
+      ]
+    );
+
+  for (const arr of splittedTrim) {
```
polosanya, line 34: Instead of creating a new array 'splittedTrim' and then iterating over it again, you could simply add each key-value pair to 'keysAndValues' inside the first loop. This would make the code more efficient.

polosanya (approved):
Good job!

PR #2357, opened 2023-07-07
solution

tndeineka:


src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObj = {};
+
+  const styles = sourceString.split(';');
+
+  for (let i = 0; i < styles.length; i++) {
+    const style = styles[i];
```
VitaliyBodnar, line 9: there is only one use of index 'i', so better to use here the loop for of

VitaliyBodnar (changes requested):
Well done 🔥
Let's improve your solution 🧑‍💻

VitaliyBodnar (approved):

PR #2360, opened 2023-07-08
add solution

denis-kononenko:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+  const result = {};
+
+  const properties = sourceString.trim().split(/\s*;\s*/);
+
+  for (const item of properties) {
+    if (item === '') {
+      continue;
+    }
+
+    const propertyAndValue = item.split(/\s*:\s*/);
```
volodymyr-soltys97, line 26: The regex '/\s*:\s*/' can fail if there is no space between property and value or if there are multiple spaces. Consider using a simpler split like '.split(':')' and then trim each resulting string.

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+    result[propertyAndValue[0]] = propertyAndValue[1];
```
volodymyr-soltys97, line 28: Consider checking if the 'propertyAndValue' array has two items before assigning to avoid undefined values.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

nazarmatsevych (approved):

PR #2362, opened 2023-07-09
Solution

Hladkyi03:


src/convertToObject.js
```diff
@@ -12,7 +13,20 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+
+  sourceString
+    .replace(/\s+/g, ' ')
+    .split(';')
+    .map(propertie => propertie.trim())
+    .filter(propertie => propertie)
```
TarasHoliuk, line 22:
```suggestion
    .filter(Boolean)
```
result: `Boolean(property)` returns boolean value. You will learn about classes/constructors later, so it's just FYI

src/convertToObject.js
```diff
@@ -12,7 +13,20 @@
...
+    .map(propertie => {
```
TarasHoliuk, line 23: why `map`? you don't save result array. to just iterate an array there is another method

TarasHoliuk (changes requested):

VitaliyBodnar (approved):
Well done )

PR #2365, opened 2023-07-10
convertor to obj

dimkamg21:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    let [key, value] = property.split(':');
+
+    key = key.replace(/\n| /g, '');
+    value && (value = value.trim());
+
+    if (key === '') {
+      return;
+    }
+
+    return { [key]: value };
```
demetrxx, line 28: You return undefined or object in a map. This will result in smth like: 
`[undefined, { color: 'red' }, undefined, { border: '1px solid' }]`

consider following common solution, and improve your code here:

```
const styleObject = {};
  
    .split(';')
    .map((elem) => elem.trim())
    .filter((elem) => elem.length > 0)
    .map((elem) => elem.split(':'))
    .map((elem) => elem.map((n) => n.trim()));
  properties.forEach(([property, value]) => (styleObject[property] = value));
  return styleObject;
```

demetrxx (changes requested):
Good job! But let's improve your code a bit.

Viktor-Kost (changes requested):
Hello!)
Please apply [recommendation](https://github.com/mate-academy/js_style-to-object/pull/2365#discussion_r1259547585) from previous mentor 
If you need any help - feel free to ask at fe_chat

etojeDenys (approved):
GJ

PR #2367, opened 2023-07-11
Solution

mukutiuk:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+      .split(':'))
+    .filter((item) => item.length !== 1);
+
+  const association = newArry.map((item) => item.map((elem) => elem.trim()));
+
+  association.map(function(item) {
+    obj[item[0]] = item[1];
+  });
+
+  return obj;
```
demetrxx, line 30:
```suggestion
  const styleObject = {};
  
  const properties = sourceString
    .split(';')
    .map((elem) => elem.trim())
    .filter((elem) => elem.length > 0)
    .map((elem) => elem.split(':'))
    .map((elem) => elem.map((n) => n.trim()));

  properties.forEach(([property, value]) => (styleObject[property] = value));

  return styleObject;
```

demetrxx (changes requested):
Good job! But let's improve your code a bit.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
 /**
  * Implement convertToObject function:
  *
- * Function takes a string with styles (see an example in [stylesString.js](./stylesString.js))
+ * Function takes a string with styles (see an example in
+ *  [stylesString.js](./stylesString.js))
```
OleziO, line 7: Remove all comments form the code.

OleziO (changes requested):

demetrxx (approved):
Well done 👍 Go ahead!

PR #2368, opened 2023-07-12
Develop

AsyaYeromina:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const resultObj = {};
+  const clearProperties = sourceString
+    .replaceAll('\n', '')
+    .split(';')
+    .map(
+      rule => rule.split(':'))
+    .filter(
+      rule => rule.length === 2);
```
Andriy-Shymkiv, line 25:
```suggestion
  const clearProperties = sourceString
    .replaceAll('\n', '')
    .split(';')
    .map(rule => rule.split(':'))
    .filter(rule => rule.length === 2);
```

Andriy-Shymkiv: don't understand why you need to filter the array by `length === 2`? mayble let's try `length > 0`?

Andriy-Shymkiv (approved):

src/convertToObject.js
AsyaYeromina: I used it to filter arrays with empty strings/strings with spaces. 
.filter(([key, value]) => key != false); also can be used for that, but I didn't want to use this equality operator because of linter. 

Please give some advice for better solving.

PR #2378, opened 2023-07-18
solution

NataliMax:


src/convertToObject.js
```diff
@@ -1,18 +1,29 @@
...
+      const [ a, b ] = arr;
+
+      return [a.trim(), b.trim()];
+    })
+    .reduce((prev, item) => {
+      return {
+        ...prev,
+        [item[0]]: item[1],
+      };
+    }, {});
```
IvanFesenko, line 24: you can make it more shorter)
```suggestion
    .map(([ a, b ]) => [a.trim(), b.trim()])
    .reduce((prev, [key, value]) => ({
        ...prev,
        [key]: value,
      }), {});
```

IvanFesenko (approved):
Well done

PR #2385, opened 2023-07-21
add task solution

PavloMykhalov:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
   // write your code here
+  const styles = {};
+  const properties = sourceString.split(';\n');
+
+  properties.forEach(property => {
+    const [name, value] = property
+      .split(':')
+      .map(part => part.trim());
+
+    if (name !== '') {
```
Anastasiia-Svintsova, line 26: The if-statement checks if the name is not an empty string, but it does not check if the value is not undefined. If a style property has no value, this could introduce undefined values into your styles object.

Anastasiia-Svintsova (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const styles = {};
+  const properties = sourceString.split(';\n');
```
anastasiiavorobiova, line 19: The split method is splitting the string by ';\n', but there might be cases where the properties are separated only by ';' and the CSS code would be valid even in this case

anastasiiavorobiova (changes requested):
Almost done! Added a tiny comment to fix

volodymyr-soltys97 (approved):
Well done 👍

PR #2390, opened 2023-07-22
first commit

Faridun92:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+ * in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const newArr = sourceString.split('\n')
+    .filter((element) => element.length >= 5)
```
SanyaBratashchuk, line 18:
```suggestion
    .filter((element) => element.length >= 5)
```
Why exactly 5? It doesn't seem flexible

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    .map((element) => element.replace(';', ''));
+  const styleObj = {};
+
+  newArr.forEach((element) => {
+    const [ prop, val ] = element.split(':');
+
+    styleObj[prop.trim()] = val.trim();
+  });
+
+  return styleObj;
```
SanyaBratashchuk, line 28: use reduce

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #2395, opened 2023-07-29
Convert styles to objects, solution

AndrzejPec:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const contentTrimmedOnTheOutside = sourceString
+    .trim()
+    .split(';')
+    .map(item => item.trim())
+    .filter(Boolean);
+
+  const styles = contentTrimmedOnTheOutside.reduce((acc, item) => {
+    const [property, value] = item.split(':').map(s => s.trim());
```
choeqq, line 25: I think key, value would be more explicit
```suggestion
    const [key, value] = item.split(':').map(s => s.trim());
```

choeqq (approved):
Well done ✅

PR #2397, opened 2023-07-30
solution

Sancz0pansa:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  let current;
+  const properties = sourceString.split(';');
+
+  for (let i = 0; i < properties.length; i++) {
+    current = properties[i].split(':');
```
choeqq, line 22: you can destructure key and value from properties
```suggestion
    const [key, value] = properties[i].split(':');
```

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+ *  (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  let current;
```
choeqq, line 18: if you destructure key and value from properties this can be omited

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+
+  for (let i = 0; i < properties.length; i++) {
+    current = properties[i].split(':');
+
+    if (!current[1]) {
+      continue;
+    }
+    current[0] = current[0].trim();
+    current[1] = current[1].trim();
+    obj[current[0]] = current[1];
```
choeqq, line 29: you can just set trimmed values here
```suggestion
    obj[key.trim()] = current[value.trim()];
```

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+  }
```
choeqq, line 30: you can use reduce method to loop through properties, then you won't have to initialize new obj

choeqq (changes requested):
Almost there 👍

mvjl000 (changes requested):
Code looks great! Please remove old commented code. And also I would rename the `obj` constant.

asakevych (approved):
Great job! 🚀 🔥 👏

PR #2399, opened 2023-07-30
Solution

Patryk-Filipiak:


src/convertToObject.js
```diff
@@ -3,16 +3,23 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+
+  sourceString.split(';').forEach(stringProperty => {
+    const [prop, value] = stringProperty.split(':');
```
choeqq, line 15: I think key would be more exlicit
```suggestion
    const [key, value] = stringProperty.split(':');
```

choeqq (approved):
Good job ✅

PR #2400, opened 2023-07-30
Develop

anpiet:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+    if (styles[i].length === 0) {
+    }
+  });
+  styles = styles.filter(el => el.length > 0).map(el => el.split(':'));
+
+  const result = {};
+
+  styles.forEach(el => {
+    result[el[0].trim()] = el[1].trim();
+  });
```
choeqq, line 31: `el` could be destructured:
```suggestion
  styles.forEach(el => {
    const [key, value] = el;
    result[key.trim()] = value.trim();
  });
```

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let styles = sourceString.replace(/\n*/g, '').split(';');
+
+  styles.forEach((el, i) => {
+    styles[i] = el.trim();
+
+    if (styles[i].length === 0) {
+    }
```
choeqq, line 23: that's an empty if statement, you can remove that

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+  });
```
choeqq, line 24: can you provide explanation to this loop? I can't wrap my head around it 🤔

choeqq (changes requested):
Few adjustments needed, good job ✅

src/convertToObject.js
anpiet: yes, thank you. good point!

src/convertToObject.js
anpiet: a trimmed all lines, to make empty lines with 0 length (to filter it in next line). It's not necessary - I have added trim to filter in next line and it also works

asakevych (approved):
Great job!! 🚀 🔥 👏

PR #2401, opened 2023-07-30
add task solution

ZakZh:


src/convertToObject.js
```diff
@@ -1,18 +1,15 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';')
+    .map((styleStr) => styleStr.split(':').map(val => val.trim()))
+    .filter(val => val.length === 2)
+    .reduce((prev, next) => (
+      {
+        ...prev,
+        [next[0]]: next[1],
+      }
```
VitaliyBondarenko1982, line 11: looks good, but let''s improve solution.

map and filter loops are redundant.
try to make everything in reduce method.

src/convertToObject.js
```diff
@@ -1,18 +1,15 @@
...
- */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';')
+    .map((styleStr) => styleStr.split(':').map(val => val.trim()))
+    .filter(val => val.length === 2)
+    .reduce((prev, next) => (
+      {
+        ...prev,
+        [next[0]]: next[1],
```
VitaliyBondarenko1982, line 10: use destructuring for next
```
const [key, value] = next
```

VitaliyBondarenko1982 (changes requested):

Viktor-Kost (approved):

PR #2402, opened 2023-07-31
add task

VladyslavKolisnyk:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const propertiesObject = {};
+
+  sourceString
+    .split(';')
+    .map(line => line.split(':').map(element => element.trim()))
```
volodymyr-soltys97, line 21: Calling map() method but not using its result. If you just need to iterate over an array and perform side effects, use forEach(). If you want to transform an array into a new array, use map() and assign its result to a variable or return it.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    .forEach(propertyPair => {
+      if (propertyPair[0]) {
```
volodymyr-soltys97, line 23: You are checking if propertyPair[0] exists before adding it to the propertiesObject. This will add undefined as a key if propertyPair[1] is not defined. Consider checking both propertyPair[0] and propertyPair[1] before adding them to the object to prevent this.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

anastasiiavorobiova (approved):
Great work!

PR #2409, opened 2023-08-02
Style to object

Kacper-Leman:
Soultion.

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -19,9 +19,7 @@ function convertToObject(sourceString) {
       style.trim() !== '');
 
   const stylesObject = stylesArray.reduce(function(result, style) {
-    const parts = style.split(':').map(part => part.trim());
-    const property = parts[0];
-    const value = parts[1];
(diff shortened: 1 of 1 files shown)


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString
+    .split(';').filter(style =>
+      style.trim() !== '');
+
+  const stylesObject = stylesArray.reduce(function(result, style) {
+    const parts = style.split(':').map(part => part.trim());
```
darokrk, line 22: you can destruct property from part: `const [property, value] = style.split(':').map(part => part.trim());`

and then you can remove redundant:

```
const property = parts[0];
const value = parts[1];
```

darokrk (changes requested):
Good job 👍🏽   one improvement left and we good to go 💪🏽

EwaKobiela (approved):
Good job!

darokrk (approved):
Good job 👏🏽

PR #2411, opened 2023-08-02
task solution

HubertKlamann:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectOfStyles = {};
+  const arrOfStyles = sourceString.split(';');
+
+  for (const style of arrOfStyles) {
+    const [keyOfObject, ...value] = style.trim().split(':')
```
choeqq, line 21: can't you just go with value?
```suggestion
    const [keyOfObject, value] = style.trim().split(':')
```

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+      .map(item => item.trim().split(' ').filter(v => v).join(' '));
+
+    if (keyOfObject && value) {
+      objectOfStyles[keyOfObject] = value.join(' ');
+    }
```
choeqq, line 26: that's the perfect usecase for reduce

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectOfStyles = {};
+  const arrOfStyles = sourceString.split(';');
```
choeqq, line 18: you can also trim it, so you won't have to do it in next lines

choeqq (changes requested):
Well done, I think you can improve your solution 🚀

choeqq (changes requested):
Waiting for any updates @HubertKlamann 👀

PR #2413, opened 2023-08-02
Develop

Matylda-Lewandowska:
Task solution - style-to-object

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -3,16 +3,34 @@
 /**
  * Implement convertToObject function:
  *
- * Function takes a string with styles (see an example in [stylesString.js](./stylesString.js))
+ * Function takes a string with styles
+ * (see an example in [stylesString.js](./stylesString.js))
  * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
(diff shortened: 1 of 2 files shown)


domalewski (approved):
looks awesome 🔥 🚀 🌔

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  for (const array of arrayWithinArray) {
+    const rawKey = array[0].split('\n');
+    const filteredKey = rawKey.filter(v => v.length > 0);
+    const finaleKey = filteredKey.join(' ').trim();
+
+    if (array[1] !== undefined && finaleKey.length > 0) {
+      result[finaleKey] = array[1].trim();
+    }
+  }
```
choeqq, line 29: that's the perfect case for using reduce

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stringToArray = sourceString.split(';');
+  const arrayWithinArray = stringToArray.map(v => v.split(':'));
+  const result = {};
+
+  for (const array of arrayWithinArray) {
+    const rawKey = array[0].split('\n');
```
choeqq, line 22: you can destructure both key and value from array 
```suggestion
    const [key, value]= array;
```

choeqq, line 22: please consider trimming array
```suggestion
    const rawKey = array[0].trim();
```

choeqq (changes requested):
Looks good, but I think you can do little better 🚀

Matylda-Lewandowska:
I've made changes and I hope that it's good now :D

choeqq (approved):

PR #2414, opened 2023-08-02
Solution

szymonpachucki:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+      i--;
+    }
+  }
+  styles = styles.map(style => style.split(':'));
+
+  const result = {};
+
+  for (const style of styles) {
+    result[style[0].trim()] = style[1].trim();
+  }
```
choeqq, line 33: you can destructure key and value from style
```suggestion
  for (const style of styles) {
    const [key, value] = style;
    result[key.trim()] = value.trim();
  }
```

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+
```
choeqq, line 34: that's the perfect usecase for reduce method 👍

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+  let styles = sourceString.replace(/\n*/g, '').split(';');
+
+  for (let i = 0; i < styles.length; i++) {
+    styles[i] = styles[i].trim();
+
+    if (styles[i].length === 0) {
+      styles.splice(i, 1);
+      i--;
+    }
+  }
```
choeqq, line 26: you can also move that logic into map down below, then you will have only 1 loop

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  let styles = sourceString.replace(/\n*/g, '').split(';');
+
+  for (let i = 0; i < styles.length; i++) {
+    styles[i] = styles[i].trim();
+
+    if (styles[i].length === 0) {
+      styles.splice(i, 1);
+      i--;
```
choeqq, line 24: you can also use filter method for that

choeqq (changes requested):
Almost there 🚀

choeqq (approved):

PR #2415, opened 2023-08-02
Solution

jakubMikolajczykTech:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  for (const array of arrayWithinArray) {
+    const rawKey = array[0].split('\n');
+    const filteredKey = rawKey.filter(v => v.length > 0);
+    const finaleKey = filteredKey.join(' ').trim();
+
+    if (array[1] !== undefined && finaleKey.length > 0) {
+      result[finaleKey] = array[1].trim();
+    }
+  }
```
choeqq, line 29: you can use reduce for that

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stringToArray = sourceString.split(';');
+  const arrayWithinArray = stringToArray.map(v => v.split(':'));
```
choeqq, line 18: maybe trim you array here

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+  const result = {};
+
+  for (const array of arrayWithinArray) {
+    const rawKey = array[0].split('\n');
```
choeqq, line 22: you can desctructure both key and value from array
```suggestion
    const [key, value] = array.split('\n');
```

choeqq (changes requested):

src/convertToObject.js
choeqq, line 29: that's the perfect usecase for reduce

choeqq (changes requested):
Waiting for any updates @kabakTech 👀

choeqq (approved):

PR #2416, opened 2023-08-02
Develop

Dolnys:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+
+  const styleLine = singleLineString.split(';');
+
+  for (const style of styleLine) {
+    const [key, value] = style.split(':');
+
+    if (key && value) {
+      result[key.trim()] = value.trim();
+    }
+  }
```
choeqq, line 30: I think you can use reduce for that

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  const singleLineString = sourceString.replace(/\n/g, '');
```
choeqq, line 20: nice regexp 👍

choeqq (changes requested):
Looks good, please take a look at my commnet ⬇️

choeqq (approved):

PR #2419, opened 2023-08-02
add task solution

kuzgan:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
-  // write your code here
+  const styleObj = {};
+
+  sourceString.split(';').forEach((prop) => {
+    const [key, value] = prop.split(':');
+
+    if (value !== undefined) {
+      styleObj[key.trim()] = value.trim();
+    }
+  });
```
choeqq, line 25: that's the perfect usecase for reduce method

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObj = {};
+
+  sourceString.split(';').forEach((prop) => {
+    const [key, value] = prop.split(':');
+
+    if (value !== undefined) {
```
choeqq, line 22: since you're checking if value is defined, you can also check if key is defined 😄

choeqq (changes requested):
Looks good and concise, I think you can improve it 🚀

choeqq (approved):

PR #2420, opened 2023-08-02
solution

PawelP2001:


src/convertToObject.js
```diff
@@ -1,18 +1,28 @@
...
+    const [key, value] = line.split(':').map(item => item.trim());
+
+    if (key && value) {
+      const processedValue = value.endsWith(';')
+        ? value.slice(0, -1).trim()
+        : value;
+
+      resultCss[key] = processedValue;
+    }
+  });
```
choeqq, line 23: you can use reduce for that

src/convertToObject.js
```diff
@@ -1,18 +1,28 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const styleLines = sourceString.trim().split('\n');
```
choeqq, line 10: I think when you'll use trim() there will be no new lines (\n), maybe try to split it with `;`, then split will return you style lines
```suggestion
  const styleLines = sourceString.trim().split(';');
```

choeqq (changes requested):
Almost there, you can improve your solution 👍

PR #2421, opened 2023-08-02
Solution

Vira-V:


src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
+
+  sourceString
+    .split(';')
+    .map(item => item.trim())
+    .filter(item => item.length > 0)
+    .forEach(item => {
+      const [key, value] = item.split(':').map(v => v.trim());
+
+      stylesInObject[key] = value;
+    });
```
choeqq, line 20: that's the perfect usecase for reduce, then you can omit stylesInObject object 😄

choeqq (changes requested):
Looks great, if you'll implement reduce method, it will be perfect 🚀

choeqq (approved):

PR #2422, opened 2023-08-02
add convert to obj function

Meg-Sowka:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+  const stylesList = sourceString.split(';').filter(v => v.trim());
+  const stylesObj = {};
+
+  stylesList.map(style => {
+    const newStyle = style.split(':');
+
+    stylesObj[newStyle[0].trim()] = newStyle[1].trim();
+
+    return newStyle;
+  });
```
choeqq, line 26: that's the perfect usecase for reduce function

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesList = sourceString.split(';').filter(v => v.trim());
+  const stylesObj = {};
+
+  stylesList.map(style => {
+    const newStyle = style.split(':');
```
choeqq, line 21: you can descturcutre key and value from style
```suggestion
    const [key, value] = style.split(':');
```

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesList = sourceString.split(';').filter(v => v.trim());
+  const stylesObj = {};
+
+  stylesList.map(style => {
```
choeqq, line 20: please create new value for mapping result, and then return it

choeqq (changes requested):
Looks really good, but you can improve a little bit 🚀

asakevych (approved):
Greate job!! 🚀  🔥 👏

PR #2425, opened 2023-08-03
Solution

Anmonnn:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const array = sourceString.trim().split(';');
+  const obj = {};
+
+  for (let i = 0; i < array.length; i++) {
+    const arrayIn = array[i].trim().split(':');
```
anastasiiavorobiova, line 22:
```suggestion
    const arrayIn = array[i].trim().split(':').map(el => el.trim())
```

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+
+    if (arrayIn[0] !== '') {
+      const key = arrayIn[0].trim();
+      const value = arrayIn[1].trim();
+
```
anastasiiavorobiova, line 27: Here you can delete trimming

anastasiiavorobiova (approved):
Awesome!

PR #2430, opened 2023-08-03
Solution

emil-owczarek:


src/convertToObject.js
```diff
@@ -2,17 +2,24 @@
...
+  const styles = sourceString.split(';');
+  const stylesObject = {};
 
+  for (const style of styles) {
+    const [property, value] = style.split(':');
+
+    if (property && value) {
+      stylesObject[property.trim()] = value.trim();
+    }
+  }
```
choeqq, line 21: you can also use reduce for that 😄

choeqq (changes requested):

choeqq (approved):

PR #2432, opened 2023-08-04
task solution

Nightycake:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+  const styles = sourceString.split(';');
+  const stylesObj = {};
+
+  styles.forEach((style) => {
+    const [key, value] = style.split(':');
+
+    if (key && value) {
+      stylesObj[key.trim()] = value.trim();
+    }
+  });
```
choeqq, line 26: you can also use reduce for that

choeqq (changes requested):

PR #2433, opened 2023-08-04
test

WaveOfDandelions:


src/convertToObject.js
```diff
@@ -3,16 +3,22 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .trim()
+    .split(';')
+    .map((item) => item.trim().split(':'))
+    .filter((item) => item.length > 1)
+    .reduce(function(prev, item) {
+      prev[item[0].trim()] = item[1].trim();
```
choeqq, line 18: you can desctructure both key and value from item

choeqq (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,22 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .trim()
+    .split(';')
+    .map((item) => item.trim().split(':'))
+    .filter((item) => item.length > 1)
+    .reduce(function(prev, item) {
```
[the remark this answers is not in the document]
WaveOfDandelions: fixed

src/convertToObject.js
WaveOfDandelions: > you can desctructure both key and value from item

Can you provide an example?

JoyCoffeeAddict:
```js 
const [key, value] = item
```

asakevych (approved):
Great job!! 🚀 🔥 👏

PR #2435, opened 2023-08-04
solution

vensecue:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .map(item => item.trim())
+    .filter(item => item.length > 0)
```
AndrzejPec, line 20: I think
```suggestion
    .filter(item => item)
```
or
```suggestion
    .filter(Boolean)
```
would work aswell :)

AndrzejPec (commented):
Not bad, not bad :)

choeqq (approved):

PR #2438, opened 2023-08-05
style-to-object_solution

olha-rypich:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+      (prev, entry, i) => {
+        const key = entry.slice(0, entry.indexOf(':'));
+        const value = entry.slice(entry.indexOf(':') + 1);
+
+        return {
+          ...prev, [key.trim()]: value.trim(),
+        };
+      }, {});
+
+  return object;
```
VitaliyBondarenko1982, line 30: good job.
just want to show little another solution.
filtering looks like redundant and possible make all in reduce.
```suggestion
  return allEntries
    .reduce(
      (prev, entry, i) => {
        const [key, value] = entry.split(':').map(item => item.trim());

        if (!key) {
          return prev;
        }

        return {
          ...prev, [key]: value,
        };
      }, {});
```

VitaliyBondarenko1982 (approved):

PR #2439, opened 2023-08-06
js_style-to-object

MaksymShevch:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const allEntries = sourceString.split(';');
+  const objectEntries = allEntries.filter(entry => entry.trim());
+  const object = objectEntries
+    .reduce(
+      (prev, entry, i) => {
+        const key = entry.slice(0, entry.indexOf(':'));
```
nazarmatsevych, line 22: You can try to use split instead of slice

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+        const value = entry.slice(entry.indexOf(':') + 1);
```
nazarmatsevych, line 23: same here, use split instead of slice and indexOf

nazarmatsevych (approved):

PR #2440, opened 2023-08-06
solution

MaciejKujawa:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const obj = {};
+  const str = sourceString;
```
choeqq, line 18: I don't think you have to reassign sourceString, we aren't modifying it directly in this task

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+  for (let i = 0; i < strArr.length; i++) {
+    if (strArr[i].includes(':')) {
+      const keyValue = strArr[i].split(':');
+
+      const forKey = keyValue[0].trim();
+      const forValue = keyValue[1].trim();
 
+      obj[forKey] = forValue;
+    }
+  }
```
choeqq, line 31: that's the perfect usecase for reduce

choeqq (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+  const str = sourceString;
+  const stringWithNoBreakLInes = str.replace(/\n*/g, '');
+  const strArr = stringWithNoBreakLInes.split(';');
+
+  for (let i = 0; i < strArr.length; i++) {
+    if (strArr[i].includes(':')) {
+      const keyValue = strArr[i].split(':');
+
+      const forKey = keyValue[0].trim();
+      const forValue = keyValue[1].trim();
```
choeqq, line 27: you can desctructure those from keyValue
```suggestion
      const [key, value] = keyValue;
```

choeqq (approved):

PR #2441, opened 2023-08-06
feat: add task solution

FrontKid:


Wesses (approved):
Well done

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @return {object}
  */
-function convertToObject(sourceString) {
-  // write your code here
-}
+const convertToObject = sourceString =>
+  sourceString
+    .split('\n')
+    .map(rule => rule.trim()
+      .replace(/[\t;]/g, ''))
```
33Orange, line 20: You are replacing tab and semicolon characters with a space. But semicolon (;) is used to separate style rules, so removing it this way might cause problems. You should split by semicolon first, then trim and handle each rule separately.

33Orange, line 20: `.map((style) => style.split(':').map((value) => value.trim()))`

roman-mirzoian (approved):
Well done 👍

PR #2442, opened 2023-08-06
add task solution

MateuszSlezak:


src/convertToObject.js
```diff
@@ -1,18 +1,18 @@
...
+  const cssObject = {};
+  const filteredItems = sourceString.split(';').filter(x => x.includes(':'));
+
+  filteredItems.forEach((element) => {
+    const part = element.split(':');
+    const trimKey = part[0].trim();
+    const trimValue = part[1].trim();
+
+    cssObject[trimKey] = trimValue;
+  });
```
choeqq, line 13: that's the perfect usecase for reduce

src/convertToObject.js
```diff
@@ -1,18 +1,18 @@
...
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssObject = {};
+  const filteredItems = sourceString.split(';').filter(x => x.includes(':'));
+
+  filteredItems.forEach((element) => {
+    const part = element.split(':');
+    const trimKey = part[0].trim();
+    const trimValue = part[1].trim();
```
choeqq, line 10: you can destructure those from part
```suggestion
    const [key, value] = part
```

choeqq (changes requested):

choeqq (approved):

PR #2443, opened 2023-08-07
Style to object done

roman-tarasov90:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  let prop = [];
+  const strToArr = sourceString.split(';').map(key => key.trim());
+
+  for (let i = 0; i < strToArr.length; i++) {
```
VitaliyBondarenko1982, line 22: consider to use reduce method instesd for loop

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    prop = strToArr[i].split(':').map(key => key.trim());
```
VitaliyBondarenko1982, line 23:
```suggestion
    const [key, value] = strToArr[i].split(':').map(key => key.trim());
```
destructure array instead read by index

VitaliyBondarenko1982 (changes requested):
Good job.
Let's improve solution.

IvanFesenko (approved):
well done

PR #2445, opened 2023-08-07
add task solution

Wesses:


src/convertToObject.js
```diff
@@ -1,18 +1,33 @@
...
+    .split(';')
+    .map(style => style.trim())
+    .filter(style => style !== '');
+
+  const objectOfStyles = arrayOfStyles.reduce((obj, style) =>
+    (
+      {
+        ...obj, [style.split(':')[0].trim()]: style.split(':')[1].trim(),
+      }
+    ), {});
```
DenisChakhalian, line 28: you can create variables for
`style.split(':')[0].trim()` and `style.split(':')[1].trim()`
to make it more visually beautiful

and also move to a new line after `...obj,`

DenisChakhalian (approved):
good job!

src/convertToObject.js
33Orange, line 28: Let's use destructuring assignment
Think about readability

```suggestion
  const objectOfStyles = arrayOfStyles.reduce((obj, style) => {
      const [key, value] = style.split(':');
      
      return {
        ...obj, 
        [key.trim()]: value.trim(),
      }
    }, {});
```

33Orange (changes requested):
Almost done

src/convertToObject.js
```diff
@@ -1,18 +1,35 @@
...
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayOfStyles = sourceString
+    .split(';')
+    .map(style => style.trim())
+    .filter(style => style !== '');
```
dffUqp, line 21: we go through array 2 times, let's remove this `map` and `filter` 

I think we can make check inside `reduce`. 

If our value is undefined we skip iteration

dffUqp (changes requested):
Almost done:smiley:

src/convertToObject.js
```diff
@@ -1,18 +1,42 @@
...
+    .split(';')
+    .reduce((newArr, style) => {
+      if (style.trim() === '') {
+        return newArr;
+      }
+
+      newArr.push(style.trim());
+
+      return newArr;
+    }, []);
```
roman-mirzoian, line 28: Transfer this logic to the second `reduce`.

roman-mirzoian (changes requested):

src/convertToObject.js
```diff
@@ -1,18 +1,34 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .reduce((obj, style) => {
+      if (style.trim() === '') {
+        return obj;
+      }
+
+      const [key, value] = style.trim().split(':');
```
roman-mirzoian, line 25: You're using `style.trim()` twice, so it could be moved to a separate variable.

src/convertToObject.js
```diff
@@ -1,18 +1,34 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .reduce((obj, style) => {
```
roman-mirzoian, line 20: Try to choose more meaningful names.
```suggestion
    .reduce((styleObject, currentStyle) => {
```

roman-mirzoian (approved):
Well done, but pay attention to the comments.

PR #2446, opened 2023-08-07
solution

DenisChakhalian:


Koliras (approved):
Good job

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+ * and values are the values of related CSS properties (see an exampl in
+ * [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceObject = sourceString.trim().split(';')
```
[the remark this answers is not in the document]
BudnikOleksii: Good point!
image

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    .map(property => property.trim().split(':'))
+    .filter(property => property[0] !== '')
```
```suggestion
    .filter(property => property[0])
```
If you need to check if the value exists, you can just return it in the filter.

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    .reduce((properties, property) => {
+      const keyProperty = property[0].trim();
+      const valueProperty = property[1].trim();
+
+      return {
+        ...properties,
+        [keyProperty]: valueProperty,
+      };
```
BudnikOleksii, line 27:
```suggestion
    .reduce((properties, [keyProperty,  valueProperty]) => {
      return {
        ...properties,
        [keyProperty.trim()]: valueProperty.trim(),
      };
```
Also, you can use destructuring, something like this

BudnikOleksii (changes requested):
Nicely done!
Left some comments for improvements.

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceObject = sourceString
+    .trim()
+    .split(';')
+    .map(property => property
+      .trim()
+      .split(':'))
+    .filter(property => property[0])
```
dffUqp, line 23: we go through array 3 times, let's remove this `map` and `filter` 

I think we can call `split(":")` in our reduce and get `key` and `value`

dffUqp (changes requested):
Almost done:smiley:

PR #2447, opened 2023-08-07
add task solution

Koliras:


davinnchii (approved):
LGTM!!
Gj, also fyk you could fill your object with forEach method, as for me it look a bit simpler

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectifiedString = sourceString
+    .split(';')
+    .map(prop => prop.trim().split(':'))
+    .filter(prop => prop[0] !== '')
```
```suggestion
    .filter(prop => prop[0])
```
If you need to check if the value exists, you can just return it in the filter.

BudnikOleksii (approved):
Good job!

PR #2449, opened 2023-08-08
convert to object solution

Denys14Samoilenko:


voievodik (changes requested):
Nice! Good job!

Denys14Samoilenko:
> Nice! Good job!
thanks) 
but I got a reject from you) I understand that you did it by accident? :)

voievodik (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayToStyles = sourceString
+    .trim()
+    .split(';')
+    .filter(styleProp => styleProp.includes(':'))
```
dffUqp, line 20: I think we can remove this `.filter` and just check inside out `reduce` if our value is not `undefined`

dffUqp (changes requested):
Almost done:smiley:

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
```
33Orange, line 20: Check @dffUqp comment please

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+    .reduce((stylesList, stylesParameters) => {
+      const [property, value] = stylesParameters.split(':');
+
+      if (property !== undefined && value !== undefined) {
+        return {
+          ...stylesList,
+          [property.trim()]: value.trim(),
+        };
```
33Orange, line 28: It would be better to not create a new object and iterate through the styles list when using spread every time.

```suggestion
        stylesList[property.trim()] = value.trim();
```

33Orange (changes requested):
Almost done

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayToStyles = sourceString
+    .split(';')
+    .reduce((stylesList, stylesParameters) => {
+      const [property, value] = stylesParameters.split(':');
+
+      if (value !== undefined) {
+        stylesList[property.trim()] = value.trim();
+      }
```
BudnikOleksii, line 24:
```suggestion
      if (value) {
        stylesList[property.trim()] = value.trim();
      }
```
You don't need to compare it with undefined, just check if a property exists.

BudnikOleksii (approved):
Great job🚀

PR #2450, opened 2023-08-08
Develop

voievodik:


AndrMar1939 (approved):
chaining) looks great!

src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .map(el => el.trim().split(':'))
+    .filter(el => el[0].length !== 0)
+    .map(el => [el[0].trim(), el[1].trim()])
```
dffUqp, line 21: we go through array 3 times, let's remove this `map` and `filter` 

Move all logic to `reduce` method

dffUqp (changes requested):
LGTM but needs a few fixes:smiley:

use more descriptive naming

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .reduce((result, el) => {
```
roman-mirzoian, line 19:
```suggestion
    .reduce((result, element) => {
```

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+      const [key, value] = el.split(':').map(part => part.trim());
+
+      if (key.length !== 0) {
```
roman-mirzoian, line 22:
```suggestion
      if (key.length) {
```

roman-mirzoian (approved):
Well done 👍

PR #2451, opened 2023-08-08
initial commit

maksym-mishchanchuk:


Vayts (approved):
lgtm, but index.html?

maksym-mishchanchuk:
> lgtm, but index.html?

its for test))

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const separateStylesInArray = sourceString
+    .split(';').map(oneStyle => oneStyle.trim())
+    .filter(oneStyle => oneStyle !== '');
+
+  return separateStylesInArray.reduce((acc, oneStyle) => {
```
Moroz-Dmytro, line 22: Can it be combined to one chain 🤔

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const separateStylesInArray = sourceString
+    .split(';').map(oneStyle => oneStyle.trim())
+    .filter(oneStyle => oneStyle !== '');
```
Moroz-Dmytro, line 20: I guess you can avoid using `map` and `filter` if you are using `reduce` later. You can implement this logic in `reduce` to avoid using so many loops

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const separateStylesInArray = sourceString
+    .split(';').map(oneStyle => oneStyle.trim())
```
Moroz-Dmytro, line 19: Would be nice to create constants for such "magic values" - ';' ':'

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+    .filter(oneStyle => oneStyle !== '');
+
+  return separateStylesInArray.reduce((acc, oneStyle) => {
+    const styleUpdate = oneStyle
+      .split(' ')
+      .filter(letter => letter !== '' && letter !== ':')
+      .map(stringStyle => stringStyle.split('')
+        .filter(letter => letter !== ':').join(''));
```
Moroz-Dmytro, line 27: I guess it can be simplified, you don't really need so many methods

Moroz-Dmytro: `split` and `trim` is enough

Moroz-Dmytro (changes requested):
GJ 🚀 
But let's simplify your solution

tiserett (approved):

PR #2452, opened 2023-08-08
solution

AndrMar1939:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+  return propertiesArray.reduce((style, current) => {
+    const [key, value] = current.split(':');
+    const styleKey = key
+      ? key.trim()
+      : null;
+    const styleValue = value
+      ? value.trim()
+      : null;
+
+    if (!style[styleKey] && styleKey && styleValue) {
```
AndriiKuz, line 28: `!style[styleKey] && styleKey && ` remove these conditions. It will work without these)

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const propertiesArray = sourceString.split(';');
+
+  return propertiesArray.reduce((style, current) => {
+    const [key, value] = current.split(':');
+    const styleKey = key
```
AndriiKuz, line 21: You can write it without ternary operator `const styleKey = key.trim() || null;`

AndriiKuz (changes requested):
Good job, but you can improve your work)

src/convertToObject.js
AndrMar1939: thank you, yes, I made a cheking in ternary and forgot replace it

src/convertToObject.js
AndrMar1939: key.trim() doesn`t work properly without nullish operator. 

const styleKey = key?.trim() || null; will be work properly, but our version of lenter doesn`t support nullish ?.

Without ternary I will be have error:  Cannot read property 'trim' of undefined

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const propertiesArray = sourceString.split(';');
+
+  return propertiesArray.reduce((style, current) => {
+    const [key, value] = current.split(':');
+
+    if (!!key.trim() || value) {
+      style[key.trim()] = value.trim();
+    }
```
roman-mirzoian, line 24: Try to put repetitive parts of the code in variables. It's also worth checking both the `key` and the `value`.
```suggestion 
    const property = key.trim();
    const propertyValue = value.trim();
    if (property && propertyValue) {
      style[property] = propertyValue;
    }
```

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const propertiesArray = sourceString.split(';');
```
roman-mirzoian, line 17: Try to avoid the word `array` in the naming if possible :)
```suggestion
  const splittedProperties = sourceString.split(';');
```

roman-mirzoian (changes requested):

33Orange (approved):
Good job

PR #2453, opened 2023-08-09
solved task

bormaxvel:


src/convertToObject.js
```diff
@@ -1,18 +1,32 @@
...
+      }
+    });
+
+  const res = {};
+
+  resArray.forEach(line => {
+    res[line[0]] = line[1];
+  });
+
+  return res;
```
SanyaBratashchuk, line 29: use reduce

src/convertToObject.js
```diff
@@ -1,18 +1,32 @@
...
+      const vallueArray = [];
+
+      line
+        .trim()
+        .split(':')
+        .forEach(vallue => {
+          vallueArray.push(vallue.trim());
+        });
+
+      if (vallueArray.length === 2) {
```
SanyaBratashchuk, line 18: why exactly 2 it seems not flexible

SanyaBratashchuk (changes requested):

PR #2455, opened 2023-08-09
Solved task

AndriiKuz:


yuliiaant (approved):
Good job!

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+  const propertiesList = sourceString.split(';');
+
+  propertiesList.forEach(prop => {
```
33Orange, line 20: Let's use reduce instead of forEach.
In reduce method you can put an object as an initial value.

33Orange (changes requested):
Almost done

33Orange (approved):
Well done

PR #2456, opened 2023-08-10
Solution

desevoker:


src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
 'use strict';
 
+const DECLARATION_SEPARATOR = ';';
+const PROP_VALUE_SEPARATOR = ':';
+const EMPTY_STRING = '';
+
```
Wesses, line 6: Good idea! i like using constants of separators.

src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(DECLARATION_SEPARATOR)
+    .map((entry) => entry.trim())
+    .filter((entry) => entry !== EMPTY_STRING)
+    .reduce((styleObj, entry) => {
+      const [property, value] = entry.split(PROP_VALUE_SEPARATOR);
```
Wesses, line 18: You can combine `map()` and `filter()` in one `filter()`. You can also go further and combine them both with `reduce()`.

Wesses (changes requested):
Good job. But u can do better!

src/convertToObject.js
desevoker: Good point!

Moroz-Dmytro (approved):
LGTM 🔥

PR #2457, opened 2023-08-10
add task solution

Vannnkof:


src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectWithProperties = sourceString
+    .split('\n')
+    .join('')
+    .split(';')
```
KatOlista, line 12: .split('\n')
    .join('')
    .split(';') 

 change to 

.split(';\n')

KatOlista (approved):
good job!

src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
+    .map((propertie) => propertie.split(':').map((word) => word.trim()))
+    .filter((propertie) => propertie.length !== 0)
+    .reduce((accumlator, propertie) => {
```
Moroz-Dmytro, line 15: Combine it to 1 `reduce`

src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
+      return {
+        ...accumlator, [propertie[0]]: propertie[1],
+      };
+    }, {});
+
+  delete objectWithProperties[''];
```
Moroz-Dmytro, line 21: I guess you don't really need this, you can not add property if it's empty

src/convertToObject.js
Moroz-Dmytro: Fix this

Moroz-Dmytro (changes requested):
GJ 👍 
But small comments

src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';\n')
+    .reduce((acc, cssStyle) => {
+      const [key, value] = cssStyle.split(':').map(part => part.trim());
```
Moroz-Dmytro, line 12: image

src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
+
+      if (key) {
+        acc[key] = value;
+
+        return acc;
```
Moroz-Dmytro, line 17:
```suggestion
        acc[key] = value;
```

Moroz-Dmytro: You don't need 1 more return
You will return it after `if` block

src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';\n')
```
Moroz-Dmytro, line 10:
```suggestion
    .split(';')
```

Moroz-Dmytro: It will not work?

Also, would be nice to create constants for such "magic values" ';'  ':'

Moroz-Dmytro (changes requested):
Almost done 👍 
Some fixes and it is perfect 🔥

src/convertToObject.js
```diff
@@ -1,18 +1,27 @@
...
-  // write your code here
+  const SEMICOLON = ';';
+  const COLON = ':';
+
+  return sourceString
+    .split(SEMICOLON)
+    .reduce((acc, cssStyle) => {
+      const [key, value] = cssStyle
+        .split(COLON)
+        .map(part => part.trim());
```
dffUqp, line 17: I think we can remove this `map`, and use `trim` directly for key and value

dffUqp (approved):
GJ✔️

PR #2458, opened 2023-08-10
add task solution

Vayts:


polinavafik (approved):
very sophisticated 🎀

src/convertToObject.js
```diff
@@ -3,16 +3,21 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';')
+    .filter((item) => item.trim() !== '')
```
BudnikOleksii, line 13: [CODE STYLE] - All chained methods called on the same level should have same indentation. This makes code much more readable.
image
Also, in the filter you don't need to compare with an empty string you can just return `item.trim()` and all falsy values will be excluded. `.filter((item) => item.trim())`

src/convertToObject.js
```diff
@@ -3,16 +3,21 @@
...
+    .map((item) => item.split(':'))
+    .reduce((acc, item) => {
+      return {
+        ...acc,
+        [item[0].trim()]: item[1].trim(),
+      };
```
```suggestion
    .reduce((acc, [key, value]) => {
      return {
        ...acc,
        [key.trim()]: value.trim(),
      };
```
You can use destructuring.

src/convertToObject.js
```diff
@@ -3,16 +3,22 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .filter((item) => item.trim() !== '')
+    .map((item) => item.split(':'))
+    .reduce((acc, [key, value]) => {
```
Moroz-Dmytro, line 16: We can combine these 3 loops to just 1 `reduce`

src/convertToObject.js
```diff
@@ -3,16 +3,22 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
```
Moroz-Dmytro, line 13: you can create variables for these magic values ';' ':'

Moroz-Dmytro (changes requested):
Good work 👍 
But let's improve your solution

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+    .split(SEMICOLON)
+    .reduce((acc, item) => {
+      const [key, value] = item.split(COLON);
+
+      if (key.trim() !== '') {
+        return {
+          ...acc,
+          [key.trim()]: value.trim(),
+        };
+      }
```
Moroz-Dmytro, line 28:
```suggestion

      if (key.trim()) {
        acc[key.trim()] = value.trim();
      }
```

Moroz-Dmytro (approved):
GJ 👍 
But check comment below

PR #2459, opened 2023-08-10
Solution

yuliiaant:


AntonHerasymov (approved):
Good job!

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const parts = sourceString
+    .replace(/:/g, ';')
+    .split(';')
+    .map(value => value
+      .split('')
+      .filter(symbol => symbol !== '\n')
+      .join('').trim())
+    .filter((element) => element !== '');
```
BudnikOleksii, line 25: Looks like too many unnecessary actions. Let's find a better way to do this task, without replace and nested filter, join...
Also, in the filter you don't need to compare with an empty string you can just return item.element() and all falsy values will be excluded. .filter((element) => element.trim())

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+
+  const keys = parts
+    .filter(value => parts.indexOf(value) % 2 === 0);
+  const values = parts
+    .filter(value => parts.indexOf(value) % 2 !== 0);
```
BudnikOleksii, line 30: Looks strange. Too many loops for the current task.
Also, in the filter method you have access to the index and the current array, so `parts.indexOf(value)` makes no sense.
`filter((value, index, array) => index % 2 === 0);`
Docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

BudnikOleksii (changes requested):
Nice start👍
But the solution can be simplified with the power of iteration methods, so let's improve it.

33Orange (approved):
Good job

PR #2460, opened 2023-08-10
add task solution

vserhiichuk:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const separateProperties = sourceString
+    .split(';');
+
+  const separateKeysValues = separateProperties
+    .map(el => el.trim().split(':'));
```
kaavka, line 21: Trim here is unncessesary. We are triming parts on the 24 row.

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+
+  const convertObject = separateKeysValues.reduce((obj, currentValue) => {
```
kaavka, line 23: Can we name it as 'styles'?

kaavka (approved):
Looks awesome! ✨

src/convertToObject.js
Moroz-Dmytro: Fix this. Also can we avoid using `map` and implement this logic in `reduce` ?

src/convertToObject.js
Moroz-Dmytro, line 23: Combine it into 1 chain

Moroz-Dmytro: Fix this. Also try to make better name than `obj`

Moroz-Dmytro (changes requested):
Good progress 👍 
But we can improve it, check comments below

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .reduce((acc, currentValue) => {
+      const [key, value] = currentValue
+        .split(':')
+        .map(property => property.trim());
```
dffUqp, line 22: I think we can remove this `map`, and use `trim` directly for key and value

dffUqp (approved):
GJ✔️

PR #2462, opened 2023-08-10
add task solution

kaavka:
Still in progress

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -1,18 +1,23 @@
 'use strict';
 
 /**
- * Implement convertToObject function:
- *
- * Function takes a string with styles (see an example in [stylesString.js](./stylesString.js))
- * and returns an object where CSS properties are keys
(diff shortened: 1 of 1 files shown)


Orpheus29 (approved):
Wow!
Just put another 't' in the 'formated...' names)
Altogether's great
:))

src/convertToObject.js
```diff
@@ -1,18 +1,28 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const formatedStylesParts
+  = sourceString
```
BudnikOleksii, line 10:
```suggestion
  const formatedStylesParts = sourceString
```

BudnikOleksii (changes requested):
Good work👍
But could be even better, check the comments and let's improve your solution.

33Orange (approved):
Well done

PR #2463, opened 2023-08-10
add solution

KatOlista:


src/convertToObject.js
```diff
@@ -1,18 +1,19 @@
...
-  // write your code here
+  const sccStyles = {};
+
+  sourceString
+    .split(';\n')
+    .forEach(item => {
+      if (item.trim().length > 2) {
+        const [key, value] = (item.trim().split(': '));
+
+        sccStyles[key.trim()] = value.trim();
```
BudnikOleksii, line 12: Let's use reduce method instead to create an object with styles.

BudnikOleksii (changes requested):
Great work🚀
Let's change your solution with the reduce method.

src/convertToObject.js
```diff
@@ -1,18 +1,57 @@
...
+
+
+  white-space: nowrap;
+  width: auto;
+`;
+
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';\n')
```
33Orange, line 41: The code may not handle the case when the style string ends without a semicolon or has no newline character after it. Consider using a different approach to split the string, such as by semicolon alone.

src/convertToObject.js
```diff
@@ -1,18 +1,57 @@
...
+        return accum;
+      }
+
+      accum[key.trim()] = value.trim();
+
+      return accum;
+    }, {});
 }
 
+convertToObject(stylesString);
```
33Orange, line 55: The function is being called unnecessarily. Remove this line, the function should be exported and used elsewhere, not in this file.

src/convertToObject.js
KatOlista: i fixed it, but it had worked normal. I don't understand why i need to make my code unnecessarily longer

src/convertToObject.js
```diff
@@ -1,18 +1,21 @@
...
+    .reduce((accum, item) => {
+      const [key, value] = item.split(':');
+
+      if (key === undefined || value === undefined) {
+        return accum;
+      }
+
+      accum[key.trim()] = value.trim();
+
+      return accum;
```
Moroz-Dmytro, line 17:
```suggestion
      if (key && value) {
        accum[key.trim()] = value.trim();
      }

      return accum;
```

src/convertToObject.js
```diff
@@ -1,18 +1,21 @@
...
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .join('')
+    .split('\n')
```
Moroz-Dmytro, line 7:
```suggestion
```

Moroz-Dmytro: I guess It should work without this. Maybe you will need to use `trim` for `item` in reduce

Moroz-Dmytro (changes requested):
Good job 👍 
But let's make it perfect

src/convertToObject.js
```diff
@@ -1,18 +1,17 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
```
Moroz-Dmytro, line 5: Would be nice to create constants for these "magic values" ';' ':"

Moroz-Dmytro (approved):
LGTM 👍 
Small advice

PR #2464, opened 2023-08-11
Solution

AntonHerasymov:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  const lines = sourceString.split(';').map(line => line.trim());
+
+  lines.forEach(line => {
+    if (line) {
+      const [property, value] = line.split(':').map(str => str.trim());
```
BudnikOleksii, line 23: You don't need to split your code and put the blank line between. Just remember:

Also, let's use reduce method to generate stylesObject, it's a better fit in this case.

BudnikOleksii (changes requested):
Good job!
Left some comments, let's improve your solution.

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .reduce((stylesObject, line) => {
+      const [property, value] = line.trim().split(':').map(str => str.trim());
```
Moroz-Dmytro, line 20: image

33Orange (approved):
Well done

Moroz-Dmytro (changes requested):
Small comment for the future

Moroz-Dmytro (approved):

PR #2465, opened 2023-08-11
add convert to object func

rostyslav48:


viktoria-roik (approved):
Good Job!

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+  const sourceArr = sourceString
+    .split(';\n')
+    .filter((string) => string.includes(':'));
+
+  sourceArr.forEach((style) => {
+    const styleArray = style.split(':');
+    const key = styleArray[0].trim();
+    const styleValueSet = new Set(styleArray[1].split(' '));
+
+    styleValueSet.delete('');
```
33Orange, line 22: The delete operation seems unnecessary because the trim operation at line 22 should remove empty strings. If the goal is to remove potential excess white space within the value, consider using the trim() method directly on each value.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+    const styleValueArr = [...styleValueSet];
```
33Orange, line 24: There is no need to convert the set back into an array and then join it into a string. The join method can be applied directly to the array resulting from the split method at line 23.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
-  // write your code here
+  const styleObject = {};
+  const sourceArr = sourceString
+    .split(';\n')
+    .filter((string) => string.includes(':'));
+
+  sourceArr.forEach((style) => {
+    const styleArray = style.split(':');
+    const key = styleArray[0].trim();
+    const styleValueSet = new Set(styleArray[1].split(' '));
```
33Orange, line 20: Using a Set to remove whitespace might not be the best approach. It could lead to unintended results if the style value contains duplicate valid values. Instead a direct trim operation can remove leading and trailing whitespaces.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+  const sourceArr = sourceString
+    .split(';\n')
```
33Orange, line 14: The filter operation might not be necessary. If the goal is to avoid empty strings, a better approach might be to use trim on the original string before splitting it.

33Orange (changes requested):
Almost done

src/convertToObject.js
rostyslav48: trim cannot delete all empty lines from the original string

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const formatedStyle = {};
+  const removedEmptySpaces = sourceString
+    .split(';\n')
+    .filter((string) => string.includes(':'));
+
+  removedEmptySpaces.forEach((style) => {
```
Moroz-Dmytro, line 17: In this case better would be to use `reduce`

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    const keyAndValue = style.split(':');
+    const key = keyAndValue[0].trim();
+    const clearedStyle = keyAndValue[1]
+      .trim()
+      .split(' ')
+      .map(prop => prop.trim());
```
Moroz-Dmytro, line 23:
```suggestion
    const clearedStyle = keyAndValue[1].trim();
```

Moroz-Dmytro: It will not work 🤔 ?

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const formatedStyle = {};
+  const removedEmptySpaces = sourceString
+    .split(';\n')
+    .filter((string) => string.includes(':'));
```
Moroz-Dmytro, line 15: Also would be nice to create constants for magic values - ':' ';'

Moroz-Dmytro (changes requested):
Good progress 👍 
Some comments below

src/convertToObject.js
rostyslav48: No, It will not work. Due to the reason that between style props there might be some white spaces which should be deleted

rostyslav48: Actually, directly for this task it works well, however, what would we do if we get style which has some white spaces between values
example: border: solid    2px black;

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+  const removedEmptySpaces = sourceString
+    .split(ROW_SPLITTER)
+    .filter((string) => string.includes(KEY_VALUE_SPLITTER));
+
+  return removedEmptySpaces.reduce((formatedStyle, style) => {
+    const keyAndValue = style.split(KEY_VALUE_SPLITTER);
+    const key = keyAndValue[0].trim();
+    const clearedStyle = keyAndValue[1].trim();
+
+    formatedStyle[key] = clearedStyle;
```
dffUqp, line 24:
```suggestion
    formatedStyle[key.trim()] = clearedStyle.trim();
```

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const ROW_SPLITTER = ';\n';
+  const KEY_VALUE_SPLITTER = ':';
+  // const formatedStyle = {};
+  const removedEmptySpaces = sourceString
+    .split(ROW_SPLITTER)
+    .filter((string) => string.includes(KEY_VALUE_SPLITTER));
```
dffUqp, line 17: I think we can remove this `filter` and in our reduce check if value !== undefined

dffUqp (approved):
GJ✔️

PR #2466, opened 2023-08-11
Solution

neovendano:


cty2802 (approved):
Good work)

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString.trim().split(';');
+
+  const stylesObject = {};
+
+  stylesArray.forEach(style => {
```
Moroz-Dmytro, line 21: For such case we can use `reduce` 🙃

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString.trim().split(';');
```
Moroz-Dmytro, line 17: Would be nice to create constants for such "magic values" - ';' ':'

Moroz-Dmytro (changes requested):
Good job 👍 
Fix some comments and it's perfect 🔥

PR #2469, opened 2023-08-11
Solution

ydratsevska:


src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
+  const styles = sourceString
+    .split(';')
+    .map((property) => property
+      .trim()
+      .split(':')
+      .map((word) => word.trim()))
+    .filter((isNotSpace) => isNotSpace.length > 1);
+  const objectOfStyles = {};
+
+  for (const property of styles) {
```
voronine, line 19: Good job! I would add a check for empty values:

for (const property of styles) {
    const key = property[0];
    const value = property[1];

    if (key && value !== undefined) {
      objectOfStyles[key] = value;
    }
  }

voronine (approved):

src/convertToObject.js
Moroz-Dmytro, line 19: image

src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .map((property) => property
+      .trim()
+      .split(':')
+      .map((word) => word.trim()))
+    .filter((isNotSpace) => isNotSpace.length > 1);
```
Moroz-Dmytro, line 16: Combine it to 1 `reduce`

Moroz-Dmytro (changes requested):
GJ 👍 
But some comments

src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
```
Moroz-Dmytro, line 11: But also would be good to create constants for "magic values"

Moroz-Dmytro (approved):

PR #2471, opened 2023-08-11
Solution

cty2802:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesOfCss = sourceString.split(';')
+    .reduce((accum, elem) => {
```
Moroz-Dmytro, line 18:
```suggestion
  const stylesOfCss = sourceString
    .split(';')
    .reduce((accum, elem) => {
```

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+      const key = String(pair[0]).trim();
+      const value = String(pair[1]).trim();
+
+      const currentKeyValue = {};
+
+      currentKeyValue[key] = value;
+
+      return key ? {
+        ...accum, ...currentKeyValue,
+      } : accum;
```
```suggestion
      return key ? {
        ...accum,
        [key]: value,
      } : accum;
```

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesOfCss = sourceString.split(';')
+    .reduce((accum, elem) => {
+      const pair = elem.split(':');
+      const key = String(pair[0]).trim();
+      const value = String(pair[1]).trim();
```
Moroz-Dmytro, line 21:
```suggestion
      const [key, value] = elem
        .split(':')
        .map(item => item.trim());
```

src/convertToObject.js
Moroz-Dmytro: Try to make up better name, not just `elem`

annak1104 (approved):
Great job!

BudnikOleksii (approved):
Nicely done!

PR #2472, opened 2023-08-11
solution

00Mass00:


src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
...
+    .split(';')
+    .reduce((result, style) => {
+      const [property, value] = style
+        .split(':')
+        .map(part => part
+          .trim());
+
+      if (value) {
+        result[property] = value;
+      }
```
33Orange, line 25: I think it would be better to avoid some nested iterations in reduce. 

```suggestion
      const [property, value] = style.split(':');

      if (value) {
        result[property.trim()] = value.trim();
      }
```

33Orange (changes requested):
Almost done

src/convertToObject.js
```diff
@@ -1,18 +1,22 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
```
Moroz-Dmytro, line 10: Also would be nice to create constants for "magic values" - ';' ':'

Moroz-Dmytro (approved):
LGTM 👍 
Only small advice

PR #2474, opened 2023-08-11
finish solution

serkrops:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+    .reduce((acc, currentValue) => {
+      const [key, value] = currentValue.split(':');
+
+      if (value === undefined) {
+        return acc;
+      }
+
+      acc[key.trim()] = value.trim();
+
+      return acc;
```
```suggestion
      if (value) {
        acc[key.trim()] = value.trim();
      }

      return acc;
```

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = sourceString
+    .split(';')
+    .sort((a, b) => a[0].localeCompare(b[0]))
```
Moroz-Dmytro, line 19: Why do we need this sorting?

Moroz-Dmytro (changes requested):
GJ 👍 
Small comments

src/convertToObject.js
serkrops: you are rigth, it was from the first version

AnastasiiaHombalevska (approved):
good job!

Moroz-Dmytro (approved):
LGTM 👍

PR #2475, opened 2023-08-11
task solution

viktoria-roik:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+  const splittedString = sourceString.split(';');
+
+  return splittedString.reduce((acc, word) => {
+    const [key, value] = word
+      .split(':')
+      .map(item => item.trim());
+
+    if (key && value) {
+      acc[key] = value;
+    }
```
33Orange, line 27: I think it would be better to avoid some nested iterations in reduce. 

```suggestion
    const [key, value] = word.split(':')

    if (key && value) {
      acc[key.trim()] = value.trim();
    }
```

33Orange (changes requested):
Almost done

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const splittedString = sourceString
+    .split(';');
+
+  return splittedString
+    .reduce((acc, word) => {
```
Moroz-Dmytro, line 22: Better would be to combine it into one chain, as the previous mentor mentioned

image

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const splittedString = sourceString
+    .split(';');
```
Moroz-Dmytro, line 19: Would be nice to create constants for these "magic values" - ';' ':'

Moroz-Dmytro (changes requested):
Good job 👍 
But it can be perfect, some comments below 🙃

Moroz-Dmytro (approved):
LGTM 👍

PR #2476, opened 2023-08-11
Solution

ivannaromanovych:


OlhaMomot (approved):
Seems perfect to me👌 Nothing to add)

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
```
Moroz-Dmytro, line 18: Would be nice to create constants for these "magic values" - ';' ':'

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+    .reduce((prev, styleStr) => {
+      const style = styleStr.split(':');
+
+      if (style.length > 1) {
+        const styleName = style[0].trim();
+        const styleValue = style[1].trim();
```
```suggestion
      const styleName = style[0].trim();
      const styleValue = style[1].trim();
      
      if (styleName && styleValue) {
```

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+
+        return {
+          ...prev,
+          [styleName]: styleValue,
+        };
```
```suggestion
        prev[styleName] = styleValue;
```

Moroz-Dmytro: Don't need to create new copy

Moroz-Dmytro (changes requested):
Good work 🚀 
But let's make it perfect

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(PROPERTY_SEPARATOR)
+    .reduce((prev, styleStr) => {
+      const style = styleStr.split(PROPERTY_VALUE_SEPARATOR);
+
+      if (style[0].trim() && style[1].trim()) {
+        const styleName = style[0].trim();
+        const styleValue = style[1].trim();
```
Moroz-Dmytro, line 27:
```suggestion
      const styleName = style[0].trim();
      const styleValue = style[1].trim();
      
      if (styleName && styleValue) {
```

Moroz-Dmytro: Why you can't simplify it like this?

Moroz-Dmytro (approved):
GJ 👍 
But please, check comment below

src/convertToObject.js
ivannaromanovych: style[1] can be undefined, so there is an error cannot trim() undefined, so I can't do this

ivannaromanovych: I've figured it out!

PR #2479, opened 2023-08-12
Solution

Oleh0293:


k1NAs (approved):
GJ! 💪

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  const pairs = sourceString.split(';');
+
+  pairs.forEach(pair => {
```
Moroz-Dmytro, line 15: Can we combine it to one chain?
And also in this case `reduce` is better desicion than `forEach`

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+    const parts = pair.split(':');
+
+    if (parts.length === 2) {
+      const property = parts[0].trim();
+      const value = parts[1].trim();
```
Moroz-Dmytro, line 20:
```suggestion
    const property = parts[0].trim();
    const value = parts[1].trim();
      
    if (property && value) {
```

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
- *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  const pairs = sourceString.split(';');
```
Moroz-Dmytro, line 13: Also would be nice to create constants for magic values - ';' ':'

Moroz-Dmytro (changes requested):
Good progress 👍 
Some comments below

PR #2480, opened 2023-08-12
solution

OlhaMomot:


Oleh0293 (approved):
Good job)

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const properties = sourceString
+    .split(';')
```
Moroz-Dmytro, line 18: Would be nice to create constant values for such "magic values" - ';' ':'

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
+    .map(prop => prop.trim().split(':'))
+    .filter(prop => prop[0] !== '');
```
Moroz-Dmytro, line 20: Maybe we can implement this logic in `reduce` 🤔 
To avoid using 2 additional loops

Moroz-Dmytro (changes requested):
Good job 👍 
But let's make your solution even better

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
  */
+const propertySeparator = ';';
+const keySeparator = ':';
+
 function convertToObject(sourceString) {
-  // write your code here
+  const properties = sourceString
+    .split(propertySeparator)
+    // .map(prop => prop.trim().split(keySeparator))
+    // .filter(prop => prop[0] !== '');
```
Moroz-Dmytro, line 23: Delete comments

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+      if (line[0] !== '') {
+        acc.push(line);
+      }
+
+      return acc;
+    }, []);
+
+  const styleList = properties.reduce((acc, prop) => {
+    return Object.assign(acc, { [prop[0].trim()]: prop[1].trim() });
+  }, {});
```
Moroz-Dmytro, line 36: It can be done in 1 `reduce`. In this case, we will avoid using 1 more loop and increase performance of your code 🙃

Moroz-Dmytro (changes requested):
GJ 👍 
Small comments and it's perfect 🚀

roman-mirzoian (approved):
Well done 👍

PR #2481, opened 2023-08-12
solution

helenmarth:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectOfProps = {};
+  const arrOfProps = sourceString
+    .split('\n')
```
etojeDenys, line 19: i think we can split by `;`

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    .filter(item => item.includes(':'));
+
+  arrOfProps.forEach((current) => {
+    const [key, value] = current.split(':');
+
+    objectOfProps[key.trim()] = value
+      .slice(0, value.length - 1)
```
etojeDenys, line 26: and this slice will be redundant

etojeDenys (changes requested):

VitaliyBodnar (approved):
Good job!

PR #2483, opened 2023-08-13
solution convert to Object

ihorid:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .map(string => string.trim().split(':'))
+    .filter(el => el.length > 1)
+    .reduce((acc, element, i) => {
```
Yevheniia-Sid, line 21: You are not using ```i``` variable so you can remove it

Yevheniia-Sid (approved):
Good job!

src/convertToObject.js
Moroz-Dmytro, line 21: We can combine it to only 1 `reduce`

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+      return {
+        ...acc, [element[0].trim()]: element[1].trim(),
+      };
```
```suggestion
      return {
        ...acc,
        [element[0].trim()]: element[1].trim(),
      };
```

Moroz-Dmytro (changes requested):
GJ 👍 
But try to simplify your solution

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .reduce((acc, element) => {
+      const property = element.split(':');
+
+      if (property[0].trim().length !== 0) {
```
Moroz-Dmytro, line 22:
```suggestion
      if (property[0].trim()) {
```

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+        const propertyName = property[0].trim();
+        const propertyValue = property[1].trim();
+
+        return {
+          ...acc,
+          [propertyName]: propertyValue,
+        };
```
```suggestion
        acc[propertyName] = propertyValue;
```

Moroz-Dmytro: You can not create a copy

Moroz-Dmytro (approved):
Good Job 👍 
Some advice how it can be simplified even more

src/convertToObject.js
ihorid: Thank you!

PR #2484, opened 2023-08-13
add task solution

thaiduk:


dpidlutska (approved):
Good job!

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .map(property => property.trim().split(':'))
+    .filter(property => property[0])
+    .reduce((properties, property) => {
```
Moroz-Dmytro, line 21: It can be combined to 1 `reduce`

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+      const key = property[0].trim();
+      const value = property[1].trim();
+
+      return {
+        ...properties, [key]: value,
+      };
```
Moroz-Dmytro, line 27:
```suggestion
      return {
        ...properties,
        [key]: value,
      };
```

Moroz-Dmytro (changes requested):
GJ 🚀 
But let's improve it

PR #2488, opened 2023-08-13
add task solution

irahos:


src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
+    .join('; ')
+    .split(': ')
+    .map(str => str.trim())
+    .join(': ')
+    .split(';')
+    .filter(element => element !== ' ' && element !== '')
+    .map(element => element.trim())
+    .sort((element1, element2) => {
+      return element1.localeCompare(element2);
+    });
```
Moroz-Dmytro, line 28: I guess we don't need sorting here

src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
+  const transformedToSortedString = sourceString.split(';')
+    .map(str => str.trim())
+    .join('; ')
+    .split(': ')
+    .map(str => str.trim())
+    .join(': ')
+    .split(';')
+    .filter(element => element !== ' ' && element !== '')
+    .map(element => element.trim())
+    .sort((element1, element2) => {
```
Moroz-Dmytro, line 26: Too many methods, can it be simplified 🤔 

P.S. You can do it using only `split` and `reduce`

src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
+      return element1.localeCompare(element2);
+    });
+
+  const cssObject = {};
+
+  transformedToSortedString.forEach(element => {
```
Moroz-Dmytro, line 32: You can use `reduce` in such case

Moroz-Dmytro (changes requested):
Good progress 👍 
But let's simplify your solution

PR #2490, opened 2023-08-14
add task solution

dpidlutska:


src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const lines = sourceString.split(';').map((line) => line.trim());
```
Taras2705, line 4: 4. [CODE STYLE] - All chained methods called on the same level should have same indentation. This makes code much more readable.
BAD EXAMPLE:

GOOD EXAMPLE:

const cssStyles = styles
 .map(style => style)
 .slice(0, 5)

Taras2705 (commented):
good job !!

src/convertToObject.js
Moroz-Dmytro: Good point, please fix it

Moroz-Dmytro, line 4: Would be nice to create constants for "magic values" - ';' ':'

src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
+  const cleanLines = lines.map((line) => {
+    const parts = line.split(':').map((subPart) => subPart.trim());
+
+    return parts;
+  });
```
Moroz-Dmytro, line 9: Maybe we can avoid this 1 more loop(`map`) and implement this in `reduce`?

Moroz-Dmytro (changes requested):
Good work 👍 
But let's improve your solution

serkrops:
nice done!

PR #2491, opened 2023-08-14
js_style-to-object

vatatan:


NazarVotkanych (approved):
Good job!

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
```
Moroz-Dmytro, line 18: Would be nice move these values ';' ':' to constants

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    .reduce((resultOfReduce, element) => {
```
Moroz-Dmytro, line 19: Try to make up better names

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+      const [key, value] = element
+        .split(':')
+        .map(partOfElement => partOfElement.trim());
+
+      if (key.length) {
```
```suggestion
      if (key) {
```

Moroz-Dmytro: I guess, It will also work, because:
image

Moroz-Dmytro (changes requested):
Good job 👍 
Only small comments

src/convertToObject.js
vatatan: done

src/convertToObject.js
vatatan: done

src/convertToObject.js
vatatan: done

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const separatorSemicolon = ';';
+  const separatorColon = ':';
```
Moroz-Dmytro, line 18: Only such constants should be named in UPPER_CASE, and would be better to move them outside the function

Moroz-Dmytro (approved):
Good job 👍 
Small advice about naming

polosanya (approved):
👍

PR #2495, opened 2023-08-14
Solution

okanivan:


roman-mirzoian (changes requested):
Try to choose a more meaningful naming.

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splittedSource = sourceString.split(';');
+
+  const convertedObject = {};
+
+  splittedSource.forEach(style => {
```
Moroz-Dmytro, line 21: Better would be to use `reduce` in this case

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splittedSource = sourceString.split(';');
```
Moroz-Dmytro, line 17: Also would be nice to create constants for "magic values"

Moroz-Dmytro (approved):
Good work 👍 
But please check the comments below, how you can improve it

PR #2497, opened 2023-08-14
solution

Valentyn-Bilyk:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceStringToObj = sourceString
+    .split(';')
+    .reduce((styleObject, currentStyle) => {
+      if (currentStyle.trim() === '') {
```
dffUqp, line 15: we used `currentStyle.trim()` twice, let's create a separate variable for this

dffUqp (changes requested):
LGTM but needs a few minor fixes:smiley:

tiserett (approved):

PR #2500, opened 2023-08-15
Solution

oksom:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  const lines = sourceString.split('\n');
```
lerastarynets, line 19: Splitting the sourceString by '\n' may not cover all cases as CSS properties in the string could be separated by ';' and not necessarily a newline. Consider splitting by ';' instead to cover all potential cases.

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+
+  lines.forEach(line => {
+    const match = line.match(/^\s*([\w-]+)\s*:\s*([^;]*)\s*;/);
+
+    if (match) {
+      const propertyName = match[1].trim();
```
lerastarynets, line 25: There is no real need to trim() the propertyName and propertyValue variables as the regex used in line 23 already excludes leading and trailing whitespaces.

lerastarynets (approved):
Great job overall! Left a few comments to make your code more advanced)

PR #2502, opened 2023-08-15
Solution

Nikitavyskreb:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleArray = sourceString
+    .split(';')
+    .filter(property => property.trim())
```
roman-mirzoian, line 19: It is not necessary to `filter` the array separately, this logic can be transferred to the middle of the `reduce` and `trim` for each element before starting work.

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+      const [key, value] = property.split(':');
+
+      if (key && value) {
+        acc[key.trim()] = value.trim();
+      }
+
+      return acc;
+    }, {});
+
+  return styleArray;
```
roman-mirzoian, line 30: If you want to make changes by calling them in a chain, you don't need to create a separate variable and you can immediately return the result.

roman-mirzoian (changes requested):

dpidlutska (approved):

PR #2506, opened 2023-08-15
solution

Duster-1:


src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  const styleArray = sourceString.split(';');
+
+  for (const style of styleArray) {
+    const styleParams = style.split(':');
```
nazarmatsevych, line 14:
```suggestion
    const [key, value] = style.split(':');
```

Use array destructuring

src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
+
+    if (styleParams[0] && styleParams[1]) {
```
nazarmatsevych, line 16:
```suggestion
    if (key && value) {
```

src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
+      result[styleParams[0].trim()] = styleParams[1].trim();
```
nazarmatsevych, line 17:
```suggestion
      result[key.trim()] = value.trim();
```

nazarmatsevych (changes requested):
Good work, left minor comments to be fixed

src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
+
+  const styleArray = sourceString.split(';');
+
+  for (const style of styleArray) {
+    const [key, value] = style.split(':');
+
+    if (key && value) {
+      result[key.trim()] = value.trim();
+    }
+  }
```
etojeDenys, line 19: you can try to rewrite it using `reduce`

etojeDenys (approved):

PR #2507, opened 2023-08-16
Solution

serhiilapin:


src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
+      removeEmptyKeys.push(arrayOfKeys[i]);
+    }
+  }
+
+  for (let i = 0; i < removeEmptyKeys.length; i++) {
+    removeEmptyKeys[i] = removeEmptyKeys[i].split(':');
+  }
+
+  for (const key of removeEmptyKeys) {
+    key[0] = key[0].trim();
```
irini132, line 36: You are modifying the array you are iterating over (removeEmptyKeys), this could lead to unexpected results. You should modify the object values directly instead.

src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
+  const arrayOfKeys = sourceString.split('\n');
+  const removeEmptyKeys = [];
+  const resultObject = {};
+
+  for (let i = 0; i < arrayOfKeys.length; i++) {
+    arrayOfKeys[i] = arrayOfKeys[i].replace(/^[^\w\d-]*/, '')
+      .replace(/[; ]+/g, ' ')
+      .replace(/:\s+/g, ':');
+
+    if (arrayOfKeys[i].length > 2) {
```
irini132, line 26: You are checking if the length of the key is greater than 2, but keys can be one character long. You should check if the key is not an empty string instead.

irini132 (changes requested):
Fix comments please

vlasiuksasha (changes requested):
check comments

PolinaRomanenko (approved):

PR #2514, opened 2023-08-17
solution

oleg-matvienko:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  const bundlesToCheck = sourceString
+    .replace('\n', '')
+    .split(';');
+
+  bundlesToCheck.forEach(bundle => {
```
Moroz-Dmytro, line 23: In this case better to use `reduce`

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+    const trimmedBundle = bundle.trim();
+
+    if (trimmedBundle) {
+      const [property, value] = trimmedBundle
+        .split(':')
+        .map(item => item.trim());
+
+      stylesObject[property] = value;
```
Moroz-Dmytro, line 31:
```suggestion

      stylesObject[property.trim()] = value.trim();
```

src/convertToObject.js
Moroz-Dmytro, line 23: Combine it to one chain

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  const bundlesToCheck = sourceString
+    .replace('\n', '')
+    .split(';');
```
Moroz-Dmytro, line 21: Would be nice to create constants for all "magic values" ';' ':' ...

Moroz-Dmytro (changes requested):
Good work but let's improve it

src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const space = '\n';
```
BudnikOleksii, line 17: It's not a space, it's a LF (Line Feed) → Used as a new line character in Unix/Mac OS X. and you don't need to replace them.
Also, not local constants should be named in uppercase and you don't need to keep them inside the function.

BudnikOleksii (changes requested):
Good job!
Just a few moments, check them.

roman-mirzoian (approved):
Well done 👍

PR #2517, opened 2023-08-19
task has been done

vladkorobka:


src/convertToObject.js
```diff
@@ -11,8 +11,68 @@
...
+
+  propsArray.forEach(item => {
+    const key = item[0];
+    const value = item[1];
+
+    cssObject[key] = value;
+  });
+
+  return cssObject;
 }
```
nazarmatsevych, line 76: Solution looks too comlicated, try to use split and reduce methods to improve it

nazarmatsevych (changes requested):

anastasiiavorobiova (approved):
Awesome work!

PR #2527, opened 2023-08-24
string with styles was convert to object

EkaterinaBessonnaya:


anastasiia-tilikina (approved):
GJ!

lerastarynets (approved):
You're doing brilliantly, keep it up!

PR #2537, opened 2023-08-29
solution

Nikita211112:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArr = sourceString.split(';');
+  const newObj = {};
+
+  for (const ch of sourceArr) {
+    const [key, value] = ch.split(':');
+
+    if (value !== undefined) {
```
DarkMistyRoom, line 23:
```suggestion
    if (value) {
```

DarkMistyRoom: it gives the same result

DarkMistyRoom: But what if there are only white spaces after splitting? Should we add these styles?

DarkMistyRoom (changes requested):
Good job! Let's handle empty values

src/convertToObject.js
Nikita211112: > But what if there are only white spaces after splitting? Should we add these styles?

No, I think it's pointless to write down these styles

maxim2310 (approved):
Good work!!!

PR #2543, opened 2023-08-30
solution

yuliiaaaaa:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  const arrayOfProperties = sourceString.split(':').join().split(';');
```
anastasiiavorobiova, line 18: There is no need to split properties and join properties if you are going to split them by the comma few lines below. Consider removing redundant logic

anastasiiavorobiova (changes requested):
Great work! Added the tiny comment to fix

PR #2544, opened 2023-08-30
Solution

SergiyVS:


src/convertToObject.js
```diff
@@ -3,16 +3,43 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrOfProperties
```
lerastarynets, line 17: The variable 'arrOfProperties' declaration and assignment are separated by a newline. This decreases readability and is not in line with common JavaScript syntax. It would be better to declare and assign on the same line.

src/convertToObject.js
```diff
@@ -3,16 +3,43 @@
...
+
+          convertedProp[0] = convertedProp[0].trim();
+          convertedProp[1] = convertedProp[1].trim();
+
+          return convertedProp;
+        }
+      });
+
+  const objOfProperties = arrOfProperties.reduce((prev, prop) => {
+    if (prop === undefined) {
```
lerastarynets, line 31: Instead of checking if 'prop' is undefined, you should filter out the undefined values after the map function. This would make your code cleaner and more efficient.

src/convertToObject.js
```diff
@@ -3,16 +3,43 @@
...
+          const convertedProp = property.split(':');
+
+          convertedProp[0] = convertedProp[0].trim();
+          convertedProp[1] = convertedProp[1].trim();
+
+          return convertedProp;
+        }
+      });
+
+  const objOfProperties = arrOfProperties.reduce((prev, prop) => {
```
lerastarynets, line 30: pls make use of method chaining

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

lerastarynets (changes requested):
.

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
 /**
  * Implement convertToObject function:
  *
- * Function takes a string with styles (see an example in [stylesString.js](./stylesString.js))
+ * Function takes a string with styles (see an example in
```
OleziO, line 6: Remove all comments from the code.

OleziO (changes requested):

lerastarynets (approved):
You're on the right track!

PR #2546, opened 2023-08-31
Solution

GIroSTarK:


src/convertToObject.js
```diff
@@ -3,16 +3,44 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splittedBySemicolon = sourceString.split(';');
+  const resultArr = [];
+  const clearStrings = [];
+
+  for (const string of splittedBySemicolon) {
```
anastasiiavorobiova, line 21: This cycle seems redundant, you can just trim elements from splittedBySemicolon array by mapping through it

anastasiiavorobiova (changes requested):
Great work! Almost done

mykhalenych (approved):

PR #2549, opened 2023-09-04
Solution

HachevskyyMarkiyan:


src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
+    properties[i] = properties[i].trim();
+
+    if (properties[i] === EMPTY_LINE || properties[i] === SEMICOLON) {
+      delete properties[i];
+    } else {
+      const temp = properties[i].split(COLON);
+
+      result[temp[0].trim()] = temp[1].substring(0, temp[1].length - 1).trim();
+    }
+  }
```
maxim2310, line 35: you can just skip this case
```suggestion
    if (!properties[i] === EMPTY_LINE && !properties[i] === SEMICOLON) {
      const temp = properties[i].split(COLON);

      result[temp[0].trim()] = temp[1].substring(0, temp[1].length - 1).trim();
    }
  }
```

maxim2310 (changes requested):
Good job! Let's improve your code a bit)

VitaliyBodnar (approved):
Good job!

PR #2550, opened 2023-09-05
task done

IntensoSergey:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArrey = sourceString.split(';');
```
Anastasiia-Svintsova, line 17: Typo in the variable name 'sourceArrey'. It should be 'sourceArray'.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  return sourceArrey.reduce((result, item) => {
+    const splitedItem = item.split(':');
```
Anastasiia-Svintsova, line 20: You can destructure it 
```suggestion
    const [key, value] = item.split(':');
```

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+    if (splitedItem.length !== 2) {
```
Anastasiia-Svintsova, line 22: Check if there are key and value instead of length

Anastasiia-Svintsova (changes requested):
Good start!

anastasiiavorobiova (approved):
Awesome!

PR #2554, opened 2023-09-06
add task solution

mishakoshilka:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+ * and values are the values of related CSS properties (see an exampl in
+ * [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const newWords = sourceString.split('\n');
```
etojeDenys, line 17:
```suggestion
  const newWords = sourceString.split(';');
```

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+
+  for (let i = 0; i < newWords.length; i++) {
+    const target = newWords[i].trim();
+    const current = target.split(':');
+
+    if (current.length === 2) {
+      const property = current[0].trim();
+      let value = current[1].trim();
+
+      value = value.slice(0, -1);
```
etojeDenys, line 28: and this line will be redundant

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+      value = value.trim();
```
etojeDenys, line 29: as this one

etojeDenys (changes requested):

PR #2556, opened 2023-09-06
Solution

valeriaoleynik:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const parts = sourceString.split(';').map(elem => elem.split(':'));
```
lerastarynets, line 17: The input string might have leading or trailing whitespaces, you should trim it **before** splitting and other actions on it.

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

witflash (approved):
Well done 👍

PR #2557, opened 2023-09-07
style to object solution

Masha-Cactus:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
-  // write your code here
+  const source = sourceString.split(';');
+  const result = {};
+  let key = '';
+  let value = '';
+  let startKeyIndex = 0;
+  let endKeyIndex = 0;
+
+  for (let prop = 0; prop < source.length; prop++) {
+    startKeyIndex = source[prop].lastIndexOf('\n ') + 1;
```
mykhalenych, line 25:
```suggestion
    const [key, value] = source[prop].split(':');
```

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+    endKeyIndex = source[prop].indexOf(':');
+    key = source[prop].slice(startKeyIndex, (endKeyIndex)).trim();
+    value = source[prop].slice(endKeyIndex + 1).trim();
+
+    if (key && value) {
+      result[key] = value;
```
mykhalenych, line 31:
```suggestion
      result[key.trim()] = value.trim();
```

mykhalenych (changes requested):
Try to simplify your code

src/convertToObject.js
Masha-Cactus: rly, in this way it looks much simple, thanks. 
Can U tell me what does \n mean? Why did it disappear after split method?

anastasiiavorobiova (approved):
Awesome!

PR #2567, opened 2023-09-13
function developed

StepanMustafin:


maxim2310 (changes requested):
Good work!!!

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  allCommands.forEach((item) => {
+    const commandParts = item.split(':');
+
+    if (commandParts[0] !== undefined && commandParts[1] !== undefined) {
+      const key = commandParts[0].trim();
+      const value = commandParts[1].trim();
+
+      result[key] = value;
+    }
```
VitaliyBodnar, line 28: let's improve your solution 
```suggestion
    const commandParts = item.split(':');
    const [firstCommandPart, secondCommandPart] = commandParts;

    if (firstCommandPart && secondCommandPart) {
      const key = firstCommandPart.trim();
      const value = secondCommandPart.trim();

      result[key] = value;
    }
```

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):
Good job!

PR #2568, opened 2023-09-14
add task solution

ololopos:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const NEW_LINE = '\n';
+  const EMPTY_LINE = '';
+  const SEMICOLON = ';';
+  const COLON = ':';
+
+  const properties = sourceString.split(NEW_LINE);
```
etojeDenys, line 22: we can just split by SEMICOLON

etojeDenys (changes requested):

src/convertToObject.js
ololopos: done

ololopos:
fixed , please review

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+  const properties = sourceString.split(SEMICOLON);
+  const result = {};
+
+  for (const i in properties) {
+    properties[i] = properties[i].trim();
+
+    if (properties[i] !== EMPTY_LINE && properties[i] !== SEMICOLON) {
+      const temp = properties[i].split(COLON);
+
+      result[temp[0].trim()] = temp[1].substring(0, temp[1].length).trim();
```
etojeDenys, line 30: why do we need to call substring here?

etojeDenys (changes requested):

src/convertToObject.js
ololopos: fixed

ololopos:
fixed please review

plodnikate (approved):
good job !

PR #2572, opened 2023-09-18
Solution

denys-khvashchenko:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssString = sourceString;
+  const cssObject = {};
+
+  cssString
+    .split(';')
+    .map((pair) => {
```
VitaliyBodnar, line 22: method `map`  should return new array and don't mutate current, because map it a `clear function`
you can use method `forEach` instead of `map`
but in this case, it's better to use method reduced, because you can avoid creating additional variable `cssObject `

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssString = sourceString;
```
anastasiia-tilikina, line 17: you do not really need to assign parameter to var since we don't mutate initial string

denedweb: agree

maxim2310 (approved):
Good work!!!

PR #2575, opened 2023-09-19
solution task

Butovych:
I used the "if (itemToArray[1] === 'both      ')" construct because the trim() method doesn't work for this row.


VitaliyBodnar (changes requested):

sTorba24 (approved):
Well done !

PR #2576, opened 2023-09-19
add solution

h1deeeee:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+ * and values are the values of related CSS
+ * properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrSourse = sourceString.split('\n');
```
Viktor-Kost, line 17: we can just split by `;` symbol

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+  const convertedObj = {};
+
+  for (let i = 0; i < arrSourse.length; i++) {
+    const element = arrSourse[i].replace(';', '').trim();
+
+    if (element === '' || element === ';') {
+      continue;
+    }
+
+    const part = element.split(':');
```
Viktor-Kost, line 27: it's enough to do like this 
```suggestion
  const [key, value] = arrSourse[i].split(':');
```

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+
+    const key = part[0].trim();
+    const value = part[1].trim();
+
+    convertedObj[key] = value;
```
Viktor-Kost, line 32: you can exchange this way
```suggestion
    if (key && value) {
      convertedObj[key.trim()] = value.trim();
    }
```

Viktor-Kost (changes requested):

VitaliyBodnar (approved):

PR #2580, opened 2023-09-20
Converting styles to object

Shostyy:


src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  const KEY_INDEX = 0;
+  const VALUE_INDEX = 1;
+  const SPLIT_BY_SYMBOL = `\n`;
```
DarkMistyRoom, line 8: why don't spit by `;` ?

src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
+  const SEPARATE_BY_SYMBOL = ':';
+  const MINIMAL_LENGTH_ALLOW = 3;
+
+  sourceString.split(SPLIT_BY_SYMBOL)
+    .filter((currentElement) =>
+      currentElement.length > MINIMAL_LENGTH_ALLOW)
```
DarkMistyRoom, line 14: would be better to filter after splitting by `: `because we possibly can get `style:  ;` with an empty value or vice versa

PR #2583, opened 2023-09-21
add solution js_style-to-objec

A1eeex:


koros-rk (approved):
Looks great ;)

src/convertToObject.js
```diff
@@ -1,18 +1,33 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
polosanya, line 17: It is better to specify what exactly the result is)

src/convertToObject.js
```diff
@@ -1,18 +1,33 @@
...
+  const getParametrs = sourceString.trim().split(';');
+
+  getParametrs.forEach((item, i) => {
+    if (item !== '') {
+      const [key, value] = (item.trim().split(':'));
+
+      if (key !== undefined && value !== undefined) {
```
polosanya, line 24:
```suggestion
      if (key && value) {
```
You can simplify it

polosanya (approved):
Nice!

PR #2588, opened 2023-09-24
Solution

katya-sn:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const getData = sourceString.trim().split(';');
+
+  getData.forEach((item) => {
+    if (item !== '') {
+      const [key, value] = (item.trim().split(':'));
+
+      if (key !== undefined && value !== undefined) {
```
volodymyr-soltys97, line 24: Consider how you can rewrite this condition so that you don't explicitly write a check for undefinеd, as this is bad practice
```suggestion
      if (key !== undefined && value !== undefined) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/convertToObject.js
katya-sn: smthng like this :
 if (!key &&& !value) ?

etojeDenys: yeah, exactly

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #2590, opened 2023-09-24
Solution

YaroslavYarynych:


src/convertToObject.js
```diff
@@ -1,18 +1,36 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourseArr = sourceString.split('\n');
+
+  const result = sourseArr.reduce((acc, el) => {
+    const parts = el.split(':');
+    const key = parts[0].trim();
+
+    let value = parts.slice(1).toString().trim();
```
adioskv, line 15: 4. [CODE STYLE] - All chained methods called on the same level should have same indentation. This makes code much more readable.


BAD EXAMPLE:
```
```

GOOD EXAMPLE: 
```
const cssStyles = styles
 .map(style => style)
 .slice(0, 5)
```
It's also belongs to string methods, we need to make them readable too

src/convertToObject.js
```diff
@@ -1,18 +1,36 @@
...
+      value = value.trim();
+    }
+
+    acc[key] = value;
+
+    return acc;
+  }, {});
+
+  delete result[''];
+  delete result[';'];
```
adioskv, line 31: To avoid this, you can use method `.filter` for your array (Styles always containing ':' symbol)

adioskv (changes requested):
Good job! Just complete checklist requirements, and you'll be a king 👑

adioskv (approved):
Good job!

StasSokolov (approved):
🔥

PR #2593, opened 2023-09-25
add solution convert string to object

Kuper18:
Hi, here is my solution

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -3,16 +3,45 @@
 /**
  * Implement convertToObject function:
  *
- * Function takes a string with styles (see an example in [stylesString.js](./stylesString.js))
+ * Function takes a string with styles (see an example in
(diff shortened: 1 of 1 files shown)


src/convertToObject.js
```diff
@@ -3,16 +3,45 @@
...
+    .split('\n')
+    .filter((item) => item !== '')
+    .map((item) =>
+      item
+        .split(' ')
+        .filter((value) => value !== '' && value !== ';' && value !== ':')
+    );
+
+  normalizeValues.forEach((item) => {
+    let key = item.slice(0, 1).join('');
```
OlhaKoziuk, line 28: maybe you can use destructuring

OlhaKoziuk (approved):
Great job! but maybe you can make it shorter) with destructuring for example

polosanya (approved):
👍

PR #2595, opened 2023-09-25
solution

OlhaKoziuk:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+ * and values are the values of related CSS properties
+ *  (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
yulabakun, line 17: [CHECKLIST] 2.[CODE STYLE] - Avoid variable namings like array , object arr, obj, result. It will decrease code readability.

StasSokolov, line 17: [CHECKLIST] 2.[CODE STYLE] - Avoid variable namings like array , object arr, obj, result. It will decrease code readability.

StasSokolov (changes requested):
Great job, check please only one comment!

PR #2597, opened 2023-09-25
added task solution

Stako21:


src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const newArr = sourceString
+  .split(';')
+  .map(val => val.trim().split(':'))
+  .filter(val => val.length > 1);
```
VitaliyBodnar, line 18: code it not aligned. 
```suggestion
  const newArr = sourceString
    .split(';')
    .map(val => val.trim().split(':'))
    .filter(val => val.length > 1);
```

src/convertToObject.js
```diff
@@ -12,7 +12,15 @@
...
+
+  const newObj = {};
+  newArr.map((arr) => (newObj[arr[0].trim()] = arr[1].trim()))
```
VitaliyBodnar, line 21: it's better to use array destructuring as it will make a code cleaner, for instance:
`const [key, value] = arr;`

VitaliyBodnar, line 21: method map is a clear function, so you shouldn't modify a data in this method
other way, you can use method forEach for that

VitaliyBodnar (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
-  // write your code here
+  const newArr = sourceString
+    .split(';')
+    .map(val => val.trim().split(':'))
+    .filter(val => val.length > 1);
+
+  const newObj = {};
+
+  newArr.forEach((...arr) => {
+    const [[ key, value ]] = arr;
```
plodnikate, line 25: Can be simplified 
```
newArr.forEach(([key, value]) => {
    newObj[key.trim()] = value.trim();
});
```

plodnikate (changes requested):
Good job! Let's improve a little!

etojeDenys (approved):
awesome

PR #2599, opened 2023-09-25
Converting string to styles object

ElenaGrasovskaya:


sviatikkondur (approved):
LGTM

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArray = sourceString.split(';');
+  const stylesObject = {};
+  const MIN_STYLE_LENGTH = 5;
+
+  sourceArray.forEach(el => {
+    if (el.length > MIN_STYLE_LENGTH) {
```
vasylmiskiv, line 22: interesting way to get rid of unnecessary spaces)
You can destructure your key value from each string like so
`const [key, value] = el.split(:);`
And instead of `if (el.length > MIN_STYLE_LENGTH` you could check 
```
if (key && value) {
  stylesObject[key.trim()] = value.trim();
}
```

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+      const stylePair = el.split(':');
+
+      Object.assign(stylesObject, {
```
vasylmiskiv, line 25: You can just set directly key with value at each iteration for your object

vasylmiskiv (approved):

src/convertToObject.js
ElenaGrasovskaya: Thanks! Looks much better!

src/convertToObject.js
ElenaGrasovskaya: Did that as well

PR #2604, opened 2023-09-26
add solution

Vitalikmy:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssProperties = {};
+  const propertyValuePairs = sourceString.split(';');
+
+  for (const pair of propertyValuePairs) {
```
Drach-Illya, line 20: **[IMPORTANT]**

This for ... of cycle is ALSO meant to be an iteration. Please find a way to remove it and incorporate an iteration in its place.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+    const [property, value] = pair.split(':').map((str) => str.trim());
```
Drach-Illya, line 21: [CODE STYLE] - All chained methods called on the same level should have same indentation. This makes code much more readable.


BAD EXAMPLE:
```
```

GOOD EXAMPLE: 
```
const cssStyles = styles
 .map(style => style)
 .slice(0, 5)
```

Please edit the code style used on this line.

Drach-Illya (changes requested):
Very impressive code, however, there are some necessary changes according to our Checklist.

PR #2605, opened 2023-09-26
style to object

tetyana-pol:


vlevik (approved):
good job! also you could try method forEach instead of loop to practise array methods:)

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+  const styleObject = {};
+  const styleArray = sourceString
+    .split(';')
+    .map(style => style.trim())
+    .filter(style => style.length > 0)
+    .map(style => style.split(':'));
+
+  for (const [property, value] of styleArray) {
+    styleObject[property.trim()] = value.trim();
+  }
```
superpooperxxx, line 26: You can combine all of these lines either in one `forEach` or `reduce`.
1. Start the `forEach`
2. Split by `:`
3. Check if both property and value are defined
4. Add the style to `styleObject`. (Don't forget to trim everything)

superpooperxxx (changes requested):
Good job🔥See the comments

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+
+  sourceString.split(';')
+    .map(element => element.split(':'))
```
vasylmiskiv, line 20: You could do it without `map` method only with `forEach` and at every iteration you could destructure `key` and `value` from the string like so
`const [key, value] = element.split(':')`

vasylmiskiv (approved):

src/convertToObject.js
tetyana-pol: thank you

PR #2609, opened 2023-09-26
add task solution

tilugdmytro:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    .split('\n')
+    .map(line => line.trim())
+    .filter(line => line.length > 1)
+    .map(ch => ch.replace(';', '').split(':'));
+
+  const styledProperties = {};
+
+  for (const line of lines) {
+    styledProperties[line[0].trim()] = line[1].trim();
+  }
```
BetterSol, line 27: good job! 
it would be great to improve your code even more according to the checklist:
`5. [CODE PRACTICE] - Avoid using loops in this task, use iteration methods instead.`

BetterSol (changes requested):
Great job! 
Suggested one small detail to make your code even better:)

StasSokolov (approved):
Good job🔥

PR #2610, opened 2023-09-26
add task solution

mpohorenyi:


dolphinnq (approved):
Great job!

superpooperxxx (changes requested):
Well, it works🔥, but you basically have SIX loops here, which is too much. What if I say, that you can combine line 19-26 into one reduce))

StasSokolov (approved):
Great job!

PR #2611, opened 2023-09-26
solution

shchipina:


src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+
+  sourceString.trim().split(';')
+    .map(item => item.trim().split(':'))
```
MVoutch, line 20: Good work, but you need to make a few corrections

You don't need trim() here, and to avoid using trim() in the future, you can add a map() after split() in which you will iterate over the values and add trim() to them

```
.map(item => item
        .split(':')
        .map(x => x.trim()))
```

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
+    .filter(item => item.length > 1)
+    .forEach(([name, property]) => {
+      styles[name.trim()] = property.trim();
```
MVoutch, line 23: Since we went through the properties in the beginning and removed whitespace with trim(), we no longer need trim() here

`styles[name] = property;`

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+
+  sourceString.trim().split(';')
```
MVoutch, line 19: In this case, trim() does nothing

MVoutch (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+
+  sourceString
+    .split(';')
+    .map(style => style.split(':'))
+    .filter(style => style.length > 1)
```
superpooperxxx, line 22: I am pretty sure you can implement this task without map and filter on lines 21 and 22. Just use `forEach`, split by `:` and trim values  inside it

superpooperxxx (changes requested):
Good job🔥, but you can do cleaner😉

polosanya (approved):
Well done!

PR #2614, opened 2023-09-26
first solution

Clover247:


HardCodWorker (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+  const stylesArray = sourceString.split(';');
+
+  stylesArray.filter(style => style.includes(':')).forEach(element => {
+    const [key, value] = element.split(':');
+
+    stylesObject[key.trim()] = value.trim();
+  });
```
superpooperxxx, line 24: 1. I don't think that `filter` is an efficient solution here. I think you can check inside `forEach` if the key and value exists and then add it to object. The less loops the better.
2. 🚨Formatting: If you apply a chain of methods to an array each of them should start on new line
```
  stylesArray
    .filter(style => style.includes(':'))
    .forEach(element => {

      stylesObject[key.trim()] = value.trim();
    });

**Review by superpooperxxx — CHANGES_REQUESTED** [review]
Good job🔥 You are almost there, check the comments

File: src/convertToObject.js [hunk 2]
```diff
@@ -3,16 +3,31 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+  const stylesArray = sourceString.split(';');
+
+  stylesArray.forEach(element => {
+    const [key, value] = element
+      .split(':')
+      .map(item => item.trim());
+
```
superpooperxxx, line 24: Much better , the only thing instead of `.map(item => item.trim());` I would do this `stylesObject[key.trim()] = value.trim();`

superpooperxxx (approved):
🔥

PR #2617, opened 2023-09-27
done

Victoriia-Melnyk:


src/convertToObject.js
```diff
@@ -1,18 +1,49 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const splited = sourceString.split(';');
```
VitaliyBodnar, line 19: - your solution look like a bit complicated, let's try to improve it )
 - you can overwrite this code with only one cycle for
- please check this solution

Screenshot 2023-09-27 at 13 10 12

VitaliyBodnar (changes requested):

lerastarynets (approved):

PR #2618, opened 2023-09-27
Solution

SerhiiUnhurian:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resultObj = {};
+  // const properties = [];
+  // const styles = [];
```
VitaliyBodnar, line 19: remove all comments

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+  const stylesArr = sourceString.split(';');
+
+  for (const property of stylesArr) {
+    const splittesProperty = property.split(':');
+
+    if (splittesProperty[0].trim().length
+    && splittesProperty[1].trim().length > 0) {
```
VitaliyBodnar, line 26: it's better to use array destructuring here, for instance:
`const [key, value] = splittesProperty`;

VitaliyBodnar (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  for (const property of stylesArr) {
+    const splittesProperty = property.split(':');
+
+    const [key, value] = splittesProperty;
+
+    if (key.trim().length
+    && value.trim().length > 0) {
+      resultObj[key.trim()] = value.trim();
+    }
```
VitaliyBodnar, line 28: let's align your code 
```suggestion
    const splittesProperty = property.split(':');
    const [key, value] = splittesProperty;

    if (
      key.trim().length &&
      value.trim().length
    ) {
      resultObj[key.trim()] = value.trim();
    }
```

VitaliyBodnar (changes requested):

src/convertToObject.js
SerhiiUnhurian: okay

Viktor-Kost (approved):

PR #2620, opened 2023-09-27
convertToObjects task

HardCodWorker:


Stim1 (approved):
Looks nice for me!

src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resObject = {};
+  const arrProperties = sourceString.split(';');
+
+  for (const property of arrProperties) {
+    const proprContainer = property.split(':');
```
ArtemTeslenko, line 19: Maybe better to say propsContainer.

src/convertToObject.js
```diff
@@ -12,7 +12,20 @@
...
+
+    const [key, value] = proprContainer;
+
+    if (key.trim().length && value.trim().length) {
+      resObject[key.trim()] = value.trim();
```
ArtemTeslenko, line 24: You use twice key.trim(), the same with value. Think how to avoid it.

ArtemTeslenko (approved):
Good job! There are some comments, consider next time.

PR #2621, opened 2023-09-27
task solution

annabasenyuk:


tetyana-pol (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceObject = {};
+  const sourceArray = sourceString.split(';');
+
+  sourceArray.filter(style => style
+    .includes(':'))
+    .forEach(element => {
```
superpooperxxx, line 18: 1. I don't see that this `filter` is needed here. All of the strings inside the array have `:`
2. Formatting
```
  sourceArray
    .filter(style => style.includes(':'))
    .forEach(element => {

**Review by superpooperxxx — CHANGES_REQUESTED** [review]
Hey, good job🔥 I think you can do it in 2 steps: 
1. split by `;`
2. forEach (check inside if value and key exists, then add to object)

File: src/convertToObject.js [hunk 2]
```diff
@@ -3,16 +3,25 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceObject = {};
+  const sourceArray = sourceString.split(';');
+
+  sourceArray.forEach(element => {
+    const [key, value] = element.split(':').map(style => style.trim());
```
vasylmiskiv, line 17: map is not needed here because you have only  two elements key and value. And you are creating a new array via `map`, which is unnecessary

src/convertToObject.js
```diff
@@ -3,16 +3,25 @@
...
+
+    if (key && value) {
+      sourceObject[key] = value;
```
vasylmiskiv, line 20: Instead of it you could use trim method here
`sourceObject[key.trim()] = value.trim();`

vasylmiskiv (approved):

src/convertToObject.js
annabasenyuk: Thanks, I'll know)

PR #2625, opened 2023-09-27
Solution

Ayogoharo:


Clover247 (approved):
Your solution is ingenious!

src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
+/*
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const strSplited = sourceString.split(';');
+  const resault = {};
```
superpooperxxx, line 11: 1. Let's always try to name variables more descriptive. Let's call it `objectWithStyles`
3. Typo `result`

src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
+
+  strSplited.forEach((element) => {
+    const [param, value] = element.split(':');
+
+    if (param !== '' && value !== undefined) {
+      resault[param.trim()] = value.trim();
+    }
+  });
```
superpooperxxx, line 19: Your condition can be just
```
if (param && value)

superpooperxxx (changes requested):
The solution is perfect, nice and efficient🔥 Just some minor mistakes

polosanya (approved):
Well done!

PR #2628, opened 2023-09-27
add task slution str to obj

denzzlle171:


src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resultObject = {};
+  const stylesArray = sourceString
+    .split(';')
+    .map((item) => item.trim())
+    .filter(Boolean);
```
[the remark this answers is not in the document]
denzzlle171: we just remove all values that will be set to false, well, in principle, a simple comparison with an empty string will probably be enough here

src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+  const stylesArray = sourceString
+    .split(';')
+    .map((item) => item.trim())
+    .filter(item => item !== '');
```
superpooperxxx, line 13: I don't think you really need these 2 lines.
1. You trim it later anyways
2. Just add a condition inside `forEach` to check if both value and key are defined then ...

superpooperxxx (changes requested):
I think we can improve the efficiency of your code🙏🏻

tetyana-pol (approved):

src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  sourceString
+    .split(';')
+    .forEach((rule) => {
+      const [key, value] = rule.split(':').map((part) => part.trim());
```
superpooperxxx, line 14: You could also `trim` the elements like this `stylesObject[key.trim()] = value.trim();`, cause there are always only 2 elements in this array and `map` might be an overkill.

superpooperxxx (approved):
🔥

PR #2633, opened 2023-09-30
Solution

RVDotsenko:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+}
+
+function getObject(array) {
+  const objStyle = {};
+
+  array.forEach(element => {
+    const colon = element.indexOf(':');
+
+    if (colon > 0) {
+      objStyle[element.slice(0, colon).trim()]
```
loralevitska, line 30: Long expression `element.slice(0, colon).trim()` and `element.slice(colon + 1).trim()` move to variable with meaningful variable name for more readability.

loralevitska (changes requested):
Almost perfect) Let's improve

src/convertToObject.js
RVDotsenko: Thank's for your comment!)

loralevitska (approved):
🔥

PR #2635, opened 2023-10-01
finished the task

mifyy:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const newObj = {};
+
```
superpooperxxx, line 18: Let's give more descriptive name, at the end it is not a `new object` it is an `object with styles`

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+  const arraySource = sourceString.split(';');
+
+  for (let i = 0; i < arraySource.length - 1; i++) {
+    const keysPush = arraySource[i].split(':')
+      .map(str => str.trimEnd().trimStart());
+
```
superpooperxxx, line 24: 1. You could use `for of` it would look more elegant
2. You can use destructuring instead of defining `keysPush` variable.
3. We have a shorthand `trim()` method that replaces this `str.trimEnd().trimStart()`
4. `map` method might be an overkill.

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    if (keysPush.length > 1) {
+      newObj[keysPush[0]] = keysPush[1];
+    }
```
superpooperxxx, line 27: I'd recommend not checking the length, but checking if property and value is defined. Again it is more descriptive that length

superpooperxxx (changes requested):
Overall good job🔥 The code works, just some improvements

mifyy:
Hi! I have made the corrections.

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObj = {};
+
+  const arraySource = sourceString.split(';');
+
+  for (const key of arraySource) {
+    const [property, value] = key.split(':')
+      .map(str => str.trim());
```
polosanya, line 23:
```suggestion
    const [property, value] = key
      .split(':')
      .map(str => str.trim());
```
image

polosanya (approved):
👍

PR #2636, opened 2023-10-02
Solution

hrysta:


src/convertToObject.js
```diff
@@ -2,17 +2,31 @@
...
- *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const converted = {};
+  const sourceArray = sourceString
+    .replace(/[\r\n]/gm, '')
```
vasylmiskiv, line 12: You shouldn't use regular expressions here. This is not necessary at all

src/convertToObject.js
```diff
@@ -2,17 +2,31 @@
...
+    .split(';')
+    .filter((a) => a.length > 4)
+    .sort((a, b) => a.localeCompare(b));
+
+  for (const el of sourceArray) {
+    const splittedEl = el.split(':');
```
vasylmiskiv, line 18: You can destructure your properties from your array like so 
`const [key, value] = el.split(':')`

src/convertToObject.js
```diff
@@ -2,17 +2,31 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const converted = {};
+  const sourceArray = sourceString
+    .replace(/[\r\n]/gm, '')
+    .split(';')
+    .filter((a) => a.length > 4)
+    .sort((a, b) => a.localeCompare(b));
```
vasylmiskiv, line 15: sort method is unnecessary here

src/convertToObject.js
```diff
@@ -2,17 +2,31 @@
...
+
+  for (const el of sourceArray) {
+    const splittedEl = el.split(':');
+    const key = splittedEl[0].trim();
+    const value = splittedEl[1].trim();
+
+    converted[key] = value;
+
+    if (converted[key] === undefined) {
```
vasylmiskiv, line 24: What is the meaning of this logic? Is it necessary here?

vasylmiskiv (changes requested):

src/convertToObject.js
```diff
@@ -2,17 +2,28 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const converted = {};
+  const sourceArray = sourceString
+    .split(/\r?\n/)
+    .join('')
+    .split(';')
+    .filter((a) => a.length > 4);
```
superpooperxxx, line 15: This is just too much manipulations and each of them comes at its own cost.
Steps:
1. Just split by `;`
2. Do the rest inside the `for loop` (split by `:`, trim and update the `converted` object)

superpooperxxx (changes requested):
Good job🔥, but you can do much cleaner

src/convertToObject.js
StasSokolov, line 15: will be better use reduce for this

StasSokolov (approved):
Great job, let's improve your code a bit more!

PR #2637, opened 2023-10-02
update convertToObject

shevchenkotaras:


lerastarynets (approved):

PR #2643, opened 2023-10-04
Solution

ihor-pavlov:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+  styleArray.forEach(style => {
+    const [property, value] = style.split(':');
+
+    if (property && value) {
+      const propertyName = property.trim();
+      const propertyValue = value.trim();
+
+      styleObject[propertyName] = propertyValue;
+    }
+  });
```
VitaliyBondarenko1982, line 29: suggest to replace forEach method on reduce and immediately return result of reduce
```
 return styleArray.reuduce((acc, style) => {
   // do something
 }, {})
```

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest a little refactoring.'

src/convertToObject.js
VitaliyBondarenko1982: ok, i will approve but left possible solution. may be you will find what was wrong in your attemts with reduce)
it can be something like that.
```
return styleArray.reduce( (acc, style) => {

        if (!key) {
          return acc;
        }

        return {
          ...acc, [key]: value,
        };
  }, {});
```

VitaliyBondarenko1982 (approved):
Well done

PR #2648, opened 2023-10-05
solution

mshmigelska:


src/convertToObject.js
```diff
@@ -11,8 +11,24 @@
...
+  for (let i = 0; i < styleArray.length; i++) {
+    const propName = styleArray[i].trim().split(':');
+
+    if (propName.length === 2 && propName[0].trim()) {
+      const property = propName[0].trim();
+      const value = propName[1].trim();
+
+      styleObject[property] = value;
+    }
+  }
```
nazarmatsevych, line 29:
```suggestion
  for (const prop of styleArray) {
    const [property, value] = prop.trim().split(':');

    if (property && value) {
      styleObject[property] = value;
    }
  }
```

Same but a bit shorter :)

nazarmatsevych (approved):
Good work, left minor comment

nazarmatsevych (approved):
*

PR #2649, opened 2023-10-05
add task solution

tetyana-shatrova:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+  const array = sourceString.split(';');
+  const arrayRes = array
+    .map((item) => item.trim())
+    .filter((item) => item !== '')
+    .map((item) => item.split(':'));
+
+  arrayRes.forEach((item) => {
+    item[0] = item[0].trim();
+    item[1] = item[1].trim();
+  });
```
etojeDenys, line 26: forEach doesn't change the array, so this loop looks redundant

etojeDenys (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+
+  const res = {};
+
+  for (let i = 0; i < arrayRes.length; i++) {
+    res[arrayRes[i][0]] = arrayRes[i][1];
+  }
```
etojeDenys, line 32:
```suggestion
  const res = {};

  for (let i = 0; i < arrayRes.length; i++) {
    res[arrayRes[i][0].trim()] = arrayRes[i][1].trim();
  }
```
we can simplify it

etojeDenys (changes requested):

anastasiia-tilikina (approved):
👍🏼

PR #2651, opened 2023-10-06
create

LeoSpartaDeV:


src/convertToObject.js
```diff
@@ -1,18 +1,35 @@
...
+  let sourseArr = sourceArr.filter((item, i) => {
+    if (item === '\n' || (item === ' ' && sourceArr[i - 1] === ' ')) {
+      return false;
+    }
+
+    return true;
+  });
+
+  sourseArr = sourseArr.join('');
+  sourseArr = sourseArr.split(';');
```
etojeDenys, line 21: we can just split it by ';'

etojeDenys: and then remove spaces using the `trim` method

etojeDenys (changes requested):

src/convertToObject.js
LeoSpartaDeV: thanks, I went step by step, through the debugger and the console, and confused myself

src/convertToObject.js
```diff
@@ -1,18 +1,24 @@
...
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const cssClass = {};
+  const sourceArr = sourceString.split(';');
+
+  sourceArr.forEach((elem) => {
+    const LastIndOfKey = elem.indexOf(':');
```
IvanFesenko, line 13: camelCase 
```suggestion
    const lastIndOfKey = elem.indexOf(':');
```

IvanFesenko (approved):
Pay attention on code style

PR #2655, opened 2023-10-07
solution

nikkias:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+      .map(property => property.trim())
+    );
+
+  for (const [key, value] of stylesArr) {
+    if (key) {
+      obj[key] = value;
+    }
+  }
+
+  return obj;
```
SanyaBratashchuk, line 34: let's practice using array iteration methods. Try to rewrite this part using `reduce`

SanyaBratashchuk (changes requested):

VitaliyBodnar (approved):
Good job!

PR #2659, opened 2023-10-09
Solution

basiak89:


src/convertToObject.js
```diff
@@ -1,18 +1,18 @@
...
- *
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
   // write your code here
-}
+  const result = {};
+  const sourceArray = sourceString.split(';');
```
vlasiuksasha, line 6: It is better to handle possible trailing semicolon in the sourceString before splitting it. Otherwise, you may end up with an empty string as the last element of the array, which will create an unnecessary property in the resulting object. You can use trim() function to remove trailing semicolon.

src/convertToObject.js
```diff
@@ -1,18 +1,18 @@
...
+
+  sourceArray.map(elem => {
```
vlasiuksasha, line 8: You are using map function but not returning anything from it. The map function should be used when you want to transform elements in an array. In this case, you want to perform side effects (modifying external 'result' variable), so 'forEach' is more appropriate.

src/convertToObject.js
```diff
@@ -1,18 +1,18 @@
...
+    const smallArray = elem.split(':');
 
+    if (smallArray.length > 1) {
+      result[smallArray[0].trim()] = smallArray[1].trim();
+    }
+  });
+
+  return result;
```
vlasiuksasha, line 16: Code style: There is an unnecessary empty line before the 'return' statement. It's good practice to keep your code concise and avoid unnecessary blank lines.

vlasiuksasha (commented):
check the comments

vlasiuksasha (changes requested):
check the comments

src/convertToObject.js
basiak89: fixed

src/convertToObject.js
basiak89: fixed

src/convertToObject.js
basiak89: fixed

basiak89:
hello @vlasiuksasha please could you have a look at my second commit?

PR #2663, opened 2023-10-11
solution

ArtemSydun:


src/convertToObject.js
```diff
@@ -1,18 +1,27 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const string = sourceString.split('\n')
```
polosanya, line 11: Try to provide more meaningful names instead of `result` and `string`.
Actually, it's not even string but array

src/convertToObject.js
```diff
@@ -1,18 +1,27 @@
...
+    .filter(item => item.length > 0)
+    .map(item => item.split(':'))
+    .map(word => word.map(restyle => restyle.replace(';', '').trim()))
+    .filter(item => item.length > 1);
+
+  for (let i = 0; i < string.length; i++) {
+    const key = string[i][0];
+    const value = string[i][1];
```
polosanya, line 19: What are `key` and `value`? Would be great to specify what exactly these values are

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments)

PR #2665, opened 2023-10-12
task

MariiaSuper:


src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
-  // write your code here
+  const res = {};
+
+  const arrStr = sourceString.split('\n');
+
+  for (const element of arrStr) {
+    let newElement = element.trim();
+
+    if (newElement.includes(':')) {
+      newElement = newElement.split(':');
```
nazarmatsevych, line 12:
```suggestion
      [key, value] = newElement.split(':');
```

Use array destructuring

nazarmatsevych (approved):
Good work, left minor comment to improve your solution

PR #2667, opened 2023-10-13
Testing

Olesichka:


src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const cssObject = {};
+
+  
+  const cssRules = sourceString.split(';');
+
+  for (const rule of cssRules) {
+
+    const [property, value] = rule.split(':');
```
vlasiuksasha, line 11: You should trim the property and value immediately after splitting to ensure no leading or trailing white spaces.

src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssObject = {};
+
+  
+  const cssRules = sourceString.split(';');
```
vlasiuksasha, line 7: Before splitting the string, you should trim the source string to avoid leading and trailing white spaces.

src/convertToObject.js
```diff
@@ -1,18 +1,26 @@
...
+
+  for (const rule of cssRules) {
+
+    const [property, value] = rule.split(':');
+
+    if (property && value) {
+
```
vlasiuksasha, line 14: The if condition is not necessary. If the property or value is undefined, it would not affect the final object.

vlasiuksasha (changes requested):
check the comments

4ksandrey (approved):
Accepted, well done

PR #2675, opened 2023-10-17
solution

BodyaKutsyk:


src/convertToObject.js
```diff
@@ -12,7 +14,25 @@
...
+  formatedString.forEach(elem => {
+    keysArray.push(elem.slice(0, elem.indexOf(':')).trim());
+    valuesArray.push(elem.slice(elem.indexOf(':') + 1).trim());
+  });
+
+  keysArray.forEach((elem, index) => {
+    if (elem !== '') {
+      outputObject[elem] = valuesArray[index];
+    }
+  });
```
nazarmatsevych, line 31: We can simplify it using reduce method:

```
const result = formattedString.reduce((outputObject, elem) => {
  const [key, value] = elem.split(":").map(item => item.trim());
  if (key !== '') {
    outputObject[key] = value;
  }
  return outputObject;
}, {});
```

nazarmatsevych (approved):
Well done!

PR #2676, opened 2023-10-19
add task solution

PodvAx:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+  const result = {};
+  const cssArr = sourceString
+    .split(';')
+    .map(property => property.trim())
+    .map(property => property.split(/\s*:\s*/));
+
+  for (const property of cssArr) {
+    if (property.length === 2) {
+      result[property[0]] = property[1];
+    }
```
VitaliyBodnar, line 26: - it's better to use destructuring here, as it will make a code cleaner 
- please also move 2 to the constant

```suggestion
    if (property.length === 2) {
      const [key, value] = property;
      result[key] = value;
    }
```

VitaliyBodnar (changes requested):

anastasiiavorobiova (approved):
Awesome work!

PR #2677, opened 2023-10-20
Solution

andreysakhno:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split('\n')
+    .map(rule => rule.replace(';', '').trim())
+    .filter(rule => rule)
```
VitaliyBodnar, line 20: let's optimize your code
- you can delete this filter and add a condition to the reduce method, for instance 
```
.reduce((prev, rule) => {
 if (rule) {
  ...
 }
```

VitaliyBodnar (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split('\n')
+    .reduce((prev, rule) => {
+      const next = { ...prev };
+      const currentRule = rule.replace(';', '');
```
anastasiiavorobiova, line 21: Consider using split here or replace all occurrences

anastasiiavorobiova (commented):
Great work! Added a tiny comment to fix

anastasiiavorobiova (changes requested):
Great work! Added a tiny comment to fix

PR #2682, opened 2023-10-22
The 'convertToObject' function has ben implemented

Dmytro-Bashlai:


src/convertToObject.js
```diff
@@ -3,16 +3,47 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArr = sourceString.replace(/\s +|\n/g, '').split(';');
+
+  return sourceArr
+    .filter(item => item !== '')
+    .map(str => str.replace(/^:/, ''))
+    .reduce((objectWithStyle, el) => {
+      const key = el.split(':')[0];
+      const value = el.split(':')[1].replace(/^\s/g, '');
```
VitaliyBodnar, line 24: don't see any sense to do `el.split(':')` two times here, you can do it once and save the result to variable

src/convertToObject.js
```diff
@@ -3,16 +3,47 @@
...
+  //   .split(':')
+  //   .map(substring => substring.trim())
+  //   .join(':')
+  //   .split(';')
+  //   .map(substring => substring.split(':'))
+  //   .reduce((styles, sub, index) => {
+  //     return {
+  //       ...styles, [sub[0]]: sub[1],
+  //     };
+  //   }, {});
```
VitaliyBodnar, line 46: remove commented code

VitaliyBodnar (changes requested):

TarasHoliuk (approved):

PR #2686, opened 2023-10-24
task solution

Svyatoslav-kolom:


src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
-  // write your code here
+  const cssObject = {};
+  const lines = sourceString.split('\n');
+
+  for (const line of lines) {
+    const parts = line.split(':');
+
+    if (parts.length === 2) {
+      const property = parts[0];
+      const value = parts[1];
```
VitaliyBodnar, line 26: it's better to use arrays destructuring as it will make a code shorter
 ```suggestion
      const [property, value] = parts;
```

**Review by VitaliyBodnar — CHANGES_REQUESTED** [review]
Almost done!

File: src/convertToObject.js [hunk 2]
```diff
@@ -3,16 +3,38 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssObject = {};
+  const lines = sourceString.split('\n');
```
etojeDenys, line 19: we can simplify the logic and split by the `;`

etojeDenys (changes requested):

Svyatoslav-kolom:
like that?

nazarmatsevych (approved):
Good work!

PR #2690, opened 2023-10-26
Solution style-to-object

TVChulkova:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+  const result = {};
+  const str = sourceString.split(';');
+
+  for (const item of str) {
+    const idx = item.indexOf(':');
+    const property = item.slice(0, idx).trim();
+
+    if (property) {
+      result[property] = item.slice(idx + 1).trim();
+    }
```
etojeDenys, line 26: you can try to simplify this logic and split the item by ":"

etojeDenys (changes requested):

src/convertToObject.js
TVChulkova: Me sims this logic isn't difficult. Could you please clarify what I need to change for the right solution?

TVChulkova: Is that what you mean?

etojeDenys (approved):
awesome

src/convertToObject.js
etojeDenys: It looks better, right?

PR #2695, opened 2023-10-28
Style to {}_task

vashivan:


src/convertToObject.js
```diff
@@ -1,18 +1,21 @@
...
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObj = {};
+
+  const splittedStr = sourceString.split(';');
+
+  splittedStr.map((item) => {
```
VitaliyBodnar, line 8: method map is supposed to be used when you need to return a new array, but in this you are not return anything 
so, make sense to use method forEach here

VitaliyBodnar (changes requested):
Good job!

VitaliyBodnar (changes requested):
.

etojeDenys (approved):
good job

PR #2704, opened 2023-11-01
add task solution

ABahatiy:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+
+  // split all props to separate strings
+
+  const props = sourceString.split(';')
+    .map((prop) => prop.replace(/[\n, \s]+/, ''))
+    .filter((prop) => prop.length > 0);
```
lerastarynets, line 23:
```suggestion
    .filter((prop) => !!prop.length);
```

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

ABahatiy:
Thank`s for review !

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+  const propsObj = {};
+
+  // spilt each propetry to key value and assing to object
+  for (let i = 0; i < props.length; i++) {
+    const prop = props[i].split(':');
+
+    propsObj[prop[0].trim()] = prop[1].trim();
+  }
+
+  return propsObj;
```
IvanFesenko, line 35: last loop can be replaces wia Array.reduce()
```suggestion
  return sourceString.split(';')
    .map((prop) => prop.replace(/[\n, \s]+/, ''))
    .filter((prop) => !!prop.length)
    .reduce((acc, styleRow) => {
      const [key, value] = styleRow.split(':');

      acc[key.trim()] = value.trim();
      
      return acc;
    }, {})
```

IvanFesenko (approved):
Well done

PR #2709, opened 2023-11-01
add task solution

7yrak34:


src/convertToObject.js
```diff
@@ -12,7 +12,31 @@
...
+    );
+
+  const props = properties
+    .filter(prop => prop.length > 1);
+
+  const result = props
+    .reduce((styles, property) => {
+      if (property.length === 0) {
+        return styles;
+      }
```
etojeDenys, line 32: looks redundant 'cause you already filtered by length

etojeDenys (approved):
good job

PR #2710, opened 2023-11-03
finished solution for converting styles to object

MrHunterBoi:


yevtok (approved):
Nice solution!

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const splittedString = sourceString.split(';');
+  const stylesObj = {};
+
+  const propsNamesAndValues = splittedString
+    .map(prop => prop.trim().split(':'))
+    .filter(prop => prop[0]);
+
+  propsNamesAndValues.forEach(prop => {
```
ArtemTeslenko, line 24: what is the goal of interruption of chaining? It is possible to use `forEach` after `filter` here

ArtemTeslenko (approved):
Good job 👍 See comment)

src/convertToObject.js
MrHunterBoi: Thought it'd be less confusing

PR #2711, opened 2023-11-04
feat [js-adv]: implement `string-to-object` mapper

lexzane:


src/convertToObject.js
```diff
@@ -1,18 +1,35 @@
...
+  const result = {};
+
+  sourceString.split(SEMICOLON)
+    .map(item => item.trim().replace(TRIM_COLON_REGEX, COLON))
+    .filter(value => value)
+    .forEach(item => {
+      const properties = item.split(COLON);
+
+      result[properties[0]] = properties[1];
+    });
```
SerhiiKirik, line 30: Try to make it better, use reduce instead `forEach`

SerhiiKirik (changes requested):
Great 🔥 
Let's improve your code 🧑‍💻

PR #2712, opened 2023-11-04
solution

yevtok:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arrayCreate = sourceString.split(';');
```
artemforsoff, line 17: It is not clear from the variable name what it contains.
If another developer has to work with this code in the future, it will not be immediately clear to him what is meant.
Would be better, like, for example, `cssRules`.

artemforsoff (approved):
Good and neat solution 🔥

PR #2713, opened 2023-11-04
solution

errisi:


Yevhen-Kolosov (commented):
Great job! 👍 
The only thing is a slight checklist mismatch.

Yevhen-Kolosov (commented):
-

Yevhen-Kolosov (changes requested):
_Suggested changes above._

src/convertToObject.js
```diff
@@ -11,8 +11,24 @@
  *
  * @return {object}
  */
+/* eslint-enable */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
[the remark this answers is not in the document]
BudnikOleksii: > According to the checklist, you should give this variable a different name:
> 
> 2. [CODE STYLE] - Avoid variable namings like `array` , `object` `arr`, `obj`, `result`. It will decrease code readability

Good point, naming is important

src/convertToObject.js
```diff
@@ -11,8 +11,24 @@
...
+
+  sourceString
+    .split(';')
+    .forEach(style => {
```
BudnikOleksii, line 21: You can use `reduce` method to generate an object instead of creating an empty object and `forEach`

BudnikOleksii (changes requested):
Good job 👍
Don't forget about the checklist and how to name your variables

src/convertToObject.js
```diff
@@ -11,8 +11,24 @@
  *
  * @return {object}
  */
+/* eslint-enable */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const styleToObject = {};
```
roman-mirzoian, line 17: A variable must report what it stores:
```suggestion
  const convertedStyles = {};
```

src/convertToObject.js
roman-mirzoian: Not fixed 😢

roman-mirzoian (changes requested):
Not all comments from the previous review have been corrected, be more careful next time, and if you have questions, ask them in the chat or on QnA.

roman-mirzoian (approved):
Well done 👍

PR #2718, opened 2023-11-06
add solution

yefimov77:
- [x] done!

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -3,16 +3,35 @@
 /**
  * Implement convertToObject function:
  *
- * Function takes a string with styles (see an example in [stylesString.js](./stylesString.js))
+ * Function takes a string with styles (see an example
+ * in [stylesString.js](./stylesString.js))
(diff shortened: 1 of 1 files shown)


ArtemSydun (approved):
good job

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+      .split('')
+      .filter((item) => item !== '\n')
+      .join('');
+  });
+
+  for (const part of finalArr) {
+    const pair = part.split(':');
+    const [style, key] = pair;
 
+    if (!result[style] && key !== undefined) {
```
ArtemTeslenko, line 30: try to avoid comparing with undefined or null, if there is key it will return you true like: `if (!result[style] && key)`

ArtemTeslenko (approved):
Good job) I am approving it, but do not use contraction when naming, in all tasks, it is not good, use meaningful names. When you work in team nobody use contractions 😉

PR #2721, opened 2023-11-07
add task solution

mykhailoyurkovskyi:


PonomarchukLana (approved):
:v:

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  */
 function convertToObject(sourceString) {
   // write your code here
+  const result = {};
+  const arrOfStyles = sourceString.split(';');
+
+  arrOfStyles.forEach(pair => {
+    const property = pair.split(':');
+
+    if (property.length === 2) {
```
ArtemTeslenko, line 24: `2` should be a CONSTANT for example `const EXPECTED_PAIR_LENGTH = 2;`

ArtemTeslenko (approved):
Good job) I am approving it, but work with naming, see comments 😉

PR #2723, opened 2023-11-07
add task solution

ruslanagnatenko:


ViktorAfk (approved):

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+  const sources = sourceString.split(';')
+    .map(
+      string => string.trim().split(':')
+        .map(str => str.trim()),
+    );
+
+  return sources.reduce((styleObject, style) => {
+    styleObject[style[0]] = style[1];
+
+    delete styleObject[''];
```
BudnikOleksii, line 26:
```suggestion
    if (style[0] && style[1]) {
      styleObject[style[0]] = style[1];
    }
```
You can add properties only if they exist instead of deleting every time

BudnikOleksii (approved):
Nicely done 🔥

PR #2725, opened 2023-11-07
done task

olegnovikov10:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
```
OstinM, line 18: 2. [CODE STYLE] - Avoid variable namings like `array` , `object` `arr`, `obj`, `result`. It will decrease code readability

OstinM (changes requested):
Good job! but need to fix something!

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
-  // write your code here
+  const stylesObj = {};
+
+  const arrFromSource = sourceString.split(';');
+  const newArrSplit = arrFromSource.map(item => item.split(':'));
+  const arrFilter = newArrSplit.filter(item => item.length > 1);
+
+  arrFilter.forEach(item => {
+    stylesObj[item[0].trim()] = item[1].trim();
+  });
```
ArtemTeslenko, line 26: 1. do not use contractions - arr, obj etc.
2. it can be chained like split().map().filter().......

ArtemTeslenko (commented):
Good job 👍 I am approving, but see comments)

ArtemTeslenko (approved):
Good job!

PR #2726, opened 2023-11-07
style to object

PonomarchukLana:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.split(';')
```
mickhakk, line 17: [checklist] 2. [CODE STYLE] - Avoid variable namings like `array` , `object` `arr`, `obj`, `result`. It will decrease code readability
same on lines 21, 24-26

mickhakk (approved):
Good job!

src/convertToObject.js
BudnikOleksii: > [checklist] 2. [CODE STYLE] - Avoid variable namings like `array` , `object` `arr`, `obj`, `result`. It will decrease code readability same on lines 21, 24-26

Good point

BudnikOleksii: Fix naming, variables should describe them, `arr/result/el/item` are bad names.

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  const result = {};
+
+  arr.forEach(item => {
+    const el = item.split(':');
+    const key = el[0].trim();
+    const value = el[1].trim();
+
+    result[key] = value;
+  });
```
BudnikOleksii, line 29: You can create an object with `reduce` method directly

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.split(';')
+    .map(el => el.trim())
+    .filter(el => el !== '');
```
```suggestion
    .filter(el => el);
```
You don't need to compare with an empty string, you can check if a value exists.

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments and suggestions.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const declarations = sourceString.split(';')
+    .map(el => el.trim())
+    .filter(el => el);
```
ArtemTeslenko, line 19: do not use contractions and use meaningful names

ArtemTeslenko (approved):
Good job! Do not use contractions, work with naming 😉

PR #2729, opened 2023-11-07
Solution

zaura333:


src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const styled = {};
+  const parts = sourceString
+    .split(';');
```
Patrykmclaren14, line 20: Using the `replace` function before the `split` will save us a lot of time in the subsequent steps.

src/convertToObject.js
```diff
@@ -3,16 +3,38 @@
...
+  parts.forEach(part => {
+    const colonIndex = part.indexOf(':');
+
+    if (colonIndex !== -1) {
+      const key = part
+        .slice(0, colonIndex)
+        .trim();
+      const value = part
+        .slice(colonIndex + 1)
+        .trim();
```
Patrykmclaren14, line 31: Instead of using `if` and truncating from x to y at the start, you can use `replace` to remove spaces in the array. Then, split it into two parts using `split(':')`. The style of the value is separated by a colon. Finally, apply key-value parsing with `const [key, value] = str.trim().split(':')` to obtain the key and value.

Patrykmclaren14 (approved):
The code works well and meets the requirements. I provided some hints on my approach and how I addressed this task. 🙂

mvjl000 (approved):
Great job! I love the simplicity 🏆

PR #2730, opened 2023-11-07
Develop

OstinM:


ksyvvkaa (approved):
nice solution 💪 I guess it works faster than mine, with maps and filter

src/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Document</title>
+</head>
+<body>
+  asdas
```
ArtemTeslenko, line 9: remove extra code

src/convertToObject.js
```diff
@@ -1,18 +1,39 @@
...
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const styleCssList = {};
+  const splitedString = sourceString.split(';');
+
+  splitedString.forEach(el => {
+    const listElement = el.split(':');
```
ArtemTeslenko, line 24: do not use contractions, `el` is not meaningful name

ArtemTeslenko (approved):
Good job 👍 I am approving, but do not use contractions)

PR #2731, opened 2023-11-07
Solution

mickhakk:


src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
-  // write your code here
+  return sourceString
+    .split(';')
+    .filter(style => style.trim())
+    .map(style => {
+      const [property, value] = style.split(':').map(part => part.trim());
+
+      return { [property]: value };
+    })
+    .reduce((result, style) => ({
```
VeDono, line 25: Amazing realization 👀, it mostly easy for reading and so compact.
Your realization better of my.

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .filter(style => style.trim())
+    .map(style => {
+      const [property, value] = style.split(':').map(part => part.trim());
```
VeDono, line 21: i think 'property' must be named like: 'key'. ¯\_(ツ)_/¯
But it's my opinion.

VeDono (approved):

src/convertToObject.js
ArtemTeslenko, line 25: good decision to use `reduce`, but with `reduce` you can simplify your code and get rid of `map` inside `map`, like 

> return sourceString.split(";").reduce((result, style) => {
>     const stylePair = style.split(":");
>     const key = stylePair[0]?.trim();
>     const value = stylePair[1]?.trim();
>     if (key && value) {
>       result[key] = value;
>     }
>     return result;
>   }, {});

ArtemTeslenko (changes requested):
Good job 👍 But do not overfill your code by logic, map inside map is not a good idea, you can make it easier 😉

src/convertToObject.js
```diff
@@ -3,16 +3,28 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';').reduce((result, style) => {
+    const stylePair = style.split(':');
+    const key = stylePair[0] ? stylePair[0].trim() : undefined;
+    const value = stylePair[1] ? stylePair[1].trim() : undefined;
```
ArtemTeslenko, line 20: If you do just `const key = stylePair[0]?.trim();` it will work, if you want to specify nothing it is better to use `null` in such cases, because `undefined` it is variable that has been declared but does not have value

ArtemTeslenko (approved):
Good job 👍 Please see comment

PR #2734, opened 2023-11-07
Solution

mberezhna:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString.split(';');
+  const styleObject = {};
+
+  styles.forEach(style => {
```
BudnikOleksii, line 20: You can use `reduce` method instead to create an object

BudnikOleksii (approved):
Looks good 👍 
Just one comment left

PR #2737, opened 2023-11-07
convertToObject task solution

VeDono:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
+ * and values are the values of related CSS properties
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
polosanya, line 15: Let's specify what exactly the result is
image

VeDono: Ok, i pay attention on this in future works. Thanks 🤝

PR #2739, opened 2023-11-08
initial commit

Barosz30:


src/convertToObject.js
```diff
@@ -3,16 +3,54 @@
...
+  });
+
+  // cleaning inside, this time triming each element
+  array = array.map((key) => {
+    const fixedKey = key.map((el) => el.trim());
+
+    return fixedKey;
+  });
+
+  // console.log(array);
```
zaura333, line 42: We can get rid of unnecessary comments

src/convertToObject.js
```diff
@@ -3,16 +3,54 @@
...
+  array = array.map((key) => {
+    return key.split(':');
+  });
+
+  // cleaning inside, this time triming each element
+  array = array.map((key) => {
+    const fixedKey = key.map((el) => el.trim());
+
+    return fixedKey;
+  });
```
zaura333, line 40: Instead of overwriting array, you could try to implement chaining of array methods

zaura333, line 40: E.g. here you could use two methods on key ( .split(...).map(...) ) while mapping the whole array only one time

zaura333 (approved):
Overall good job! :) Code works and the purpose of each step is clear, but there's room to make it more compact and limit unnecessary repeating

mvjl000 (changes requested):
I totally agree with the previous reviewer. You are iterating through the array way more times than necessary.

It works, good for you ;) But I would love to see a 'cleaner' solution

mvjl000 (approved):
Looks good!

PR #2740, opened 2023-11-08
Develop

YehorHeneralov:


mberezhna (approved):
Good job!

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const startedArray = sourceString
+    .split(';')
+    .map(el => el.trim());
```
BudnikOleksii, line 19: array/object/el are bad names, variable names should describe what type of data is stored in it.
For example, `el` - property, `formattedObject` - `styles`

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+
+  const arrayWithoutEmptyStrings = startedArray.filter(el => el.length > 0);
```
```suggestion
  const startedArray = sourceString
    .split(';')
    .map(el => el.trim())
    .filter(el => el.length > 0);
```
You can avoid creating this variable

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+
+  const formattedObject = {};
+
+  arrayWithoutEmptyStrings.forEach(eachElement => {
```
BudnikOleksii, line 25: You can use `reduce` method to generate this object instead

BudnikOleksii (changes requested):
Good work 👍 
Just a few moments need to be fixed.
Regarding variable names, from the checklist:

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const startConvertion = sourceString
+    .split(';')
+    .map(property => property.trim());
+
+  const removeEmptyStrings = startConvertion.filter(property =>
+    (property.length > 0));
+
+  const formattedStyle = removeEmptyStrings.reduce((acc, currentProperty) => {
```
ArtemTeslenko, line 24: `reduce` allows to put all logic inside itself, try to `split` entry value by `;` and after that use `reduce` with whole logic inside

ArtemTeslenko (approved):
Good job 👍 I am approving it, but see comment and try to refactor your code, it will be good practice for you 😉

PR #2742, opened 2023-11-08
done task

Anton-Kuchmasov:


ilyapgitt (approved):
Awesome!!!!

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const propsDividedToStrings = sourceString.split(';');
+
+  const resultObject = {};
```
BudnikOleksii, line 20: [CODE STYLE] - Avoid variable namings like array , object arr, obj, result. It will decrease code readability

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+
+  propsDividedToStrings.forEach((propertyPair) => {
+    if (propertyPair.includes(': ')) {
+      const propAndValue = propertyPair.split(': ')
+        .map((proper) =>
+          proper.trim());
+
+      resultObject[propAndValue[0]] = propAndValue[1];
+    }
+  });
```
BudnikOleksii, line 30: You can create an object with `reduce` method instead of creating an empty object and adding properties in `forEach`.

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+
 function convertToObject(sourceString) {
-  // write your code here
+  const propsDividedToStrings = sourceString.split(';');
+
+  const resultObject = {};
+
+  propsDividedToStrings.forEach((propertyPair) => {
+    if (propertyPair.includes(': ')) {
+      const propAndValue = propertyPair.split(': ')
```
BudnikOleksii, line 24: What if the property is written without space? `width:100px`

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions, check them

BudnikOleksii (approved):
Nicely done 🚀

PR #2747, opened 2023-11-08
task solution

adrian200002:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssProperties = {};
+
+  const lines = sourceString.split(';');
+  const filteredLines = lines.filter(el => el.trim() !== '');
+
+  for (const css of filteredLines) {
```
mosowiecka, line 22: From checklist: Avoid using loops in this task, use iteration methods instead.

mosowiecka (changes requested):
Also, all chained methods called on the same level should have same indentation.

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+  sourceString
+    .split(';')
+    .filter(el => el.trim() !== '')
+    .forEach(css => {
+      const [code, value] = css.split(':');
+
+      cssProperties[code.trim()] = value.trim();
+    });
+
+  return cssProperties;
```
JoyCoffeeAddict, line 28: That is great! But I believe that you can find a solution using `Array.prototype.reduce` instead of `forEach`

JoyCoffeeAddict (changes requested):
Your code does work, but you can make it cleaner

mvjl000 (approved):
Great!

PR #2748, opened 2023-11-08
Solution 

MarcinLigmann:


src/convertToObject.js
```diff
@@ -14,14 +14,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  const arr = sourceString.split(';').map(el => el.trim())
-    .filter(element => element.length > 0);
-  const array = arr.map(el => el.split(':'));
-  const array1 = array.map(el => el.map(x => x.trim()));
   const object = {};
```
Patrykmclaren14, line 17: You can change the variable name `from` object to `cssStyled`; it looks better.

src/convertToObject.js
```diff
@@ -14,14 +14,17 @@
...
+    .map(el => el.trim())
+    .filter(element => element.length > 0);
+
+  const splitedArray = arrFromString.map(el => el.split(':'));
+
+  const trimmedArray = splitedArray.map(el => el.map(x => x.trim()));
+
+  for (const el of trimmedArray) {
     object[el[0]] = el[1];
   }
```
Patrykmclaren14, line 29: You can also refactor this `for` loop into a `map` function for a more professional appearance: 
```
trimmedArray.map(el => {
   --the same code
  })
```

Patrykmclaren14 (approved):
The code is good and works properly. My comment is just a suggestion based on how I see the solution.

src/convertToObject.js
JoyCoffeeAddict: agreed.

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  const arrFromString = sourceString.split(';')
+    .map(el => el.trim())
+    .filter(element => element.length > 0);
+
+  const splitedArray = arrFromString.map(el => el.split(':'));
+
+  const trimmedArray = splitedArray.map(el => el.map(x => x.trim()));
+
+  for (const el of trimmedArray) {
```
JoyCoffeeAddict, line 27: try deconstructing `el` into an array. `el` is an array consisting of `[key, value]`. You always have two parts in the same order.

JoyCoffeeAddict (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+  const splitedArray = arrFromString.map(el => el.split(':'));
+
+  const trimmedArray = splitedArray.map(el => el.map(x => x.trim()));
+
+  trimmedArray.forEach(el => {
+    const key = el[0];
+    const value = el[1];
+
+    cssStyled[key] = value;
+  });
```
JoyCoffeeAddict, line 32: how about: 
```suggestion
  trimmedArray.forEach(([key, value]) => {
    cssStyled[key] = value;
  });
```

JoyCoffeeAddict (changes requested):

src/convertToObject.js
MarcinLigmann: yes, it looks more coherent.

JoyCoffeeAddict (approved):

PR #2749, opened 2023-11-08
Develop

mosowiecka:


src/convertToObject.js
```diff
@@ -3,16 +3,43 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
-}
+  const stylesObject = {};
```
Timi2424, line 18: Use snake case for this is const naming convention

Timi2424 (approved):
works, and looks good

src/convertToObject.js
JoyCoffeeAddict: We rarely use snake case in JS. It's fine the way it is right now.

src/convertToObject.js
```diff
@@ -3,16 +3,43 @@
...
+
+  sourceString
+    .split(';')
+    .map((element) => element
+      .split(':')
+      .map((element2) => element2.trim())
+    ).forEach((element) => {
+      if (element[0] !== '' && element[0] !== undefined) {
+        stylesObject[element[0]] = element[1];
+      }
```
JoyCoffeeAddict, line 28: You can use destructuring of an `element` into an array here. `element` is an array consisting of two values: `[key, value]`. It will make this code a bit more readable. You can also filter before the `forEach` to check if there are any empty keys

JoyCoffeeAddict (changes requested):

mvjl000 (approved):
Good!

PR #2750, opened 2023-11-08
solution

Timi2424:
have a good day

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -1,18 +1,23 @@
 'use strict';
 
 /**
- * Implement convertToObject function:
- *
- * Function takes a string with styles (see an example in [stylesString.js](./stylesString.js))
- * and returns an object where CSS properties are keys
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const STRING_ARRAY = sourceString.split(';');
+  const STYLES_OBJECT = {};
+
+  STRING_ARRAY.forEach((style) => {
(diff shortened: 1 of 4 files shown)


src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
- *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const STRING_ARRAY = sourceString.split(';');
+  const STYLES_OBJECT = {};
```
Sibunas, line 10: I think it is better to name the variables what they are in the program, what function they perform, and not what their type is.

Sibunas (approved):

src/convertToObject.js
mvjl000, line 10: I kind of agree with previous reviewer.

In general, we only want to name uppercase & snake case  (for example `STRING_ARRAY`) constants that never really change and typically are not a result of some function/method.
Do you remember transportation on vacation task? In there you had a perfect example of such a constants like `const LONG_TERM = 7;` (If you want some more explanation on this, contact me on slack).

So, for those two I would use camel case `stringArray` and `stylesObject`.
Furthermore, it's good to create descriptive and meaningful names.

Let's take a closer look at `STRING_ARRAY`. `sourceString.split(;)` results in an array of 'rows' (I just called it that way so it's easier to imagine), in which we have i.e. `'background-color: red'`.
I think it would be way better to name it `stylesRows` or whatever you think describes it well

mvjl000 (changes requested):

mvjl000 (approved):
Good job!

PR #2752, opened 2023-11-09
Solution

WiktoriaBielec:


web3-sorcerer (approved):
Your solution looks clean and all tests passed. You can change your const to UPPERCASE but apart from this, great work.

src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
   // write your code here
+  const obj = {};
+  const arrStyles = sourceString.split(';');
+
+  for (const style of arrStyles) {
+    const arrStyle = style.split(':');
+    const key = arrStyle[0].trim();
+
+    if (key.length !== 0) {
+      obj[key] = arrStyle[1];
```
JoyCoffeeAddict, line 27: What if you trim here, instead of creating another `for` loop

JoyCoffeeAddict (changes requested):

mvjl000 (approved):
Well done!

PR #2754, opened 2023-11-09
solution

DanyloSemenov:


mosowiecka:
Don't need to make copy before each operation.

mosowiecka (approved):

mvjl000 (approved):
I like this solution 👌

PR #2757, opened 2023-11-09
task finished

Fluebubble:


src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
+          : value;
+
+        return fixedValue.trim();
+      });
+    })
+    .reduce((accumulator, param) => {
+      accumulator[param[0]] = param[1];
+
+      return accumulator;
+    }, objectWithCssParams);
```
ArtemTeslenko, line 39: Good idea to use `reduce`, but you could process every item inside this method and get rid of other methods like

> return sourceString.split(";").reduce((result, style) => {
>     there you get array from the whole string style.split(":");
>     there you trim every pair key = yourPair[0]?.trim();
>     there you trim every pair value = yourPair[1]?.trim();
>     if (key && value) {
>       result[key] = value;
>     }
>     return result;
>   }, {});

ArtemTeslenko (changes requested):
Good job! You can do it in less actions and do not overfill your code by logic, even if you do it not as in my comment 😉

src/convertToObject.js
Fluebubble: ESLint forbids "?" before trim :(

ArtemTeslenko (approved):
Great 🔥

PR #2758, opened 2023-11-10
STO done

Mantserov:
:)

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -12,7 +12,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceObject = sourceString
+    .split(';')
+    .map((el) => el.trim())
+    .filter((el) => el.length > 0)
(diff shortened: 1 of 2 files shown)


ArtemSydun (approved):
gj

src/convertToObject.js
```diff
@@ -12,7 +12,18 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceObject = sourceString
+    .split(';')
+    .map((el) => el.trim())
+    .filter((el) => el.length > 0)
+    .map((el) => el.split(':'))
+    .reduce((acc, curr) => {
+      acc[curr[0].trim()] = curr[1].trim();
```
```suggestion
    .reduce((acc, curr) => {
      const [key, value] = curr;
      acc[key.trim()] = value.trim();
```
You can use destructing like this.
Also, a variable name `el` isn't so good, it could be `property`

BudnikOleksii (approved):
Nicely done 🚀

PR #2763, opened 2023-11-13
Solution

LabPetro:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styledObject = {};
+  const sheets = sourceString.split(';');
+
+  sheets.forEach((sheet) => {
+    const [style, value] = sheet.split(':');
```
witflash, line 21: It's better to create more valid names according to the types. 
For example:
```
const [name, value] = ...
// or
const [styleProperty, value] = ...
```

witflash (approved):
Well done 👍

PR #2764, opened 2023-11-13
solution

Markianek:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objectWithCssProperties = {};
+  const propertiesArray = sourceString.split(';');
+
+  for (const properties of propertiesArray) {
+    const propertie = properties.split(':');
```
nazarmatsevych, line 21:
```suggestion
    const [key, value]= properties.split(':');
```

Use array destructuring

nazarmatsevych (approved):
Good work, left minor comment

PR #2768, opened 2023-11-14
function make styles objekt complete

mixelio:


src/convertToObject.js
```diff
@@ -3,16 +3,54 @@
...
+
+          newWalue += word.slice(firstLetterIndex, lastLetterIndex + 1);
+          item[ind] = newWalue;
+        }
+      );
+      result[item[0]] = item[1];
+    }
+  );
+
+  return result;
```
lerastarynets, line 53: this all could be shortened to
```
const properties = sourceString.split(';');
  const styles = {};

  properties.forEach(property => {

    if (key.trim() && value.trim()) {
      styles[key.trim()] = value.trim();
    }
  });

  return styles;
```

lerastarynets (approved):
You're on the right track!

PR #2780, opened 2023-11-18
Style_to_object

Zava96:


etojeDenys (approved):
looks interesting, but too complex, you can try to rewrite it just using split() two times

PR #2781, opened 2023-11-19
Solution

tania-kuzmenko:


src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssArr = sourceString.replace(/\n/g, '').split(';');
+  const cssObj = {};
+
+  for (let i = 0; i < cssArr.length; i++) {
+    const key = cssArr[i].split(':')[0];
+    const value = cssArr[i].split(':')[1];
```
nazarmatsevych, line 20:
```suggestion
    const [key, value] = cssArr[i].split(':');
```

use array destructuring instead

nazarmatsevych (approved):
Great work, left minor comment

PR #2784, opened 2023-11-20
Solution

ruslan-kasinets:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+    tempArr[i] = tempArr[i].trim();
+
+    if (tempArr[i] !== '') {
+      finalArr.push(tempArr[i]);
+    }
+  }
+
+  for (let i = 0; i < finalArr.length; i += 2) {
+    result[finalArr[i]] = finalArr[i + 1];
+  }
```
etojeDenys, line 31: looks like you can simplify it, and remove the last loop

etojeDenys (approved):
good job

PR #2786, opened 2023-11-20
add_style_to_object

JokerMTM:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertString = {};
+  const editString = sourceString
+    .split(';')
+    .map(value => value.trim())
+    .filter(value => value !== '');
```
Andriy-Shymkiv, line 21: I think it should work like this too:
```suggestion
    .filter(value => value);
```

Andriy-Shymkiv (approved):

PR #2788, opened 2023-11-21
Solution

michal-reszka:


mbruhler (approved):
Approved, well done on using array methods, clever way of using `[property, value] `in reduce

PR #2825, opened 2023-12-08
Solution js style to object

Tvinki228:


src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
-  // write your code here
+  const result = {};
+  const stringWithoutSpaces = sourceString.replace(/\s*;\s*/g, ';');
+
+  const pairs = stringWithoutSpaces.split(';');
+
+  pairs.forEach(item => {
+    let [property, value] = item.split(':');
+
+    if (property !== undefined && value !== undefined) {
```
volodymyr-soltys97, line 25:
```suggestion
    if (property && value) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

nazarmatsevych (approved):
Good work!

PR #2836, opened 2023-12-17
solution

HoTaBu4:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
-  // write your code here
+  const obj = {};
+  const cleaned = sourceString.replace(/\n/g, '');
+  const splited = cleaned.split(';');
+
+  for (let i = 0; i < splited.length; i++) {
+    if (splited[i].trim().length !== 0) {
+      const objEntries = splited[i].split(':');
+
+      obj[objEntries[0].trim()] = objEntries[1].trim();
```
nazarmatsevych, line 25:
```suggestion
      const [key, value]= splited[i].split(':');

      obj[key.trim()] = value.trim();
```

use array destructuring

nazarmatsevych (approved):
Great work, left a minor comment

PR #2837, opened 2023-12-17
task done

Manankin:


src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const arrOfProperties = sourceString.split(';');
+
+  for (const index in arrOfProperties) {
+    let elem = arrOfProperties[index];
```
olya-shyian, line 21: `elem` is not a good name

src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
+
+    while (elem.startsWith('/n')) {
+      elem = elem.slice(0, 2);
+    }
+    arrOfProperties[index] = elem.trim();
+  }
+
+  const validProps = arrOfProperties.filter((prop) => prop.length !== 0);
```
olya-shyian, line 29: you can write more simple like `.filter(prop => prop)` or even do this at the beginning of the loop

```
 for (const prop of properties) {
  if (!prop) {
      continue;
  }
```    
we just skip empty properties

src/convertToObject.js
olya-shyian, line 21: instead of using for...in loop, use for...of

```
 for (const prop of properties) 
 ```
 
 so in `prop` you will get the same `let elem = arrOfProperties[index];`

**Review by olya-shyian — CHANGES_REQUESTED** [review]
Well done!💥

File: src/convertToObject.js (hunk 1 above)
Comment (Manankin) [reply src/convertToObject.js:21]: in this case later I have to find index of prop and change the element in arrOfProperties (it is outside of loop)
so I have decided to use for ... in loop instead of for ... of loop. (line 26 in my code)

File: src/convertToObject.js [hunk 3]
```diff
@@ -17,19 +17,19 @@ function convertToObject(sourceString) {
   const result = {};
   const arrOfProperties = sourceString.split(';');
 
-  for (const index in arrOfProperties) {
-    let elem = arrOfProperties[index];
+  for (let index = 0; index < arrOfProperties.length; index++) {
```
rvoshchylo, line 20: [CODE PRACTICE] - Avoid using loops in this task, use iteration methods instead.
check it for whole code

src/convertToObject.js
```diff
@@ -17,19 +17,19 @@ function convertToObject(sourceString) {
   const result = {};
```
rvoshchylo, line 17: [CODE STYLE] - Avoid variable namings like array , object arr, obj, result. It will decrease code readability
check it for all project

rvoshchylo: READ CHECK LIST

rvoshchylo (changes requested):
Good job Dima
Let's check your code style below

src/convertToObject.js
Manankin: I have talked to Misha exactly about this itteration loop. So it done according his advices.

src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
DanilWeda, line 17: Correct this naming

src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
+
+  validProps.forEach((prop) => {
+    const singleProp = prop.split(':');
+    // remove extra spaces
+    const propName = singleProp[0].trim();
+    const propValue = singleProp[1].trim();
+
+    // add props to object
+    result[propName] = propValue;
+  });
```
DanilWeda, line 39: Try to use one array method that can return an object.

src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const arrOfProperties = sourceString.split(';');
+
+  for (let index = 0; index < arrOfProperties.length; index++) {
```
DanilWeda, line 20: This loop is not required. You can describe array methods that can do this logic
(Now you need to practice the array methods only).

dothearts (changes requested):
Great job. 
Check comments, improves some moments! 
Thanks!

src/convertToObject.js
```diff
@@ -3,16 +3,40 @@
...
+
+    return value.trim();
+  });
+
+  const validProps = allProps.filter((prop) => prop);
+
+  const stylesToCSS = validProps.reduce((styles, prop, index,) => {
+    const singleProp = prop.split(':');
+    const propName = singleProp[0].trim();
+    const propValue = singleProp[1].trim();
```
DanilWeda, line 32: as alternative

```suggestion
    const [propName, propValue] = singleProp;
```

dothearts (approved):
Good and clear! 
Thanks! 
Approved with a small comment!

PR #2839, opened 2023-12-18
solution

pivkopa:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.split(';');
+  const obj = {};
```
ssaviak, line 18: The variable has to mean what it contains, or what it does. A good way to name {arr} and {obj} for example - propertiesArray and stylesObject.

ssaviak (approved):
Good job! Please change the names of your variables:)

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  for (let i = 0; i < arr.length; i++) {
+    if (arr[i].trim()) {
+      const key = arr[i].split(':')[0].trim();
+      const value = arr[i].split(':')[1].trim();
```
DanilWeda, line 23: Alternative variant writing ( example for view )

```suggestion
      const [ key, value ] = arr[i].split(':').trim();
```

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const arr = sourceString.split(';');
+  const obj = {};
+
+  for (let i = 0; i < arr.length; i++) {
```
DanilWeda, line 20: You can try using some array method, for  best practice. Try it!)

src/convertToObject.js
DanilWeda, line 18: Change naming

DanilWeda (changes requested):
Good job! 
Try to correct some moments)
Thanks!

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const properties = sourceString.split(';');
+  const result = {};
+
+  properties.forEach((property) => {
+    if (property.trim()) {
+      if (property.trim()) {
```
DanilWeda, line 22: The same condition, not needed

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const properties = sourceString.split(';');
+  const result = {};
+
+  properties.forEach((property) => {
```
DanilWeda, line 20: Try to use reduce for best practice. Also, you don't need a result object above, you get this result object from reduce method.

DanilWeda (changes requested):
Good! Check improvment comments

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const properties = sourceString.split(';');
+
+  const result = properties.reduce((prev, property) => {
+    if (property.trim()) {
+      const key = property.split(':')[0].trim();
+      const value = property.split(':')[1].trim();
```
DanilWeda, line 22:
```suggestion
      const [key, value] = property.split(':')
```
as alternative

DanilWeda (approved):
Great job! 
Approved! 
Thanks!

PR #2840, opened 2023-12-18
Solution

avakiel:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+  const resultObj = {};
+
+  for (let i = 0; i < sourceArr.length; i++) {
+    if (sourceArr[i].trim()) {
+      const key = sourceArr[i].split(':')[0].trim();
+      const value = sourceArr[i].split(':')[1].trim();
+
+      resultObj[key] = value;
+    }
+  }
```
DanilWeda, line 27: Great realization. Please try to use a method that returns an object.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArr = sourceString.split(';');
+  const resultObj = {};
+
+  for (let i = 0; i < sourceArr.length; i++) {
+    if (sourceArr[i].trim()) {
+      const key = sourceArr[i].split(':')[0].trim();
+      const value = sourceArr[i].split(':')[1].trim();
```
DanilWeda, line 23: as alternative
```suggestion
      const [key, value] = sourceArr[i].split(':');
```

DanilWeda (changes requested):
Great job. 
Check comments for improvement.
Thanks!

olya-shyian (approved):
Great solution!🙂

PR #2841, opened 2023-12-19
add task solution

VazilX:


lerastarynets (approved):
You're on the right track!

PR #2842, opened 2023-12-19
add solution

ssaviak:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const propertiesArray = sourceString.split(';')
+    .map(prop => prop.replace('\n', '').trim())
+    .filter(prop => !!prop);
```
nabokok, line 19: conversion to boolean is unnecessary

nabokok (changes requested):
Good job! Just a small general comment. I think that it would be better to reduce the number of iterations in this function (2-3)

DanilWeda (approved):
Good job! 
Approved!)

PR #2844, opened 2023-12-19
solution

vSainiuk:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+ * and values are the values of related CSS properties
+ * (see an example in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
```
sasholk, line 17: 2. [CODE STYLE] - Avoid variable namings like `array` , `object` `arr`, `obj`, `result`. It will decrease code readability

src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+  const arrayStrings = sourceString
+    .split(';')
+    .map((string) => string.trim())
+    .filter((string) => string);
+
+  for (const string of arrayStrings) {
```
sasholk, line 23: 1. [CODE KNOWLEDGE] - Don't use iteration methods to modify an already existing array/object. (forEach is exception). These methods return a new array, so you should use it.

sasholk (approved):
Nice job!

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+  sourceString
+    .split(';')
+    .map((string) => string.trim())
+    .filter((string) => string)
+    .forEach((string) => {
+      const parts = string.split(':');
+      const [ part1, part2 ] = parts;
+
+      stylesObject[part1.trim()] = part2.trim();
+    });
```
DanilWeda, line 28: Chaining is a good thing, but I recommend rewriting the code so that it is more readable. 
Also, try to use one array method which can return an object.

DanilWeda (changes requested):
Thank you for your work! 
I have left a comment to improve your code.

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const formattedString = sourceString
+    .split(';')
+    .map((string) => string.trim())
+    .filter((string) => string);
+
+  return formattedString.reduce((accumulator, string) => {
+    const parts = string.split(':');
+    const [ part1, part2 ] = parts;
```
olya-shyian, line 24: it would be better to have more meaningful name like `const [key, value] = styleRule`

olya-shyian (changes requested):
Wonderful! Really good work!🙂

olya-shyian (approved):
I love it!🙂

PR #2845, opened 2023-12-19
add solution

nabokok:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArr = sourceString.split(';').map(prop => prop.split(':'));
+  const result = {};
```
vSainiuk, line 18: 2. [CODE STYLE] - Avoid variable namings like array , object arr, obj, result. It will decrease code readability

vSainiuk (approved):
You nailed it!

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArr = sourceString.split(';').map(prop => prop.split(':'));
+  const stylesObj = {};
+
+  sourceArr.forEach(el => {
```
DanilWeda, line 20: Try to use another array method, which can return object)

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    const key = el[0] || 0;
+    const value = el[1] || 0;
```
DanilWeda, line 22: alternative variant 
```suggestion
    const [key, value] = el || []
```

DanilWeda (changes requested):
Great job! 
Check some comment for upgrade function)

YegorKorenkov (approved):
Awesome! Nice solution!

PR #2847, opened 2023-12-19
Convert style string to object

sasholk:


src/convertToObject.js
```diff
@@ -1,18 +1,27 @@
...
+    .split(';')
+    .forEach((element) => {
+      const [property, value] = element.split(':');
+
+      if (property && value) {
+        const normalizedProperty = property.replace('\n', '').trim();
+
+        cssObj[normalizedProperty] = value.trim();
+      }
+    });
```
DanilWeda, line 22: You can use one method that returns an object immediately. 
So you don`t need a declaration above ( 10 line )

DanilWeda (changes requested):
Great! 
Check improvement comment!)

src/convertToObject.js
sasholk: Object.assign?

olya-shyian (approved):
Awesome!🔥🔥🔥

PR #2848, opened 2023-12-19
add tusk solution

rvoshchylo:


artem8746 (approved):
Good job!!

src/convertToObject.js
```diff
@@ -12,7 +13,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssNormalized = {};
+
+  sourceString.split(';').forEach((element) => {
```
olya-shyian, line 18: try to rewrite according to this requirement:

image

src/convertToObject.js
```diff
@@ -12,7 +13,19 @@
...
+    const [property, value] = element.split(':');
+
+    if (property && value) {
+      const normalProperty = property.replace('\n', '').trim();
```
olya-shyian, line 22: I think it would be better to rename from `normalProperty`  to `formattedProperty`

olya-shyian (changes requested):
I like it! Great Job!💥

olya-shyian (approved):
Awesome!💪

PR #2853, opened 2023-12-20
add task solution

artem8746:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const CONTAIN_LETTER_REGEX = /\w/;
+  const SPECIAL_SYMBOLS_REGEX = /\*+/g;
+  const styleObj = {};
+
+  const rules = sourceString
+    .split(';')
+    .filter(rule => rule.match(CONTAIN_LETTER_REGEX))
+    .map(rule => rule.replace(SPECIAL_SYMBOLS_REGEX, '').trim());
```
vpashko, line 24: I think your code will work without this replace in this case, but it can be useful under certain conditions. It's cool that you thought about it.

vpashko (approved):
Great job! Approved!

src/convertToObject.js
DanilWeda, line 24: Chaining is a good thing, but I recommend rewriting the code so that it is more readable. 
Also, try to use one array method which can return an object.

DanilWeda (changes requested):
Great. Check improvement!
Thanks!

olya-shyian (approved):
Nice work!❄️

PR #2856, opened 2023-12-20
add solution

Xilston:


dyarechuk (approved):
Good job! 🔥

src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
+ * and values are the values of related CSS properties
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const KEYVALUE_SEPARATOR = ';';
```
olya-shyian, line 15: I guess it should be `KEY_VALUE_SEPARATOR`

src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
+  const KEY_VALUE_SEPARATOR = ':';
+
+  const keyValueArr = filterStr(sourceString).split(KEYVALUE_SEPARATOR);
+  const resultObj = {};
+
+  keyValueArr.forEach(keyValue => {
+    let [key, value] = keyValue.split(KEY_VALUE_SEPARATOR);
+
+    if (value) {
+      key = filterStr(key); value = clearValue(filterStr(value));
```
olya-shyian, line 25: please move `value` on the next line

src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
+      resultObj[key] = value;
+    }
+  });
+
+  return resultObj;
+}
+
+function filterStr(str) {
```
olya-shyian, line 33: don't use `str` for naming...and try to give more clear name like `removeSpecificCharacters`

src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
+  const CHARS_TO_FILTER = ['\n'];
+  const strArr = String(str).split();
+
+  return strArr.filter(char => !CHARS_TO_FILTER.includes(char)).join().trim();
+}
+
+function clearValue(valueString) {
+  return String(valueString).split(' ').filter(value => value !== '').join(' ');
```
olya-shyian, line 41: I think you can use  this check `.filter(value => value)`

it means if `value === true` than we keep this value (empty string is always false)

src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
+}
+
+function filterStr(str) {
+  const CHARS_TO_FILTER = ['\n'];
+  const strArr = String(str).split();
+
+  return strArr.filter(char => !CHARS_TO_FILTER.includes(char)).join().trim();
+}
+
+function clearValue(valueString) {
```
olya-shyian, line 40: this name is not clear what this function does

maybe it's better to use like `removeExtraSpaces`

olya-shyian (changes requested):
Nice work!❄️

src/convertToObject.js
```diff
@@ -3,16 +3,40 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const keyValueArr = removeSpecificCharacters(sourceString).split(';');
+  const resultObj = {};
+
+  keyValueArr.forEach(keyValue => {
```
DanilWeda, line 18: You can try to rewrite with the array method, which can return an object. 
In this case resultObj on the 16 line is not needed anymore)

DanilWeda (changes requested):
All good! 
Thanks for the request! 
Check small improvements that you need to try)
Thanks!

DanilWeda (approved):
Good job! 
Approved! 
Thanks!

PR #2858, opened 2023-12-20
convertation func

vpashko:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
-  // write your code here
+  const formattedCss = {};
+
+  sourceString
+    .split(';')
+    .forEach(rule => {
+      const [property, value] = rule.split(':');
+
+      if (property && value) {
+        const formattedProperty = removeSpaces(property);
```
artem8746, line 25: i think just trim() would also be ok in this case

artem8746 (approved):
Good job!!!

src/convertToObject.js
olya-shyian: if it works, I agree =)

according to best practices we have 'to avoid making the code unnecessarily complicated'.
if you can use something simpler, then use it

olya-shyian (changes requested):
Nice work! I like it!❄️

PR #2860, opened 2023-12-20
solution cssobj

l0ad1nggg:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resultObj = {};
```
kostrubitskii, line 17: Good job!
Check checklist!
2.[CODE STYLE] - Avoid variable namings like `array` , `object` `arr`, `obj`, `result`. It will decrease code readability

kostrubitskii (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  const strParts = sourceString.split(';');
+
+  strParts.forEach(element => {
+    const [property, value] = element.split(':').map(part => part.trim());
+
+    if (property && value) {
+      cssProperties[property] = value;
+    }
+  });
```
DanilWeda, line 27: Good! 
Try to use another array method that returns an object)

DanilWeda (changes requested):
Good job! 
Check the improvement comment!

olya-shyian (approved):
Awesome!✨

PR #2861, opened 2023-12-20
Done! (try1)

EdenParadise:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleDeclarations = sourceString.split(';');
+  const stylesObject = {};
+
+  styleDeclarations.forEach(declaration => {
```
DanilWeda, line 18: You can try to rewrite with the array method, which can return an object.
In this case stylesObject on the 16 line is not needed anymore)

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    if (declaration.trim() !== '') {
+      const [property, value] = declaration.split(':');
+      const trimmedProperty = property.trim();
+      const trimmedValue = value.trim();
```
DanilWeda, line 22: You can make trim func and use it on 19 and this lines. ( not critical)

DanilWeda (changes requested):
Great job! 
Try to correct your solution with improvement comments!
Thanks!

src/convertToObject.js
```diff
@@ -3,16 +3,27 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';')
+    .map(declaration => declaration.split(':').map(trim))
+    .filter(([property, value]) => property && value)
+    .reduce((stylesObject, [property, value]) => {
+      stylesObject[property] = value;
```
DanilWeda, line 19: Good! 
You can split the code and add more readability for this part.

DanilWeda (approved):
Great!
Approve with small comments! 
Thanks!

PR #2862, opened 2023-12-21
add task solution

MaxKalalashnyk02:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleArray = sourceString.split(';').filter(Boolean);
+
+  const stylesObject = {};
+
+  styleArray.forEach(styleDeclaration => {
+    const [property, value] = styleDeclaration.split(':').map(part =>
```
OleksandrRezanov, line 22: To make code more readable, every chained method write on new line.
Like in checklist example:
```
const cssStyles = styles
 .map(style => style)
 .slice(0, 5)
```

OleksandrRezanov (approved):
Good work!

src/convertToObject.js
YegorKorenkov, line 22: helpful advice, please fix
image

YegorKorenkov (changes requested):

PR #2867, opened 2023-12-21
Solution

Yur1yBubl1k:


src/convertToObject.js
```diff
@@ -1,18 +1,23 @@
...
+      const [property, value] = element.split(':');
+
+      if (property && value) {
+        const formatProperty = property
+          .replace('\n', '')
+          .trim();
+
+        result[formatProperty] = value.trim();
+      }
+    });
```
DanilWeda, line 18: Great realization. Please try to use a method that returns an object.
You can split code for more readability.

DanilWeda (changes requested):
Great! 
Check out a few improvements!

src/convertToObject.js
Yur1yBubl1k: I dont fully understand what i need to do
I have already return an object

SerhiiKirik: use `reduce` method instad `forEach`

SerhiiKirik (changes requested):
Great 🔥 
Let's improve your solution 🧑‍💻

PR #2872, opened 2023-12-25
add solution

YevhenOstrovskiy:


rtkachuk-moon (approved):
👍

src/convertToObject.js
```diff
@@ -12,7 +12,24 @@
...
+
+    cssProperties.forEach(property => {
+      const part = property.split(':');
+      const key = part[0].trim();
+      const value = part[1] ? part[1].trim() : '';
+
+      if (key !== '') {
+        propertiesObject[key] = value;
+      }
+    });
```
DanilWeda, line 29: Try to use a method that returns an object!

src/convertToObject.js
```diff
@@ -12,7 +12,24 @@
...
+  const cssValues = sourceString.split('\n');
+  const propertiesObject = {};
+
+  cssValues.forEach(values => {
+    const cssProperties = values.split(';');
+
+    cssProperties.forEach(property => {
+      const part = property.split(':');
+      const key = part[0].trim();
+      const value = part[1] ? part[1].trim() : '';
```
DanilWeda, line 24: as alternative 
```suggestion
      const [key, value] = part;
```

DanilWeda (changes requested):
Good job! 
Check the test suite and some comments for improvement! 
Thanks!

olya-shyian (approved):
Wow, impressive!🔥

PR #2891, opened 2024-01-07
solution

Diliradon:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const objectCss = sourceString
+    .split(';')
```
l4st1m0za, line 19: Consider checking if the sourceString is not empty or null before splitting, to prevent errors.

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+    .map(item => item.split(':'))
+    .map(item => item.map(value => value.trim()))
```
l4st1m0za, line 21: The second map function is not necessary, consider combining it with the first map function.

l4st1m0za (changes requested):
Good job, some small improvements needed

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  if (!sourceString) {
+    return Error('source string is not empty or null');
```
SanyaBratashchuk, line 19:
```suggestion
    return Error('source string is empty or null');
```

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+  }
+
+  const objectCss = sourceString
+    .split(';')
+    .map(item => item.split(':').map(value => value.trim()))
+    .map(item => item.filter(value => value !== ''))
```
SanyaBratashchuk, line 25:
```suggestion
    .map(item => item.split(':').map(value => value.trim()).filter(trimmedValue => trimmedValue !== ''))
  
```

SanyaBratashchuk (changes requested):

src/convertToObject.js
SanyaBratashchuk: isn't it lead to same result?

SanyaBratashchuk (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
+    return Error('source string is empty or null');
+  }
+
+  const objectCss = sourceString
+    .split(';')
+    .map(item => item
+      .split(':')
+      .map(value => value.trim())
+      .filter(trimedValue => trimedValue !== ''))
+    .map(item => item)
```
SanyaBratashchuk, line 28:
```suggestion
```

SanyaBratashchuk (approved):

PR #2898, opened 2024-01-10
task done

taraShulak:


src/convertToObject.js
```diff
@@ -3,16 +3,39 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const NEW_STYLE = {};
+  const SOU_STR_SPLIT = sourceString.split(';');
+  const STYLE_ARR = [];
```
etojeDenys, line 19: it's not the constants, you don't need to write them in uppercase

etojeDenys (approved):
lgtm

PR #2901, opened 2024-01-12
solution

fessenko-tr:


VitaliyBondarenko1982 (changes requested):
Good job.
Suggest to reduce iterations of array and instead two map and forEach make everything in reduce method.

anastasiiavorobiova (approved):
Great work!

PR #2927, opened 2024-01-24
Solution

ayasinska:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
+
+  rows.forEach((row) => {
+    if (row === '' || !row.includes(':')) {
+      return;
+    }
+
+    const property = row.split(/:|;/)
+      .map(element => element.trim());
+
+    result[property[0]] = property[1];
```
vlasiuksasha, line 28: This could potentially assign an empty value to the result object if there are multiple ':' or ';' in the row. It would be better to ensure the property array has at least two elements before assigning.

vlasiuksasha (changes requested):
check the comments

4ksandrey (approved):
Accepted, well done

PR #2932, opened 2024-01-26
Solution

ksstassy:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
 /**
  * Implement convertToObject function:
  *
```
OleziO, line 5: Remove all comments from the code

OleziO (changes requested):

etojeDenys (approved):
awesome, you can try to simplify it and rewrite using the reduce

PR #2937, opened 2024-01-27
add task solution

ArsenSkandalenko:


src/convertToObject.js
```diff
@@ -3,16 +3,42 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const result = {};
+
+  sourceString
+    .trim()
```
VitaliyBondarenko1982, line 17:
```suggestion

  const splittedString = sourceString.split(';');
  
  return splittedString.reduce((acc, item) => {
    // convert styles logic
  }, {})
```
suggest simplify solution. 
no need to replace a lot of time each cases.
just split by `;` and make all logic in reduce method.
you will need only split and trim methods in reduce callback logic.

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some refactoring.

etojeDenys (changes requested):
image
still not fixed, feel free to ask for some help in the chat

PR #2941, opened 2024-01-30
task solution

Poviakalo:


src/convertToObject.js
```diff
@@ -3,16 +3,53 @@
...
+              return el
+                .split(':')
+                .filter(elWithoutEmptyString => elWithoutEmptyString !== '');
+            })
+            .filter(el => el.length !== 0)
+            .map(el => el[0]);
+        })
+        .filter(el => el.length !== 0);
+    })
+    .filter(el => el.length !== 0)
```
etojeDenys, line 37: looks too complex, you can juse split elements in that array by ":" and set the firstElement.trim() as the key and the secondElement.trim() as the value to the object

src/convertToObject.js
```diff
@@ -3,16 +3,53 @@
...
+
+function createObjectStyle(cssStyles = []) {
+  const objStyles = {};
+
+  cssStyles.forEach(style => {
+    const [key, ...values] = style;
+
+    objStyles[key] = values.join(' ');
+  });
+
```
etojeDenys, line 51: also, you can try to simplify your code without helper function, just using the `reduce` method

etojeDenys (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,26 @@
...
-  // write your code here
+  const cssStyles = sourceString
+    .split(';')
+    .map(style => style.split(':'))
+    .filter(style => style.length === 2);
+
+  return cssStyles.reduce((total, current) => ({
+    ...total,
+    [current[0].trim()]: current[1].trim(),
+  }), {});
```
etojeDenys, line 25:
```suggestion
  return sourceString
    .split(';')
    .map(style => style.split(':'))
    .filter(style => style.length === 2)
    .reduce((total, current) => ({
    ...total,
    [current[0].trim()]: current[1].trim(),
  }), {});
```

etojeDenys (approved):
lgtm

PR #2943, opened 2024-02-02
solution

VasulenkoIllia:
solution

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -12,7 +12,32 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let tmpObj = {};
+  const resultObj = {};
+
+  const arrOfSourceString = sourceString.split(';');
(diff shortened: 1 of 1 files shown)


src/convertToObject.js
```diff
@@ -3,16 +3,43 @@
...
+  const arrOfStrCompleteForObj = [];
+
+  clearArrOfNotEmptyStr.forEach(
+    item => arrOfStrCompleteForObj.push(item.split(',')));
+
+  arrOfStrCompleteForObj.forEach(item => {
+    tmpObj = { [item[0]]: item[1] };
+
+    Object.assign(resultObj, tmpObj);
+  });
```
nazarmatsevych, line 40: It looks quite complicated, think about how to decrease the number of iterations, alternatively, we can use the reduce method to fill the object

nazarmatsevych (approved):
Good work,  please review the comment below.

If you have any questions - feel free to ask them in fe_chat

PR #2945, opened 2024-02-03
Solution

leravlasova:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  return sourceString
+    .split(';')
+    .map((value) => value.trim())
+    .filter((value) => value !== '')
```
VitaliyBondarenko1982, line 21: redundant map and filter iterations. you can make everything in reduce

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some refactoring.

volodymyr-soltys97 (approved):
Nice work 🔥

PR #2948, opened 2024-02-05
solution

Yehorf21:


src/convertToObject.js
```diff
@@ -12,7 +12,23 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  let arr = sourceString.split(/[\n:]+/);
+  arr.shift();
+
+  arr = arr.map((el) => {
+    el = el
+      .trim()
+      .replace(';', '');
```
etojeDenys, line 22: you can just split it by `;`

etojeDenys (changes requested):

src/convertToObject.js
```diff
@@ -12,7 +12,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const obj = {};
+  let arr = sourceString.replace(/\n/g, '').split(';');
+  arr.pop()
+
+  arr = arr.map((el) => el.split(':'))
```
etojeDenys, line 19:
```suggestion
  let arr = sourceString.split(';').map((el) => el.split(':'))
```

etojeDenys (approved):
lgtm

PR #2954, opened 2024-02-06
task solution

BoGoBooM:


etojeDenys (approved):
good job, you can try to simplify it using the reduce

PR #2959, opened 2024-02-08
Develop

Denys209818:


src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const styles = {};
+
+  const strings = sourceString.split('\n').map((value) => {
```
etojeDenys, line 20: we can simplify this logic using the reduce method and splitting by the ';'

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #2962, opened 2024-02-08
Solution

pasha1932:


src/convertToObject.js
```diff
@@ -3,16 +3,49 @@
...
+
+    if (sourceString[j] === ':') {
+      values.push(sourceString.slice(j + 1, endValue).trim());
+    }
+  }
+
+  for (let i = 0; i < keys.length; i++) {
+    stylesCss[keys[i]] = values[values.length - i - 1];
+  }
+
```
etojeDenys, line 47: looks too complicated, try to firstly split it by ';' and then by ':'

etojeDenys (changes requested):

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cleanedString = sourceString.replace('\n', '');
+  const rules = cleanedString.split(';');
+  const filteredRules = rules.filter(el => el.trim() !== '');
```
Andriy-Shymkiv, line 19: should work like this too
```suggestion
  const filteredRules = rules.filter(el => el.trim());
```

Andriy-Shymkiv (approved):

PR #2963, opened 2024-02-09
first time version_1

Beliyandr:


src/convertToObject.js
```diff
@@ -11,8 +11,25 @@
...
+
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArray = sourceString.split(';');
+  const result = {};
+
+  for (let i = 0; i < sourceArray.length; i++) {
+    const element = sourceArray[i].split(':');
+
+    if (element[0] < 1 || element[1].length < 1) {
```
etojeDenys, line 22: `element[0] < 1` looks strange
element[0] is a string, so you check if the string is less than number 1?

etojeDenys (approved):
good job

src/convertToObject.js
Beliyandr: little modified

PR #2965, opened 2024-02-09
solution

Prokopovych16:


etojeDenys (approved):
lgtm, but you can try to rewrite it using the reduce method

PR #2974, opened 2024-02-15
Solution

Rusik-best86:


src/convertToObject.js
```diff
@@ -3,16 +3,36 @@
...
+
+  const resultStylesObj = stylesObj.reduce((prev, item) => {
+    if (!item) {
+      return prev;
+    }
+
+    return {
+      ...prev, ...item,
+    };
+  }, {});
```
BudnikOleksii, line 33: You can do it just with reduce instead of map + reduce

BudnikOleksii (changes requested):
Good work 👍 
Lets improve your solution slightly

BudnikOleksii (approved):
Nicely done 🚀

PR #2977, opened 2024-02-16
convert to object

VadymBryvka:


src/convertToObject.js
```diff
@@ -3,15 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
+  const comand = sourceString.split(';');
+  const result = {};
+
+  for (const i of comand) {
+    let rrr = i.split(':');
```
volodymyr-soltys97, line 21: Use a more descriptive name for variables instead of `rrr`

src/convertToObject.js
```diff
@@ -3,15 +3,30 @@
...
- * and values are the values of related CSS properties (see an exampl in [test file](./convertToObject.test.js))
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
+  const comand = sourceString.split(';');
```
volodymyr-soltys97, line 17: The same here

src/convertToObject.js
```diff
@@ -3,15 +3,30 @@
...
+  const result = {};
+
+  for (const i of comand) {
+    let rrr = i.split(':');
+
+    if (rrr[1] !== undefined) {
```
volodymyr-soltys97, line 23: Instead of explicitly checking for undefined, it is better to check for true

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/convertToObject.js
```diff
@@ -3,15 +3,31 @@
...
  */
 function convertToObject(sourceString) {
+  const styleCommand = sourceString.split(';');
+  const result = {};
+
+  for (const part of styleCommand) {
+    let twoPartsCommand = part.split(':');
+
+    if (twoPartsCommand[1]) {
+      result[twoPartsCommand[0].trimEnd().trimStart()]
```
etojeDenys, line 24:
```suggestion
      result[twoPartsCommand[0].trim()]
```

src/convertToObject.js
```diff
@@ -3,15 +3,31 @@
...
+      = twoPartsCommand[1].trimEnd().trimStart();
```
etojeDenys, line 25: same here

src/convertToObject.js
```diff
@@ -3,15 +3,31 @@
...
+      twoPartsCommand = [];
```
etojeDenys, line 26: looks redundant

etojeDenys (changes requested):

src/convertToObject.js
```diff
@@ -3,15 +3,30 @@
...
 function convertToObject(sourceString) {
+  const styleCommand = sourceString.split(';');
+  const result = {};
+
+  for (const partCommand of styleCommand) {
+    let twoPartsCommand = partCommand.split(':').map((part) => part.trim());
+
+    if (twoPartsCommand[1]) {
+      result[twoPartsCommand[0]] = twoPartsCommand[1];
+      twoPartsCommand = [];
```
etojeDenys, line 25: you can simplify it using the reduce method

src/convertToObject.js
etojeDenys: image

etojeDenys (approved):

PR #2979, opened 2024-02-17
solution

ChrisHryts:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+    .split(';')
+    .filter(item => item.trim() !== '');
+
+  const updatedSourceArray = sourceArray.map(item => {
+    const [ property, value ] = item.split(':');
+
+    return [property.trim(), value.trim()];
+  });
+
+  const sourceObject = {};
```
polosanya, line 28: Let's try to use more meaningful names
image

polosanya (changes requested):
Almost perfect!

PR #2980, opened 2024-02-17
add task solution

valentynavik:


etojeDenys (approved):
lgtm, but you can try to simplify it using the reduce

PR #2983, opened 2024-02-19
Task solution

Nirvacsh:


olduma (approved):
It looks perfect! Great job!)

src/convertToObject.js
```diff
@@ -3,16 +3,34 @@
...
+
+  const propertyValuePairs = trimmedString.map(item => {
+    const [property, value] = item.split(':');
+
+    return [property.trim(), value.trim()];
+  });
+
+  const resultingObject = {};
+
+  propertyValuePairs.map(item => {
```
BudnikOleksii, line 29: - it doesn't make sense to use `map` if you don't use returned from its array
- You can use the reduce method to generate an object from the array. And actually, you can combine the first time you mapped through an array and this one and replace just with the reduce

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution a bit

PR #2984, opened 2024-02-19
Solution js_style-to-object

RuslanSkl:


src/convertToObject.js
```diff
@@ -3,16 +3,32 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+  const cleanString = sourceString.replace(/\/\*.*?\*\//g, '');
+  const matches = cleanString.match(/([\w-]+)\s*:\s*([^;]+)/g);
+
+  if (matches) {
```
etojeDenys, line 21: you can just split it by the `;` instead of this

etojeDenys (changes requested):

lerastarynets (approved):
You've got this, keep pushing forward!

PR #2993, opened 2024-02-21
Solution

VitaliiNez:


etojeDenys (approved):
lgtm, but you can simplify it using the reduce

PR #3002, opened 2024-02-21
Solution

IRONH4MMER:


Vlobdam (approved):

src/convertToObject.js
```diff
@@ -1,18 +1,25 @@
...
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const arr = sourceString
+    .split(';')
+    .map(
+      (rule) => rule.split(':').map((ruleItem) => ruleItem.trim()),
+    )
```
roman-mirzoian, line 15: You can remove excess indents during reduce, so as not to add an extra map in the middle of which is another map and not to load the complexity of the algorithm.

roman-mirzoian (approved):
Well done, but pay attention to the comment 👍

PR #3003, opened 2024-02-21
the object is made

vhrubyi:


etojeDenys (approved):

PR #3004, opened 2024-02-21
Solution

ivan-gritsenko:


vlesia (approved):
Good job. 
It would be better if the variable was called accumulator instead of acc. 
And if a function takes only one argument, you can pass it without wrapping it in parentheses.

BudnikOleksii (changes requested):
Good work 👍 
Great that you've practiced a bit with methods, now you can simplify and get rid of most of the iterations. You can do operations that you did in `map` and `filter` in the `reduce` itself, but it's up to you

BudnikOleksii (approved):
👍

PR #3005, opened 2024-02-21
solved the task

ogerenko:


src/convertToObject.js
```diff
@@ -3,16 +3,35 @@
...
+
+      pairItemValue.push(value);
+    }
+  });
+
+  for (const element of pairItemValue) {
+    for (let i = 0; i < element.length; i++) {
+      objectStyles[element[0]] = element[1];
+    }
+  }
```
etojeDenys, line 32:
```suggestion
  arrayStyle.forEach(element => {
    if (element.trim() !== '') {
      const value = element.split(':').map(a => a.trim());

      objectStyles[value[0]] = value[1];
    }
  });
```

etojeDenys (changes requested):
you can also try to simplify it using the reduce method

ogerenko:
thanks, I don't know  how, but I was trying to do something like that. 
It didn't work.)))

lerastarynets (approved):
lgtm 🚀

PR #3006, opened 2024-02-22
add style to object solution

lbihun:


etojeDenys (approved):
lgtm, you can try to simplify it using the reduce method

PR #3011, opened 2024-02-22
Structured

DenNice-cloud:


Graf6612 (approved):
Good job)

src/convertToObject.js
```diff
@@ -3,16 +3,33 @@
...
+  return sourceString
+    .split(';')
+    .reduce((acc, words) => {
+      const [key, value] = words
+        .split(':')
+        .map((word) => word.trim());
+
+      if (key === '') {
+        return acc;
+      }
```
BudnikOleksii, line 26:
```suggestion
      if (!key) {
        return acc;
      }
```
You don't need to strictly compare with an empty string, just check if a value exists
https://developer.mozilla.org/en-US/docs/Glossary/falsy

BudnikOleksii (approved):
Good job 👍

PR #3013, opened 2024-02-22
Solution

DmitryKot96:


src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
+  const sourceStringArray = sourceString
+    .split(';')
+    .map((item) => item.split(':'));
+
+  for (const item of sourceStringArray) {
+    const [key, value] = item;
+
+    if (value) {
+      styledObject[key.trim()] = value.trim();
+    }
```
ivan-gritsenko, line 27: It would be better to use here the reduce( ) to create the object.

ivan-gritsenko (approved):
Good Solution! 👍
Overall good, but it would be better to use more array methods like reduce(), filter() e.t.c

src/convertToObject.js
BudnikOleksii: > It would be better to use here the reduce( ) to create the object.

Good point, don't use `for of`, this task is for array iteration method practice

BudnikOleksii (changes requested):
Good job 👍 
Let's rewrite your solution with `reduce` method instead

PR #3015, opened 2024-02-22
add solution

tetlisna:
1. [CODE KNOWLEDGE] - Don't use iteration methods to modify an already existing array/object. (forEach is exception). These methods return a new array, so you should use it.
2. [CODE STYLE] - Avoid variable namings like `array` , `object` `arr`, `obj`, `result`. It will decrease code readability
3. [CODE KNOWLEDGE] - Don't need to make copy before each operation. Moreover, methods like `filter` or `map` don't mutate original array.
4. [CODE STYLE] - All chained methods called on the same level should have same indentation. This makes code much more readable.

BAD EXAMPLE:
```
```

GOOD EXAMPLE: 
```
const cssStyles = styles
 .map(style => style)
 .slice(0, 5)
```
5. [CODE PRACTICE] - Avoid using loops in this task, use iteration methods instead.

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -15,20 +15,20 @@
  */
 function convertToObject(sourceString) {
   return sourceString
-    .split('\n')
+    .split(';')
     .reduce((acc, line) => {
       const parts = line
         .split(':')
-          .map((part) => part.trim().replace(';', ''));
+        .map((part) => part.trim());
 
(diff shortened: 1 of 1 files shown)


anastasia1383 (approved):
Looks good to me!

BudnikOleksii (approved):
Good work 👍

PR #3017, opened 2024-02-22
add solution

AntonSakh:


illya-onyshchuk (approved):
LGMT 👍

src/convertToObject.js
```diff
@@ -3,16 +3,31 @@
...
-  // write your code here
+  const obj = {};
+  const styles = sourceString
+    .split(';')
+    .map(element => element.trim())
+    .filter(element => element.length > 0);
+
+  for (const style of styles) {
+    const key = style.split(':').map(attribute => attribute.trim());
+
```
BudnikOleksii, line 26: Use only array iteration methods for this task, also you generate the object with `reduce`

BudnikOleksii (changes requested):
Good work 👍 

PR #3018, opened 2024-02-22
Develop

illya-onyshchuk:


src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .trim()
+    .replace(/;/g, '')
+    .split('\n');
```
Math8585, line 20: I think you can only 
`.split(';')`

Math8585 (approved):

src/convertToObject.js
BudnikOleksii: > I think you can leave only `.split(';')`

Good point

src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
+
+  const convertStyle = styles.reduce((obj, el) => {
```
BudnikOleksii, line 22: obj, el are bad names they aren't informative at all

src/convertToObject.js
```diff
@@ -3,16 +3,37 @@
...
+    if (el.trim() === '') {
+      return obj;
+    }
```
BudnikOleksii, line 25:
```suggestion
    if (!el.trim()) {
      return obj;
    }
```
You don't need to compare with an empty string, just check if a value exists
https://developer.mozilla.org/en-US/docs/Glossary/Falsy

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution

roman-mirzoian (approved):
Well done 👍

PR #3019, opened 2024-02-22
Solution

okarnaukh:


src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceStringArr = sourceString
+    .split(';')
+    .filter(str => str.includes(':') && str.trim().length > 1);
```
dima8528, line 19: The `&& str.trim().length > 1` part is unnessesary. Having ':' symbol, str.length will always be over 1 in any case, so this part of the condition will always be true and it just decreases readability. Generally, looks good to me. Fix this issue and everything is gonna be excellent )

dima8528 (changes requested):

src/convertToObject.js
roman-mirzoian: @dima8528 good point!

src/convertToObject.js
```diff
@@ -3,16 +3,29 @@
...
+  const stylesObj = {};
```
roman-mirzoian, line 20: Obj - is not informative part of naming, it is necessary to choose a more accurate name for the variable

roman-mirzoian (changes requested):

BudnikOleksii (approved):
Good work 👍

PR #3020, opened 2024-02-23
Solution

Math8585:


okarnaukh (approved):
Nice! 
Probably, `stringToObject` is not an obvious variable name.

roman-mirzoian (changes requested):

BudnikOleksii (approved):
Good work 👍

PR #3023, opened 2024-02-24
task done

Oleksandr2108:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const styleArray = sourceString.split(';')
+    .filter((item) => item.trim() !== '');
+
+  const stuleObj = {};
```
hattoriKanto, line 21: I think, you want to have a styleObj variable))

hattoriKanto (approved):
Looks good for me!

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * and values are the values of related CSS properties
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const styleArray = sourceString.split(';')
```
roman-mirzoian, line 18: The binding in the name to the data structure does not say much about the meaning of creating the variable, so you should try to name it more accurately.
```suggestion
  const splittedSource = sourceString.split(';')
```

src/convertToObject.js
roman-mirzoian, line 21: The same regarding naming
```suggestion
  const convertedStyles = {};
```

roman-mirzoian (changes requested):
Almost done!

Moroz-Dmytro (changes requested):
Good work 👍 
But your solution can be improved a bit

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * (see an exampl in [test file](./convertToObject.test.js))
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
   // write your code here
+  const splittedSource = sourceString.split(';').map((part) => part.trim())
+    .filter((item) => item.trim() !== '');
```
```suggestion
    .filter((item) => item.trim());
```
You don't need to compare with an empty string just check if value exists

BudnikOleksii (approved):
Good job 👍

PR #3024, opened 2024-02-24
Add task solution

hattoriKanto:


yaGodya (approved):
Great job! The code looks clean and readable. In my opinion and experience, I can't think of anything else to add. I'm passing it over to the mentor for review.

src/convertToObject.js
```diff
@@ -1,18 +1,20 @@
...
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString.split(";");
+  const stylesCSS = {};
+
+  styles
+    .filter((style) => (style !== "" ? style.trim() : false))
```
BudnikOleksii, line 8:
```suggestion
    .filter((style) => style.trim())
```
You don't ternary operator in this case, and compare with an empty string

src/convertToObject.js
```diff
@@ -1,18 +1,20 @@
...
+    .map((style) => {
```
BudnikOleksii, line 9: - It doesn't make sense to use `map` if you don't need returned from it array
- also, you can generate an object with `reduce`

src/convertToObject.js
```diff
@@ -1,18 +1,20 @@
...
+      const indexEnd = style.indexOf(":");
+      const styleKey = style.slice(0, indexEnd).trim();
+      const styleValue = style.slice(indexEnd + 1).trim();
```
BudnikOleksii, line 12: You don't need `indexOf` and `slice`, you can just split by `:` and you will get style key and value

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution a bit

stassokolov1 (approved):
Looks like a masterpiece, I'm approving it🔥

PR #3028, opened 2024-02-25
Solution

hooroshiy:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  const objectPairs = {};
+
+  splitPairs.forEach(pairs => {
+    const [property, value] = pairs.split(':').map(item => item.trim());
+
+    if (property && value) {
+      objectPairs[property] = value;
+    }
+  });
```
etojeDenys, line 27: good job, but you can try to simplify it using the reduce

etojeDenys (approved):

PR #3039, opened 2024-02-28
Solution

Dmitrel:


src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+ * and values are the values of related CSS properties
  *
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArr = sourceString.split(';');
+  const result = {};
```
RusLan2186, line 16: [CODE STYLE] - Avoid variable namings like array , object arr, obj, result. It will decrease code readability

RusLan2186 (approved):
Good job! 

Please change the variable name.

src/convertToObject.js
BudnikOleksii: > [CODE STYLE] - Avoid variable namings like array , object arr, obj, result. It will decrease code readability

Good point

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+
+  sourceArr.forEach(item => {
```
BudnikOleksii, line 18: You can use `reduce` to generate an object

src/convertToObject.js
```diff
@@ -3,16 +3,30 @@
...
+    const newProperty = item.split(':');
```
```suggestion
    const [key, value] = item.split(':');
```
You can use destructing to get key and value directly

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions to improve your solution

stassokolov1 (approved):
Looks good to me, I'm approving it🔥

PR #3043, opened 2024-02-29
test solution

olehmarushchak:


etojeDenys (approved):

PR #3050, opened 2024-03-01
solution

marinatea:


src/convertToObject.js
pawelkloc, line 1: This code is ok (I mean, the first commit) :)

pawelkloc: You should change the second commit to look like the first ;)

pawelkloc (changes requested):
We use lower camel-case for constants in the function body. Aaa and follow the checklist.md ;)

src/convertToObject.js
marinatea: I changed camel-case, but chane I can't ;(

pawelkloc: You use camel-case when you are inside a function body and capitals outside, e.g.: 
 const DAYS_IN_WEEK = 7; - UPPERCASE + snake_case,
const arrayCopy = [...initialArray]; - camelCase

Ok, now your code looks good to me :)

pawelkloc (approved):
It's ok now.

marinatea:
o, thanks :)

nataliaklonowska (approved):

PR #3053, opened 2024-03-01
Develop

bartKarb:
[Task completed](https://github.com/bartKarb/js_style-to-object/blob/develop/src/convertToObject.js)


src/convertToObject.test.js
choeqq, line 1: please refrain from modifying test files

choeqq (approved):

PR #3060, opened 2024-03-03
solution

Maksym-Atanasov:


anastasiiavorobiova (changes requested):
Great work! Consider using array methods here: reduce, split, map, etc.

src/convertToObject.js
```diff
@@ -6,7 +6,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const letters = sourceString.split(';');
+  const result = letters.reduce((accumulator, value, index) => {
+    const params = value.split(':').map((element) => element.trim());
+
+    accumulator[params[0]] = params[1];
```
volodymyr-soltys97, line 13: You can use destruction for `params`
```suggestion
    const params = value.split(':').map((element) => element.trim());
    
    const [key, value] = params;

    accumulator[key] = value;
```

volodymyr-soltys97 (approved):
Nice work 🔥

PR #3068, opened 2024-03-05
Solution

opalahecha:


etojeDenys (approved):
awesome, you can try to simplify it using the reduce method

PR #3074, opened 2024-03-07
Solution

monikaiengibarian:


etojeDenys (approved):
good job, you can try to simplify it using the reduce method

PR #3077, opened 2024-03-07
add task solution

Xsaites:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssObject = {};
+  const propertiesArray = sourceString.split(';');
+
+  propertiesArray.map((element) => {
```
loralevitska, line 12: Method `map` returns new array. In this case would be better to use `forEach` method

loralevitska (changes requested):
Almost perfect) Let's improve

PR #3081, opened 2024-03-10
add solution

rwlp:


src/convertToObject.js
```diff
@@ -7,6 +7,23 @@
...
+  let vl;
+  const obj = {};
+
+  for (const i of arr) {
+    aux = i.split(':');
+    ky = aux[0];
+    vl = aux[1];
+
+    if (ky && vl) {
+      obj[ky.trim()] = vl.trim();
```
jv-aquino, line 22: Clever!!!!!!!!

src/convertToObject.js
```diff
@@ -7,6 +7,23 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const arr = sourceString.split(';');
+  let aux;
+  let ky;
+  let vl;
```
jv-aquino, line 13: quick note: 
**key** and **val** are generally better naming conventions than this

generally try maintaining the first (or the most meaningful) 3 to 5 letters of the word that you want to shorten, like cellphone -> cel, address -> addr

jv-aquino (approved):
If Picasso used Javascript to paint it would definitely look like this!!! Good job with using modern js features and clever solution overall!!

PR #3084, opened 2024-03-10
task is done

BurchakDmitry:


src/convertToObject.js
```diff
@@ -3,16 +3,49 @@
...
+          }
+
+          if (match && key !== ':') {
+            property += key;
+          }
+
+          if (!match && key !== ':') {
+            value += key;
+          }
+        }
```
etojeDenys, line 42: just split it by ":"

etojeDenys (changes requested):
feel free to ask for some help in the chat

src/convertToObject.js
BurchakDmitry: Actually! Thank u!

etojeDenys (approved):
good job

PR #3096, opened 2024-03-15
Solution

aMtiIV:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
+    .map((property) => property.trim())
+    .filter((property) => property !== '')
+    .map((property) => property.split(':').map((element) => element.trim()))
+    .reduce(
+      (object, property) => ({
+        ...object,
+        [property[0]]: property[1],
+      }),
+      {},
+    );
```
BogdanMaliuta, line 20: you used here 4 itteration methods for 1 string its huge performance impact, try to use here only 1 itteration method

BogdanMaliuta (changes requested):
Its works, but let improve perfomance

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';').reduce((object, property) => {
+    const trimmed = property.trim();
```
BogdanMaliuta, line 10: for what are you trimming here?

BogdanMaliuta (approved):
GJ!
Looks more better, easier to understand and read, also better performance

src/convertToObject.js
aMtiIV: It is for protection against empty property. For example 'color: white;      ; background-color: black;' would be split for 'color: white', '      ', background-color: black' and then the second string would be trimmed to '' and later compared to '' in if statement which would result returning unmodified object.

PR #3097, opened 2024-03-15
 converted the strings into an object with keys and values

IvanPovich:


etojeDenys (approved):

PR #3098, opened 2024-03-15
task complete

boikoua:


src/convertToObject.js
```diff
@@ -6,7 +6,25 @@
...
-  // write your code here
+  const propertyObject = {};
+
+  if (sourceString === '') {
+    return {};
+  }
+
+  if (typeof sourceString !== 'string') {
+    return sourceString;
+  }
```
etojeDenys, line 17: looks redundant
```suggestion

```

etojeDenys (approved):

PR #3099, opened 2024-03-16
js style to object

georgiyevv:


src/convertToObject.js
```diff
@@ -1,12 +1,70 @@
...
 
-/**
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
   // write your code here
+    const separator = ';';
+    const styles =  formatString(sourceString).split(separator);
```
etojeDenys, line 18:
```suggestion
    const styles = sourceString.split(separator);
```
looks redundant

src/convertToObject.js
```diff
@@ -1,12 +1,70 @@
...
+    const object = {};
+
+    for (const style of styles) {
+      const values = style.split(':');
+      const [property, value] = values;
```
etojeDenys, line 23:
```suggestion
      const [property, value]  = style.split(':');
```

etojeDenys (changes requested):

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review

src/convertToObject.js
```diff
@@ -1,12 +1,56 @@
...
+  -ms-user-select: none;
+  user-select: none;
+
+
+  white-space: nowrap;
+  width: auto;
+`;
+
+
+convertToObject(complexStylesString);
```
VitaliyBondarenko1982, line 56: all tests failed because no need to replace module.exports

```suggestion
  module.exports = convertToObject;
```

VitaliyBondarenko1982 (changes requested):
Solution looks good.
Check comment, please,  and try to fix tests.

VitaliyBondarenko1982 (approved):
Well done

PR #3112, opened 2024-03-20
Added solution to task

dima-semeniuk:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const params = sourceString.split(';');
+
+  for (const param of params) {
+    if (param.trim().length > 0) {
+      const keyValue = param.split(':');
```
sarakhmen, line 14: unclear naming. Also, what if there is no `:` symbol in param? Would it be a problem to reference [1] arr element in such case?

sarakhmen (changes requested):
Let's improve your solution

PR #3117, opened 2024-03-21
Solution

valost:


src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
...
+
+    const key = attribute[0];
+    const value = attribute[1];
+
+    if (key !== undefined && value !== undefined) {
+      const keyForObj = key.trim();
+      const valueForObj = value.trim();
+
+      result[keyForObj] = valueForObj;
+    }
```
Andriy-Shymkiv, line 23:
```suggestion
    const [key, value] = attributes[element].split(':');

    if (key && value) {
      result[key.trim()] = value.trim();
    }
```

src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const attributes = sourceString.split(';').filter((attr) => attr.length > 1);
```
Andriy-Shymkiv, line 10:
```suggestion
  const attributes = sourceString.split(';').filter((attr) => attr.length);
```

Andriy-Shymkiv (changes requested):
good job! left a few suggestions

vadiimvooo (approved):
LGTM

PR #3118, opened 2024-03-21
solution

olegobiukh:


src/convertToObject.js
```diff
@@ -1,12 +1,38 @@
...
+        ? value.trim().replace(/\n\s*/g, '\n').replace(/\n/g, ' ')
+        : value;
+
+      trimmedString =
+        trimmedString.length > 80 ? func(trimmedString) : trimmedString;
+
+      obj[newKey] = trimmedString;
+    }
+  }
+
```
etojeDenys, line 34: looks too complicated, just split it by ; then by : and then call the trim to elements and create the object from them

etojeDenys (changes requested):
feel free to ask for some help in the chat

src/convertToObject.js
```diff
@@ -1,12 +1,21 @@
...
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+
+  const arr = sourceString.trim().split(';');
+
+  for (let i = 0; i <= arr.length; i++) {
+    const elem = arr[i];
```
Esceype, line 9: I don't think that you need an index here, so you can use `for of` instead of common `for`

src/convertToObject.js
```diff
@@ -1,12 +1,21 @@
...
+
+    if (elem && elem.trim() && elem.split(':')) {
+      const [key, value] = elem.split(':');
```
Esceype, line 12: Such a check is very unproductive why would you do a split twice? It is better to do it once, save it to a variable and use it in both cases

Esceype (changes requested):
GJ!
Read the following comments below:

src/convertToObject.js
olegobiukh: good idea, will do. Thank you.

etojeDenys (approved):
lgtm

PR #3126, opened 2024-03-25
Solution

keex1k:


BogdanMaliuta (changes requested):
Hello, its workks, but not good idea to use a lot of iteration methods, performance.
And after that you are using loop, u can make more things in loop

keex1k:
so what I have to correct? could U give me some tips?

yevhenii-pyl (changes requested):
Clean method chaining, but sometimes it's really worth thinking of performance and readability. It works, but let's try to look from the different angle:

So you have started by splitting your sourceString by` ;` => that's a good idea, you need to start somewhere, lets assign it to some variable. 

By using this method you already have an array of style rules, like `['color: red', ...]`.

What if you just iterate through this new array, split each item by `:` => receive a `[key, value]` pair of each style and just do something like this:

```
    if (key && value) {
      finalResut[key] = value.trim();
    }
```

This should pretty much do the job and reduce the amount of methods used as well as improve code readability.

src/convertToObject.js
```diff
@@ -7,6 +7,17 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const result = {};
+
+  const splitedTab = sourceString.split(';').filter(pom => /\S/.test(pom));
+  const doubleSplitedTab = splitedTab.map(el => el.split(':'));
```
BogdanMaliuta, line 13: you dont need map, do this split in your loop

src/convertToObject.js
```diff
@@ -7,6 +7,17 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const result = {};
+
+  const splitedTab = sourceString.split(';').filter(pom => /\S/.test(pom));
```
BogdanMaliuta, line 12: you dont need this filter, just add condition in your loop, like if (key.trim()) {}

BogdanMaliuta (approved):
Its beeter but,needs to be changed

PR #3132, opened 2024-03-27
add task solution

AlexandeeerG:
Convert Style String to Object
Implement a convertToObject function that takes a string with styles (see an example in stylesString.js and returns an object where CSS declarations are converted to keys and values

[CODE KNOWLEDGE] - Don't use iteration methods to modify an already existing array/object. (forEach is exception). These methods return a new array, so you should use it.
[CODE KNOWLEDGE] - Don't need to make copy before each operation. Moreover, methods like filter or map don't mutate original array.
BAD EXAMPLE:

GOOD EXAMPLE:

const cssStyles = styles
 .map(style => style)
 .slice(0, 5)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -124,9 +124,9 @@
       }
     },
     "@babel/eslint-parser": {
-      "version": "7.23.10",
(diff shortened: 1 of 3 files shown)


l4st1m0za (approved):
Good job!

PR #3134, opened 2024-03-28
task solution using forEach

MatheusStefan:


pedro-ruas (approved):
Great code! Very good use of array methods.
As an extra test, you could also try using a .reduce to solve it.

PR #3141, opened 2024-03-30
Solution

kinqbert:


src/convertToObject.js
```diff
@@ -6,7 +6,69 @@
...
+  const result = {};
+
+  // splitting every property for separate string
+  const props = sourceString.split(';');
+
+  // removing breaking line symbols
+  for (const i in props) {
+    props[i] = props[i].split('\n').join('');
+    props[i] = props[i].split('\t').join('');
+  }
```
alexander-ignatow, line 18: Is that necessary?

alexander-ignatow (changes requested):
Good, but let's improve your code

The main issue is that you are using too many loops one by one, while it would be more effective to create one single loop and perform all operations in one pass

```
 // splitting every property for separate string
  const props = sourceString.split(';');

  // removing breaking line symbols
  for (const i in props) {
   // split on ':'
   // cleanup
   // etc.
  }

```

kinqbert:
thank you for your review! i've improved my task solution and made it much shorter according to your first comment

alexander-ignatow (approved):
Now all is much cleaner 👍

PR #3143, opened 2024-04-01
Solution

ropanefarias:


src/convertToObject.js
```diff
@@ -6,7 +6,45 @@
...
+        value = lines
+          .map((line, index) => {
+            if (index === 0) {
+              return line;
+            } else {
+              return `          ${line}`;
+            }
+          })
+          .join(',\n');
+      }
```
EdPirro, line 34: Why split values by "," just to join them by "," in the end? If you whish to remove the space from the beginning and the end of the string simply using `.trim` in value should suffice...

This entire `if` block is unnecessary, can you see why?

EdPirro (changes requested):
Great job, @ropanefarias 👏! I added just one comment that I'd like you to go through, please ask me if you have any questions about it.

Also, as a personal challenge I'd like to see you implement this solution using `.reduce` instead of `.forEach` Can you come up with such solution?
**This is an extra challenge, if you don't wish to take it on just fix the pointed out issue and request review, this is not necessary to be approved**

src/convertToObject.js
ropanefarias: I guess I ended up getting confused and, in trying to pass the tests, I made simple things complicated.

DarkMistyRoom (approved):
Well done!

PR #3144, opened 2024-04-02
Sonic

yurii-filiushkin:


src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const workString = sourceString
+    .split(';')
+    .map((x) => x.trim())
+    .filter((x) => x !== '');
+
+  const result = {};
+
+  for (const x of workString) {
```
mrkovaliv, line 16: we can impove this solution, better use forEach
```
  declarations.forEach(declaration => {
    const [key, value] = declaration.split(':').map(part => part.trim());
    result[key] = value;
  });
```

mrkovaliv (approved):
Nice work 👍 

Added some improvements)))

PR #3145, opened 2024-04-02
adding solution

vinasxz:


EdPirro (approved):
Great job, @vinasxz! Keep it up 🚀!

As an additional challenge, could you come up with an implementation using `.reduce` instead of a `for...of`?

PR #3147, opened 2024-04-02
add solution

ordila:


src/convertToObject.js
```diff
@@ -5,8 +5,18 @@
  *
  * @return {object}
  */
-function convertToObject(sourceString) {
-  // write your code here
+
+function convertToObject(stylesString) {
+  const styleObject = {};
+  const declarations = stylesString.split(';').filter(Boolean);
```
alexander-ignatow, line 11: interesting move, but this only filters out just empty strings (' ' will pass).
If you start filtering and validating, it makes sense to check, that each substring has one `:` symbol inside that go not on the first or last position

alexander-ignatow (approved):
Good 👍 

Please check my comment

PR #3154, opened 2024-04-04
Solution

JohnyJDX:


src/convertToObject.js
```diff
@@ -6,7 +6,22 @@
...
+
+  const rows = sourceString
+    .trim()
+    .split(';')
+    .map((e) => e.split(':'));
+
+  rows.forEach((row) => {
+    const [property, value] = row;
+
+    if (property && value) {
```
mrkovaliv, line 19: hmmm but we can have property = isBoolean and value = false and it won't work

Probably improve this moment but it isn't critical

mrkovaliv (approved):
Nice work 👍 👍 👍 

Added some small impovements)))

PR #3155, opened 2024-04-04
solution

GrigoryyKach:


src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylePairs = sourceString
```
mrkovaliv, line 9: Better:

```
    .trim()
    .split(';')
    .map(pair => pair.split(':'));
```

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+    .split(';')
+    .map((pair) => pair.trim())
+    .filter((pair) => pair !== '');
+
+  const styleObj = {};
+
+  stylePairs.forEach((pair) => {
+    const [property, value] = pair.split(':').map((part) => part.trim());
```
mrkovaliv, line 17: Just:
`const [property, value] = pair.map(part => part.trim());`

mrkovaliv (changes requested):
Nice work 👍 

Added some improvements)))

mrkovaliv (approved):
🚀 🚀 🚀

PR #3158, opened 2024-04-04
done task

yanamomot:


src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+
+  toArray.forEach((part) => {
+    if (part.trim() !== '') {
+      const [property, value] = part.split(':').map((remove) => remove.trim());
+
+      if (property && value) {
+        newObject[property] = value;
+      }
+    }
+  });
```
DanilWeda, line 20: You can try to use reduce. In this case, you don't need initialize newObject on 10 line (**optional**)

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const toArray = sourceString.split(';');
+  const newObject = {};
+
+  toArray.forEach((part) => {
+    if (part.trim() !== '') {
+      const [property, value] = part.split(':').map((remove) => remove.trim());
```
DanilWeda, line 14: Try to create a constants, chaining is not a good idea for good readability

DanilWeda (changes requested):
Good. 
Check some small comments. 
Thanks!

yanamomot:
Sorry, I try doing task with reduce, but I can't :(

alexander-ignatow (approved):
Well, it works, so let's go with this for now

You'll have an opportunity to master `reduce` later

PR #3159, opened 2024-04-04
add task solution

PrytulaAlexandr:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertStyles = {};
+
+  const stylesString = sourceString.split(';');
```
mrkovaliv, line 11: You can improve your code, try don't create additional variables and instead of `forEach` use `reduce`

mrkovaliv: Example:

```
     const cssStyles = sourceString
         .trim()
         .split(';')
         .reduce((acc, style) => {some code....}
         
     return cssStyles;    
```

mrkovaliv (changes requested):
Nice work 👍 👍 👍 

Added some improvements)))

src/convertToObject.js
PrytulaAlexandr: Hello Volodymyr, added changes.

mrkovaliv (approved):

PR #3160, opened 2024-04-04
add solution

AYupko:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssStyles = sourceString
+    .split(';')
```
mrkovaliv, line 10: better:

```
    .trim()
    .split(';')
```

mrkovaliv (approved):
Nice work 👍 

Added small comment))

PR #3163, opened 2024-04-04
Solution

luyx0:


src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  // string.split.filter => 1 object[key, value]
+  const styles = {};
+  const splitedString = sourceString.split(';').filter(Boolean);
+
+  for (const result of splitedString) {
+    const [property, value] = result.split(':').map((str) => str.trim());
```
DanilWeda, line 14: Try to create a constants, chaining is not a good idea for good readability

src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  // string.split.filter => 1 object[key, value]
+  const styles = {};
+  const splitedString = sourceString.split(';').filter(Boolean);
+
+  for (const result of splitedString) {
```
DanilWeda, line 13: You can try to use reduce. In this case, you don't need initialize styles on 10 line (optional)

DanilWeda (changes requested):
Great! 
Check small comments!
Thanks!

alexander-ignatow (approved):
Nice 👍

PR #3164, opened 2024-04-04
Solution

SlawikH:
Let me know if the code needs improvement

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObj = sourceString
+    .split(';')
+    .map((style) => style.split(':'))
+    .filter((style) => style.length > 1)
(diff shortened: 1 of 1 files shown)


src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObj = sourceString
+    .split(';')
+    .map((style) => style.split(':'))
+    .filter((style) => style.length > 1)
+    .reduce((obj, [key, value]) => {
+      obj[key.trim()] = value.trim();
```
DanilWeda, line 14: All good, but try to create a constants, chaining is not a good idea for good readability

DanilWeda (changes requested):
Good job. 
Please check a small comments.
Thanks!

DanilWeda (approved):
Great.
Approved.

PR #3165, opened 2024-04-04
Solution

MarchAlbion:


src/convertToObject.js
```diff
@@ -5,8 +5,29 @@
...
+  const splittedString = trimmedString.split(';');
+  const removeDuplicated = splittedString.filter(
+    (element) => element.length > 2,
+  );
+
+  const removeSpaces = removeDuplicated.map((element) => {
+    return element.trim().split(':');
+  });
+
+  const removeEmptyString = removeSpaces.filter(
```
mrkovaliv, line 22: you don't need this code

src/convertToObject.js
```diff
@@ -5,8 +5,29 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
   // write your code here
+  const trimmedString = sourceString.trim();
```
mrkovaliv, line 11: you don't need additional variables

Example:
```

const result = sourceString
      .trim()
      .split(';')
      .reduce((acc, style) => {
        const [styleName, styleValue] = style.split(':');

        acc[styleName.trim()] = styleValue.trim();

      return acc;
    }, {});

```

src/convertToObject.js
```diff
@@ -5,8 +5,29 @@
  *
  * @return {object}
  */
+
```
mrkovaliv, line 8: remove unnecessary empty line

mrkovaliv (changes requested):
Nice work 👍 

Added some improvements))

mrkovaliv (approved):

PR #3166, opened 2024-04-04
Solution

TetianaVeremchuk:


src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  sourceString.split(';').forEach((style) => {
+    const [property, value] = style.split(':').map((part) => part.trim());
```
DanilWeda, line 12: Try to create constants, chaining is not a good idea for good readability

src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = {};
+
+  sourceString.split(';').forEach((style) => {
```
DanilWeda, line 11: You can try to use reduce. In this case, you don't need initialize stylesObject on 10 line (optional)

DanilWeda (changes requested):
Good job.
Check small comments.
Thanks!

src/convertToObject.js
TetianaVeremchuk: Created)

src/convertToObject.js
TetianaVeremchuk: Edited) Thank you!

PR #3169, opened 2024-04-06
add task solution

Follder:


src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+
+  sourceString
+    .trim()
```
alexander-ignatow, line 12: You are trimming the results of the first `split` call - so you can skip this

src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
...
+    .split(';')
+    .map((item) => item.trim().split(':'))
+    .map((item) => item.map((mapitem) => mapitem.trim()))
+    .filter((item) => item && item.length > 1)
```
alexander-ignatow, line 16: I would combine these checks under the `foreach` call

The reason is, that each `map` and `filter` call is iterating the whole collection

Looks like you are iterating more times than needed

alexander-ignatow (commented):
Overall - good job 👍 

But please - check my comments

alexander-ignatow (approved):

Follder:
Thanks! 
I`ll be fix it

PR #3170, opened 2024-04-07
add solution

D-Kondratiev:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString.split(';');
+  const CssStyles = {};
```
alexander-ignatow, line 10: Please use correct naming - constants should be named in `camelCase`

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+
+  styles
+    .filter((style) => style.trim())
+    .forEach((style) => {
+      const styleKey = style.split(':')[0];
+      const styleValue = style.split(':')[1];
```
alexander-ignatow, line 16: You performing `split` operation here two times - please fix to the single call

alexander-ignatow (changes requested):
Good, but please, fix the issues

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+
+  styles
+    .filter((style) => style.trim())
+    .forEach((style) => {
+      const parts = style.split(':');
+      const styleKey = parts[0];
+      const styleValue = parts[1];
+
+      cssStyles[styleKey.trim()] = styleValue.trim();
+    });
```
DanilWeda, line 20: Everything is good, but chaining contributes to the worse readability of the code. Try splitting the code into constants.
Also, you can use the reduce method, so you don't need to create cssStyles object.

DanilWeda (changes requested):
Good job. 
Check some improvements. 
Thanks)

alexander-ignatow (approved):
👍

PR #3171, opened 2024-04-07
Solution

OTsiolkovskiy:


src/convertToObject.js
```diff
@@ -6,7 +6,16 @@
...
-  // write your code here
+  const sourceArr = sourceString.split(';').filter((elem) => elem.length !== 0);
+  const sourceObj = {};
+
+  sourceArr.forEach((element) => {
+    const [propety, value] = element.split(':').map((name) => name.trim());
+
+    sourceObj[propety] = value;
+  });
+
```
alexander-ignatow, line 17: Reduce could be a better option here
But hit solution is also ok

alexander-ignatow (approved):
Good! 👍

PR #3173, opened 2024-04-07
Solution

lmmaarriiaa:


src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
...
-  // write your code here
+  const styles = sourceString
+    .trim()
+    .split(';')
+    .filter((style) => style !== '');
+
+  const stylesObject = styles.reduce((acc, style) => {
+    const semicolonIndex = style.indexOf(':');
+    const property = style.slice(0, semicolonIndex).trim();
+    const value = style.slice(semicolonIndex + 1).trim();
```
alexander-ignatow, line 17: 1. Why not use `split` here?
2. What if `semicolonIndex` === -1 ?

src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .trim()
```
alexander-ignatow, line 10: Are you sure you need this? You are trimming again within `reduce` so probably this one is redundand

src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
...
+    .split(';')
+    .filter((style) => style !== '');
```
alexander-ignatow, line 12: You can include this check inside the `reduce` call, to reduce the number of iterations

alexander-ignatow (changes requested):
Good, but try improving your solution a bit

src/convertToObject.js
lmmaarriiaa: Thank you!
Changed my solution a little, hope it is better now :)

alexander-ignatow (approved):
Looks good 👍

PR #3175, opened 2024-04-07
add solution

stepanyukvova200:


src/convertToObject.js
```diff
@@ -1,12 +1,26 @@
...
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const linesOfString = sourceString
+    .split(';')
+    .filter((line) => {
+      return line.includes(':');
+    })
+    .map((line) => {
+      const lineProperties = line.split(':').map((property, propertyIndex) => {
```
alexander-ignatow, line 10: `propertyIndex` is an unused variable

alexander-ignatow (changes requested):
Good, but please improve your solution a bit

src/convertToObject.js
```diff
@@ -1,12 +1,27 @@
...
+    .reduce((accumulator, line) => {
+      if (line.includes(':')) {
+        const lineProperties = line
+          .split(':')
+          .map((property) => property.trim());
+
+        accumulator.push(lineProperties);
+      }
+
+      return accumulator;
```
DanilWeda, line 15: Everything is good, but chaining contributes to the worse readability of the code. Try splitting the code into constants.
Also, you can try to use one reduce method.

DanilWeda (changes requested):
Good job.
Check a small comment for improvements.
 Thanks.

mrkovaliv (approved):

PR #3178, opened 2024-04-08
Solution

yuliaPel:


src/convertToObject.js
```diff
@@ -4,9 +4,26 @@
...
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const styleObject = {};
+
+  sourceString
+    .split(';')
+    .map(removeSpaces)
+    .filter((el) => el.length > 0)
+    .reduce((obj, el) => {
```
polosanya, line 18: image

src/convertToObject.js
```diff
@@ -4,9 +4,26 @@
...
+      const propStyle = el.split(':').map(removeSpaces);
```
polosanya, line 19: image

polosanya (changes requested):
Almost done! Let's improve your solution a bit according to the checklist)

roman-mirzoian (approved):
Well done!

PR #3179, opened 2024-04-08
solution

Natallia3105:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesArray = sourceString.split(';');
+  const stylesObject = {};
+
+  stylesArray.forEach((style) => {
+    if (style.trim() !== '') {
+      const [key, value] = style.split(':').map((item) => item.trim());
+
+      stylesObject[key] = value.includes(' ') ? `${value}` : value;
```
etojeDenys, line 16:
```suggestion
      stylesObject[key] = value;
```
looks redundant for me

etojeDenys (approved):

PR #3182, opened 2024-04-09
Solution

dyakym:


src/convertToObject.js
```diff
@@ -6,7 +6,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssStyles = sourceString.split(';').reduce((acc, style) => {
```
mrkovaliv, line 9: We can improve it a little bit:
```
    .split(';')
    .map((styleStr) => styleStr.trim())
    .filter(([property, value]) => property && value)
    .reduce((acc, style) => {
      const [property, propertyValue] = style.split(':');
```

mrkovaliv (changes requested):
Nice work but we there are some improvements)))

alexander-ignatow (approved):
👍

mrkovaliv (approved):

PR #3190, opened 2024-04-10
Solution

Illia0308:


src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  sourceString
+    .split(';')
+    .map((el) => el.trim())
+    .map((el) => el.split(':'))
```
etojeDenys, line 14:
```suggestion
    .map((el) => el.trim().split(':'))
```

etojeDenys (approved):

PR #3191, opened 2024-04-11
Solution

Andrew77788:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
+  const str = sourceString.split(';');
+  const result = {};
+
+  str.map((element) => {
+    if (element.length >= 2) {
+      const splitElement = element.split(':').map((item) => item.trim());
+
+      result[splitElement[0]] = splitElement[1];
+    }
+  });
```
anastasiia-tilikina, line 18: it is not a proper usage for map method. Map should return some array. Use either forEach of reduce method here

anastasiia-tilikina (changes requested):
Let's fix some issue here

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const str = sourceString.split(';');
+  const result = {};
+
+  str.forEach((element) => {
+    if (element.length >= 2) {
+      const splitElement = element.split(':').map((item) => item.trim());
+
+      result[splitElement[0]] = splitElement[1];
```
```suggestion
      const [property, value] = element.split(':').map((item) => item.trim());

      result[property] = value;
```

volodymyr-soltys97 (approved):
Good job 👍

PR #3207, opened 2024-04-20
add solution

mikezhylka:
hello! do not know if 

20 LINE - if (typeof dividedLine[1] !== 'undefined') 

is a good practice, but code seems to be working xD

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -6,7 +6,30 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedObject = {};
+  const lines = sourceString
+    .split('')
+    .filter((val) => val !== '\n')
+    .join('')
+    .split(';');
+
+  lines.forEach((word) => {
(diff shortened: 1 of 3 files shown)


src/convertToObject.js
```diff
@@ -6,7 +6,30 @@
...
+
+    if (typeof dividedLine[1] !== 'undefined') {
+      const key = dividedLine[0].trimStart().trimEnd();
+      let value = dividedLine[1].trimStart().trimEnd();
+
+      if (value.includes(',')) {
+        value = value.split(',').join(',\n');
+      }
+
+      convertedObject[key] = value;
```
nazarmatsevych, line 28:
```suggestion
    const [ key.trim(), value.trim() ]= cleanWord.split(':');

    if (value) { // means if it presents or it's NOT undefined
      if (value.includes(',')) {
        value = value.split(',').join(',\n');
      }

      convertedObject[key] = value;
```
We can simplify it using array destructuring

nazarmatsevych (approved):
Good work, please check the comment below

PR #3211, opened 2024-04-22
result

BHIgor:


src/convertToObject.js
```diff
@@ -6,7 +6,32 @@
...
+      arr2.push(e.split(':'));
+    }
+  });
+
+  const arrFlat = [];
+
+  arr2.flat().map((e) => {
+    const item = e.replace('\n', '  ');
+
+    if (item.length !== 0) {
```
sTorba24, line 23:
```suggestion
    if (item.length) {
```

src/convertToObject.js
```diff
@@ -6,7 +6,32 @@
...
+
+  arr.map((e) => {
+    if (e.length > 4) {
+      arr2.push(e.split(':'));
+    }
+  });
+
+  const arrFlat = [];
+
+  arr2.flat().map((e) => {
```
sTorba24, line 20:
```suggestion
  arr2.flat().map((e) => {
```

do you really need to use `flat`?

sTorba24 (approved):
Good work! Check my comments

PR #3217, opened 2024-04-23
added solution

syavaYki:


etojeDenys (approved):
LGTM, but you can simplify it just by using the reduce method

PR #3218, opened 2024-04-23
add solution

oleksandr-kovalchuk:


etojeDenys (approved):
lgtm, but you can simplify it by using the reduce method

PR #3229, opened 2024-04-29
task solution

Sywuxy:


src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const newObj = {};
+
+  const splitted = sourceString.split(';');
+
+  splitted.forEach((element) => {
+    const splittedI = element.split(':').map((str) => str.trim());
+
+    const [selector, value] = splittedI;
```
```suggestion
    const [property, value] = splittedI;
```

src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
...
+
+    if (splittedI.length > 1) {
+      newObj[`${selector}`] = value;
```
volodymyr-soltys97, line 19:
```suggestion
      newObj[property] = value;
```

src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
...
+    }
+
+    return newObj;
```
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
...
+  const splitted = sourceString.split(';');
+
+  splitted.forEach((element) => {
+    const splittedI = element.split(':').map((str) => str.trim());
+
+    const [property, value] = splittedI;
+
+    if (splittedI.length > 1) {
+      newObj[property] = value;
+    }
```
volodymyr-soltys97, line 20:
```suggestion
    const [property, value] = element.split(':').map((str) => str.trim());

    newObj[property] = value;
```

PR #3239, opened 2024-05-03
add solution

Alenalenk:


src/convertToObject.js
```diff
@@ -6,7 +6,43 @@
...
+      const value = declarations[1];
+
+      result[key] = value;
+    }
+  });
+
+  function deleteBreak(item) {
+    let start = 0;
+    let end = 0;
+    const condition = 'abcdefghijklmnoprstuvwxyz0123456789-#%()';
```
VitaliyBondarenko1982, line 27: it is looks like as hardcoded string and it is possible to avoid it and such long logic in deleteBreak
should be enough trim declaration and then trim key and value.

VitaliyBondarenko1982 (changes requested):
Good job.
Try to simplify your solution.
After split by `;` it is possible to make everything in reduce method.
use trim() method for normalize strings.

nazarmatsevych (approved):
🔥

PR #3248, opened 2024-05-09
Solution

Pavlo-lab-max-Kopalko:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+  const arrStyles = sourceString.split(';');
+
+  const objStyles = {};
+
+  for (const style of arrStyles) {
+    const nameStyle = style.split(':');
+
+    if (nameStyle[0].trim() !== '') {
+      objStyles[nameStyle[0].trim()] = nameStyle[1].trim();
+    }
```
volodymyr-soltys97, line 18: You can simplify this logic like this
```suggestion
  const [key, value] = style.split(':').map(item => item.trim());

  if (key) {
    objStyles[key] = value;
  }
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

etojeDenys (approved):

PR #3249, opened 2024-05-09
Solution

yaros-dev:


src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const objStyle = sourceString.split(';');
+
+  return objStyle.reduce((acc, item) => {
+    const prop = item.split(':');
```
nazarmatsevych, line 12:
```suggestion
    const [key, value] = item.split(':');
```

You can use array destructuring

nazarmatsevych (approved):
Good work!
Please check the comment below

PR #3253, opened 2024-05-11
added solution

PieCyberPie:


src/convertToObject.js
```diff
@@ -6,7 +6,29 @@
...
+    temp = lines[0].split(':');
+    result[temp[0].trim()] = temp[1].trim();
+
+    return result;
+  }
+
+  for (let i = 0; i < lines.length; i++) {
+    temp = lines[i].split(':');
+    result[temp[0].trim()] = temp[1].trim();
+  }
```
volodymyr-soltys97, line 29: This logic is duplicates now, you need to simplify it and use the forEach method instead if for

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
...
+  const result = {};
+
+  if (!sourceString) {
+    return {};
+  }
+
+  lines.forEach((line) => {
+    const temp = line.split(':');
+
+    result[temp[0].trim()] = temp[1].trim();
```
```suggestion
    const [key, value] = line.split(':').map((el) => el.trim());

    result[key] = value;
```

volodymyr-soltys97 (approved):
Good job 👍

PR #3256, opened 2024-05-14
add task solution

romazh1988:


src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
...
+
+  for (const declaration of declarations) {
+    const parts = declaration.split(':');
+
+    const key = parts[0].trim();
+    const value = parts[1] ? parts[1].trim() : '';
+
+    if (key && value) {
+      styleObject[key] = value;
+    }
```
volodymyr-soltys97, line 20: You can simplify your code like this
```suggestion
    const [key, value] = declaration.split(':').map((el) => el.trim());

    if (key && value) {
      styleObject[key] = value;
    }
```

volodymyr-soltys97 (approved):
Good job 👍

src/convertToObject.js
romazh1988: thank you

PR #3258, opened 2024-05-15
Solution

BagietaCyganeta:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylish = sourceString
+    .split(';')
+    .filter((things) => things.trim() !== '');
+
+  const result = {};
+
+  for (const thing of stylish) {
```
yevhenii-pyl, line 15: I would rather call it 'styleRule' or something instead of a 'thing' but it's not a big deal here

yevhenii-pyl (approved):
Clean and readable, good job! 👍

PR #3260, opened 2024-05-16
create converter to object function

kozlovoleksii:


etojeDenys (changes requested):

vadiimvooo (approved):
Well done!)

PR #3267, opened 2024-05-20
Solution

ashatskyy:


etojeDenys (approved):
you can simplify it using the reduce method, but looks good in general

PR #3272, opened 2024-05-21
Solution / JavaScript Advanced / Array Iteration Methods Usage

polinabs:
Function that takes a string with styles and returns an object where CSS declarations are converted to keys and values.


etojeDenys (approved):

PR #3287, opened 2024-05-29
add task solution

Stasandrushchak:


roman-mirzoian (changes requested):

src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
...
+
+    if (trimmedLine === '' || !trimmedLine.includes(':')) {
+      return;
+    }
 
+    const [property, value] = trimmedLine.split(':').map((part) => part.trim());
+
+    if (property && value) {
+      cssProperties[property.trim()] = value.trim();
+    }
```
roman-mirzoian, line 23: Using ``trim()`` in `map` already removes extra spaces, so an additional call to trim() for property and value is unnecessary.

Example:

```suggestion
    if (property && value) {
      cssProperties[property] = value;
    }
```

roman-mirzoian (approved):
Well done and please pay attention to the comment.

PR #3292, opened 2024-05-29
solution

Kasprukov:


roman-mirzoian (changes requested):

1) Using the reduce method with the propagation operator (...) creates a new object on each iteration, which can be less efficient.
2) There is no check for the case where styles[1] may be undefined or empty (for example, if the string contains only a key with no value).

src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const stylesObject = sourceString
+    .split(';')
+    .map((styles) => styles.split(':'))
+    .reduce((object, [key, value]) => {
+      if (key && value !== undefined && value.trim() !== '') {
```
roman-mirzoian, line 13: Insufficient checks before assigning values: The code simply checks key, but does not check that key is not an empty string after removing spaces. This can lead to incorrect keys being added to the object.

roman-mirzoian (approved):
Well done and please pay attention to the comments

PR #3293, opened 2024-05-29
task solution

FaiHamid:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssDeclarations = sourceString.split(';');
```
roman-mirzoian, line 9: Splitting with split(';') works fine, but if the CSS string contains extra semicolons or empty lines, those will need to be processed as well.

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+  const styleObject = {};
+
+  cssDeclarations.forEach((property) => {
```
roman-mirzoian, line 12: You need to make sure that property is not empty or contains only spaces before dividing by propertyArr.
```suggestion
  cssDeclarations.forEach((property) => {
     if (property.trim()) { 
```

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+    const propertyArr = property.split(':', 2);
+    const [propertyName, propertyValue] = propertyArr;
+
+    if (propertyValue) {
```
roman-mirzoian, line 16: Checking for the existence of the `propertyName` value to avoid empty keys in the object.
```suggestion
    if (propertyName && propertyValue) {
```

roman-mirzoian (changes requested):
Good work, overall the code is pretty correct, but need to add some checks to improve the input string handling.

roman-mirzoian (approved):
Well done!

PR #3295, opened 2024-05-29
made convertToObject

StsKushnir:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const sourceArray = sourceString.split(';');
+
+  const objectProperties = sourceArray.reduce((acc, property) => {
+    const propertyArray = property.trim().split(':', 2);
+    const [keyName, keyValue] = propertyArray;
+
+    if (keyName && keyValue !== undefined) {
```
roman-mirzoian, line 15: The condition ``keyName && keyValue !== undefined`` is correct, but `keyValue` cannot be undefined because 
``split(':', 2)`` always returns an array with two elements. Better to check if `keyValue` is a non-empty string.

```suggestion

// Check if the key is not an empty string and if the value exists and is not an empty string.
    if (keyName && keyValue) {
```

roman-mirzoian (changes requested):
Well done, overall the code is good, but could use some precision in the checks.

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const sourceArray = sourceString.split(';');
+
+  const objectProperties = sourceArray.reduce((acc, property) => {
+    const propertyArray = property.trim().split(':', 2);
+    const [keyName, keyValue] = propertyArray;
+
+    if (keyValue) {
```
roman-mirzoian, line 15: Here the previous comment talked about checking for undefined, but here it's better to check if both ``keyName`` and ``keyValue`` exist, this will prevent adding properties to the object with empty keys.

```suggestion
    if (keyName && keyValue) {
```

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const sourceArray = sourceString.split(';');
```
roman-mirzoian, line 9: Empty strings left after splitting with `;` can create unnecessary elements.
Example solution for  AI comment: 
```suggestion
  const sourceArray = sourceString.split(';').filter(property => property.trim() !== '');
```

roman-mirzoian (changes requested):
Good job but pay attention to the comments.

roman-mirzoian (approved):
Well done

PR #3296, opened 2024-05-30
add solution

om1gse:


src/convertToObject.js
```diff
@@ -6,7 +6,26 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssStyleObject = {};
+  const declarations = sourceString.split(';');
+
+  declarations.forEach((declaration) => {
+    const newDeclaration = declaration.trim();
```
roman-mirzoian, line 13: Here's an example of how to filter out empty lines to make your code more robust.

```suggestion
  declarations
  .map((declaration) => declaration.trim())
  .filter((declaration) => declaration)
  .forEach((declaration) => {
    const newDeclaration = declaration.trim();
```

src/convertToObject.js
```diff
@@ -6,7 +6,26 @@
...
+
+    if (!newDeclaration) {
+      return;
+    }
```
roman-mirzoian, line 17: This check is now redundant as we have already filtered out empty declarations.
```suggestion
```

roman-mirzoian (changes requested):
Good job and please pay attention to the comments.

PR #3297, opened 2024-05-30
solution

seriuksergii:


src/convertToObject.js
```diff
@@ -5,8 +5,21 @@
...
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .trim()
+    .split(';')
+    .map((style) => style.trim())
+    .reduce((stylesObject, style) => {
+      const [property, value] = style.split(':');
```
roman-mirzoian, line 15: Instead of unpacking the object on each iteration, we simply add a new property to the existing object.
```suggestion
      const [property, value] = style.split(':').map(part => part.trim());
```

src/convertToObject.js
```diff
@@ -5,8 +5,21 @@
...
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .trim()
+    .split(';')
+    .map((style) => style.trim())
+    .reduce((stylesObject, style) => {
```
roman-mirzoian, line 14: Here you can add filtering of empty lines.
```suggestion
    .map((style) => style.trim())
    .filter(style => style)
    .reduce((stylesObject, style) => {
```

src/convertToObject.js
```diff
@@ -5,8 +5,21 @@
...
+      const [property, value] = style.split(':');
+
+      if (property && value) {
+        return { ...stylesObject, [property.trim()]: value.trim() };
```
roman-mirzoian, line 18: Add a property to the object, if there is both a key and a value.
```suggestion
        stylesObject[property] = value;
```

roman-mirzoian (changes requested):
Well done here are some comments to improve the code.

seriuksergii:
Hi, I fixed my mistakes.

roman-mirzoian (approved):
Well done!

PR #3302, opened 2024-05-31
Solution

gazebo88fire:


src/convertToObject.js
```diff
@@ -7,6 +7,26 @@
...
+
+    if (trimmedcssDeclaration) {
+      const parts = trimmedcssDeclaration.split(':');
+
+      const property = parts[0].trim();
+
+      const value = parts[1].trim();
+
+      cssObject[property] = value;
+    }
```
roman-mirzoian, line 26: This part of the code can be greatly improved.
Let`s do it: 

```suggestion
// Check if both parts exist after the split
      if (parts.length === 2) {
//Using destructuring to trim key and value simultaneously.
        const [property, value] = parts.map(part => part.trim());
        
        cssObject[property] = value;
      }
    }
```

roman-mirzoian (changes requested):
Good job, your code looks very good, but there are a few small improvements that could improve its efficiency and readability.

Please pay attention to the comments.

roman-mirzoian (approved):
Well done!

PR #3303, opened 2024-06-01
Sol

sa4ok-1:


src/convertToObject.js
```diff
@@ -5,8 +5,23 @@
  *
  * @return {object}
  */
-function convertToObject(sourceString) {
-  // write your code here
+function convertToObject(styles) {
+  const styleObject = {};
+
+  // Поділ  рядків крапками з комою
+  const declarations = styles.split(';');
```
roman-mirzoian, line 12: Here is an example of an improvement:
```suggestion
  const declarations = styles.split(';').filter(declaration => declaration.trim());
```

src/convertToObject.js
```diff
@@ -5,8 +5,23 @@
  *
  * @return {object}
  */
-function convertToObject(sourceString) {
-  // write your code here
+function convertToObject(styles) {
+  const styleObject = {};
+
+  // Поділ  рядків крапками з комою
```
roman-mirzoian, line 11: If you add comments, it is better in English.
Also, comments should be moderate. You don't just have to write comments on everything.

roman-mirzoian (changes requested):
Good job and please pay attention to the comments.

roman-mirzoian (changes requested):
@sa4ok-1

roman-mirzoian (approved):
Well done!

PR #3305, opened 2024-06-03
Solution

SSDoizha:


src/convertToObject.js
```diff
@@ -7,6 +7,20 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  return sourceString
+    .trim()
+    .split(';')
+    .map((style) => style.trim())
+    .filter((style) => style)
```
VitaliyBondarenko1982, line 14: map and filter iteration are redundant.
it is possible to make everything in reduce method.

src/convertToObject.js
```diff
@@ -7,6 +7,20 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  return sourceString
+    .trim()
```
VitaliyBondarenko1982, line 11: Looks like redundant.
Enough to trim in reduce every style.

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some refactoring.

lerastarynets (approved):
You're on the right track!

PR #3307, opened 2024-06-04
Develop

mishakoshilka:


src/convertToObject.js
```diff
@@ -6,7 +6,29 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  if (sourceString.length === 0) {
+    return {};
+  }
```
volodymyr-soltys97, line 11:
```suggestion
```

src/convertToObject.js
```diff
@@ -6,7 +6,29 @@
...
+
+  const result = {};
+  const splited = sourceString.split(';');
+
+  for (let i = 0; i < splited.length; i++) {
+    const target = splited[i].split(':');
```
volodymyr-soltys97, line 17:
```suggestion
    const [key, value] = splited[i].split(':').map(item => item.trim());
```

src/convertToObject.js
```diff
@@ -6,7 +6,29 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  if (sourceString.length === 0) {
+    return {};
+  }
+
+  const result = {};
+  const splited = sourceString.split(';');
+
+  for (let i = 0; i < splited.length; i++) {
```
volodymyr-soltys97, line 16: Use a method forEach instead of for loop

src/convertToObject.js
```diff
@@ -6,7 +6,29 @@
...
+    const target = splited[i].split(':');
+
+    if (target.length < 2) {
+      continue;
+    }
+
+    const key = target[0].trim();
+    const value = target[1].trim();
```
volodymyr-soltys97, line 24:
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

anastasiiavorobiova (approved):
Great work!

PR #3309, opened 2024-06-06
My solution

Karanelus:
My solution to this task


BogdanMaliuta (changes requested):
Hello, I`m sure it`s works, but here is some problem.
You are using here to much iteration methods, what impacts performance, you can just make same with 1-2 iteration methods

natalia-klonowska (approved):
Great work 🚀

PR #3312, opened 2024-06-08
Solution

Yaroslavlesh:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedObject = {};
+
+  const splited = sourceString.split(';');
```
roman-mirzoian, line 11: It is better to change the name of the `splited` variable to example - `declarations` for better understanding.

roman-mirzoian, line 11: Instead of checking ``if (split.trim())`` in each iteration, you can trim whitespace at the very beginning and then filter out empty lines.

```suggestion
  const splited = sourceString
    .split(';')
    .map((declaration) => declaration.trim())
    .filter((declaration) => declaration);
```

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+
+  splited.forEach((split) => {
+    if (split.trim()) {
+      const [property, value] = split.split(':');
+
+      convertedObject[property.trim()] = value.trim();
+    }
```
roman-mirzoian, line 18: Let's also improve the code here (Improved forEach logic):

```suggestion
  splited.forEach((split) => {
      const [property, value] = split.split(':').map((part) => part.trim());

      if (property && value) {
        convertedObject[property] = value;
      }
    }
```

roman-mirzoian (changes requested):
Good job let's improve your code.

roman-mirzoian (approved):
Well done!

PR #3315, opened 2024-06-10
[Feature] created styles converter

Arthur-Tolkachov:
## Description
Created styles-converter function


anastasiiavorobiova (approved):
Great work!

PR #3320, opened 2024-06-12
Solution

DSB-18:


roman-mirzoian (changes requested):
Good work, please pay attention to AI comments.
Also recommend adding a check that each element after splitting contains both parts (key and value).

DSB-18:
> Good work, please pay attention to AI comments. Also recommend adding a check that each element after splitting contains both parts (key and value).

Updated, hope didn't miss smth

roman-mirzoian (approved):
Well done!

PR #3328, opened 2024-06-18
solution

denisbielykh:


src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+  const styleDeclarations = sourceString.split(';');
+
+  styleDeclarations.forEach((item) => {
+    const [property, value] = item.split(':');
```
volodymyr-soltys97, line 13:
```suggestion
    const [property, value] = item.split(':').map(el => el.trim());
```

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+
+    if (property === undefined || value === undefined) {
+      return;
+    }
+
+    styleObject[property.trim()] = value.trim();
```
volodymyr-soltys97, line 19:
```suggestion
    if (property && value) {
      styleObject[property] = value;
    }
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

etojeDenys (approved):
lgtm

PR #3345, opened 2024-07-01
Solution

MariaKolisnyk:


etojeDenys (changes requested):
you can simplify this logic using the reduce method

etojeDenys (approved):
wow

PR #3346, opened 2024-07-02
Solution

myshkivska:


etojeDenys (approved):
lgtm, but you can simplify that using the reduce method

PR #3349, opened 2024-07-03
add task solution

Daniel-spec-dell:


src/convertToObject.js
```diff
@@ -5,8 +5,20 @@
...
  */
+
 function convertToObject(sourceString) {
   // write your code here
+  const cssStyle = sourceString
+    .split(';')
+    .map((item) => item.split(':'))
+    .map((item) => item.map((value) => value.trim()))
+    .reduce((styleObject, itemObj) => {
+      styleObject[itemObj[0]] = itemObj[1];
```
TarasHoliuk, line 16: `itemObj` is not a good name here as well - it's `styleEntry`
FYI

TarasHoliuk (approved):
GJ 👍 
One small naming suggestion just FYI

PR #3357, opened 2024-07-05
Solution

AnastasiiaKorolko:


src/convertToObject.js
```diff
@@ -6,7 +6,22 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const declarations = sourceString.split(';');
+
+  declarations.forEach((declaration) => {
+    const trimDeclaration = declaration.trim();
+
+    if (trimDeclaration) {
+      const [key, value] = trimDeclaration.split(':');
```
anastasiiavorobiova, line 16:
```suggestion
      const [key, value] = trimDeclaration.split(':').map(el => el.trim());
```

src/convertToObject.js
```diff
@@ -6,7 +6,22 @@
...
+
+      if (key && value) {
+        result[key.trim()] = value.trim();
```
anastasiiavorobiova, line 19:
```suggestion
        result[key] = value;
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Great work!

PR #3361, opened 2024-07-07
add task solution

indigo04:


src/convertToObject.js
```diff
@@ -7,6 +7,18 @@
...
 function convertToObject(sourceString) {
   // write your code here
+  const arr = sourceString.split(';');
+  const result = {};
+
+  arr.forEach((style) => {
+    if (style.trim()) {
+      const [key, value] = style.split(':');
+
+      result[key.trim()] = value.trim();
```
volodymyr-soltys97, line 17: You can simplify like this
```suggestion
      const [key, value] = style.split(':').map(item => item.trim());

      result[key] = value;
```

volodymyr-soltys97 (approved):
Good job 👍

PR #3367, opened 2024-07-09
Solution

StanislavKapytsia:


etojeDenys (approved):
lgtm

PR #3368, opened 2024-07-10
solution

OPTIMISTIXX:


src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  if (!sourceString) {
+    return {};
+  }
+
+  const arrayOfStyles = sourceString
+    .trim()
+    .split(';')
+    .filter((el) => el.trim().length > 0 && !el.includes(';'));
```
volodymyr-soltys97, line 16: You can simplify it
```suggestion
  const arrayOfStyles = sourceString.split(';')
```

src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
...
+  const stylesObject = {};
+
+  arrayOfStyles.forEach((el) => {
+    const [key, value] = el.split(':').filter((elem) => elem.length > 0);
```
volodymyr-soltys97, line 20:
```suggestion
    const [key, value] = el.split(':').map(item => item.trim());
```

src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
...
+
+    stylesObject[key.trim()] = value.trim();
```
```suggestion
    stylesObject[key] = value;
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

lerastarynets (approved):
lgtm 🚀

PR #3369, opened 2024-07-10
Solution

Teollan:


src/convertToObject.js
```diff
@@ -6,7 +6,31 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const propertyStrings = sourceString
+    .split(';')
+    .map((dirtyProperty) => dirtyProperty.trim())
+    .filter((property) => property !== '');
```
DanilWeda, line 12: Try divide this chain code to constants for better readability

src/convertToObject.js
```diff
@@ -6,7 +6,31 @@
...
+
+/**
+ * @param {string} propertyString
+ *
+ * @return {object}
+ */
+function propertyStringToObject(propertyString) {
+  const [name, args] = propertyString
+    .split(':')
+    .map((dirtyProperty) => dirtyProperty.trim());
```
DanilWeda, line 31: Try divide this chain code to constants for better readability

DanilWeda (changes requested):
All good! Thanks) 
Check some minor comment)

src/convertToObject.js
```diff
@@ -6,7 +6,36 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const splitCleanProperties = sourceString
+    .split(';')
+    .map((dirtyProperty) => dirtyProperty.trim());
+
+  const validPropertyStrings = splitCleanProperties.filter((property) => {
+    return property !== '';
+  });
+
```
Moroz-Dmytro, line 16:
```suggestion
  const validPropertyStrings = splitCleanProperties.filter((property) => property);

```

And you can just add this `filter` to previous chain

src/convertToObject.js
```diff
@@ -6,7 +6,36 @@
...
+  const properties = validPropertyStrings.map(propertyStringToObject);
```
Moroz-Dmytro, line 17: You can call this function `propertyStringToObject` inside `reduce` below, to avoid using useless loop

Moroz-Dmytro (changes requested):

src/convertToObject.js
Teollan: 👍

src/convertToObject.js
Teollan: Edited filter predicate, though linter still does not like as a one-liner.

About chaining filter. I talked to Danylo Hryhorets about this, and he advised against it in [this thread](https://mate.academy/chats/222864?selectedMessageId=2994558&threadId=2993849). And I'm kinda tired of redoing it for the third time, to be honest...

BudnikOleksii (approved):
Good work 👍

PR #3375, opened 2024-07-11
Solution

Kateryna99:
I was doing 'box-shadow test' with AI help, it is still a secret for me how this test worked

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -6,27 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  const resultObject = {};
   const stringModified = sourceString
     .split(/;\s*/)
-    .filter((item) => item.trim() !== '');
+    .filter((value) => value.length)
+    .map((item) => item.split(':').map((el) => el.trim()));
 
-  stringModified.forEach((item) => {
-    const index = item.indexOf(':');
+  return stringModified.reduce((prev, next) => {
+    const acc = next[0];
+    const value = next[1];
 
-    if (index !== -1) {
-      const objectKey = item.slice(0, index).trim();
-      const objectValue = item.slice(index + 1).trim();
+    prev[acc] = value;
 
-      if (resultObject[objectKey]) {
-        resultObject[objectKey] += ', ' + objectValue;
-      } else {
-        resultObject[objectKey] = objectValue;
-      }
-    }
-  });
-
-  return resultObject;
+    return prev;
+  }, {});
 }
 
 module.exports = convertToObject;


src/convertToObject.js
```diff
@@ -6,7 +6,27 @@
...
+      const objectKey = item.slice(0, index).trim();
+      const objectValue = item.slice(index + 1).trim();
+
+      if (resultObject[objectKey]) {
+        resultObject[objectKey] += ', ' + objectValue;
+      } else {
+        resultObject[objectKey] = objectValue;
+      }
+    }
+  });
```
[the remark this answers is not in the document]
Kateryna99: get used to use forEach :(

src/convertToObject.js
```diff
@@ -6,7 +6,27 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resultObject = {};
```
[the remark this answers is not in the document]
BudnikOleksii: > Avoid using generic names like `resultObject` for variables. Use descriptive names that reflect the purpose of the variable.

Good point

src/convertToObject.js
```diff
@@ -6,7 +6,27 @@
...
+  const stringModified = sourceString
+    .split(/;\s*/)
+    .filter((item) => item.trim() !== '');
```
BudnikOleksii, line 12:
```suggestion
    .filter((item) => item.trim());
```
You don't need to compare it with an empty string

src/convertToObject.js
```diff
@@ -6,7 +6,27 @@
...
+
+  stringModified.forEach((item) => {
```
BudnikOleksii, line 14: You can use `reduce` to generate an object

src/convertToObject.js
```diff
@@ -6,7 +6,27 @@
...
+    const index = item.indexOf(':');
```
```suggestion
    const [key, value] = item.split(':');
```
You can just split instead

src/convertToObject.js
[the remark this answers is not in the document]
BudnikOleksii: > Using `forEach` to mutate an existing object is acceptable, but consider using `map` and `reduce` to transform the array into an object in a more functional style. This would eliminate the need for the `resultObject` variable and the `if` statement inside the `forEach` loop.

Good point, let's use reduce

BudnikOleksii (changes requested):
Good work 👍 

Kateryna99:
OMG ! I SO LIKE THIS CODE ! Please do not ruin my happiness by saying there is any problem :)))

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+    .filter((value) => value.length)
+    .map((item) => item.split(':').map((el) => el.trim()));
+
+  return stringModified.reduce((prev, next) => {
+    const acc = next[0];
+    const value = next[1];
+
+    prev[acc] = value;
+
+    return prev;
```
```suggestion
  return stringModified.reduce((acc, [key, value]) => {
    acc[key] = value;

    return prev;
```
The first argument usually has the name `acc` - accumulator https://www.freecodecamp.org/news/how-to-use-javascript-array-reduce-method/

BudnikOleksii (approved):
👍

src/convertToObject.js
Kateryna99: it is strange, that is not my code

PR #3377, opened 2024-07-12
add solution

RimmaBuzuluk:


src/convertToObject.js
```diff
@@ -6,7 +6,22 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssStyles = {};
+
+  sourceString.split(';').forEach((item) => {
```
BudnikOleksii, line 11: You can use `reduce` instead of `forEach` to generate an object

src/convertToObject.js
```diff
@@ -6,7 +6,22 @@
...
+    const pairs = item.trim().split(':');
```
BudnikOleksii, line 12:
```suggestion
    const [key, value] = item.trim().split(':');
```
You can use destructing

src/convertToObject.js
```diff
@@ -6,7 +6,22 @@
...
+
+    if (pairs.length === 2) {
+      const key = pairs[0].trim();
+      const value = pairs[1].trim();
+
+      if (key !== undefined && value !== undefined) {
```
BudnikOleksii, line 18:
```suggestion
      if (key && value) {
```
You don't need to compare with `undefined` https://developer.mozilla.org/en-US/docs/Glossary/Truthy

BudnikOleksii (changes requested):
Good work 👍

BudnikOleksii (approved):
👍

PR #3378, opened 2024-07-12
add solution

coldshein:


src/convertToObject.js
```diff
@@ -5,8 +5,19 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
   // write your code here
+  return sourceString
+    .split(';')
+    .map((style) => style.trim())
+    .filter((style) => style !== '')
```
BudnikOleksii, line 14: You don't need to compare with an empty string https://developer.mozilla.org/en-US/docs/Glossary/Truthy

src/convertToObject.js
```diff
@@ -5,8 +5,19 @@
...
+    .map((style) => style.split(':'))
+    .reduce(
```
BudnikOleksii, line 16: You can split inside `reduce`

BudnikOleksii (changes requested):
Good work 👍

src/convertToObject.js
coldshein: > You don't need to compare with an empty string https://developer.mozilla.org/en-US/docs/Glossary/Truthy

So, what should I do? Just remove it?

BudnikOleksii: > > You don't need to compare with an empty string https://developer.mozilla.org/en-US/docs/Glossary/Truthy
> 
> So, what should I do? I'm filtering empty string

Just check if value exists
`.filter((style) => style)`

BudnikOleksii (approved):
👍

PR #3379, opened 2024-07-12
Solution

AndGrachov:


src/convertToObject.js
```diff
@@ -6,7 +6,15 @@
...
-  // write your code here
+  const cssProperties = sourceString
+    .split(';')
+    .map((property) => property.trim())
+    .filter((property) => property.length)
+    .map((property) => {
+      return property.split(':').map((style) => style.trim());
+    });
+
+  return Object.fromEntries(cssProperties);
```
BudnikOleksii, line 17: You can simplify and generate an object with `reduce` method instead of transform entries.

BudnikOleksii (changes requested):
Looks good 👍 
Let's try to rewrite with `reduce` method

BudnikOleksii (approved):
👍

PR #3380, opened 2024-07-12
Added solution

k-marchuk:


src/convertToObject.js
```diff
@@ -1,12 +1,38 @@
...
+ *
+ * @return {Object} CSS code converted to keys and values.
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return (
+    sourceString
+      // Handle cases where the string might start with whitespace or semicolons
+      .trim()
+      // Breaks sourceString by semicolon into separate stings
```
BudnikOleksii, line 15: Remove all comments

src/convertToObject.js
```diff
@@ -1,12 +1,38 @@
...
+
+      // Breaks line into property and value and removes extra space charachters
+      .map((line) => line.split(':').map((el) => el.trim()))
+
+      // Leaves only valid elements
+      .filter(
+        (el) =>
+          el.length === 2 &&
+          // Robust check that both elements are not empty strings
+          el.every((element) => element !== ''),
```
BudnikOleksii, line 26: You don't need to compare with an empty string, just check if elements exists https://developer.mozilla.org/en-US/docs/Glossary/Truthy

BudnikOleksii (changes requested):
Good work 👍

src/convertToObject.js
k-marchuk: It was suggested by ai-mentor :robot: https://github.com/mate-academy/js_style-to-object/pull/3380#discussion_r1675958870

VenomMB (approved):
Great job!!!

PR #3381, opened 2024-07-12
add task solution

OleksiiKislukhin:


src/convertToObject.js
```diff
@@ -6,7 +7,18 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .filter((style) => style.includes(':'))
+    .reduce((stylesObject, style) => {
+      const [property, value] = style.split(':').map((el) => el.trim());
+
+      if (property !== '' && value !== '') {
```
BudnikOleksii, line 16:
```suggestion
      if (property && value) {
```
You don't need to compare with an empty string https://developer.mozilla.org/en-US/docs/Glossary/Truthy

BudnikOleksii (approved):
Looks good 👍 
Approved, just one comment left

PR #3382, opened 2024-07-12
task_done

Anastasiia-Tatarchenko:


src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleDeclarations = sourceString.trim().split(';');
+  const styleObject = {};
+
+  for (const item of styleDeclarations) {
+    if (item.trim().length > 0) {
+      const resultItemSplit = item.split(':');
```
volodymyr-soltys97, line 14: Use forEach method instead for of loop
```suggestion
 styleDeclarations.forEach(() => {
      const resultItemSplit = item.split(':');
```

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+      const key = resultItemSplit[0].trim();
+      const value = resultItemSplit[1].trim();
```
volodymyr-soltys97, line 16: You can improve your code like this
```suggestion
      const [key, value] = item.split(':').map(item => item.trim());
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Great work 🔥

PR #3386, opened 2024-07-12
Solution

YaroslavKolesnyk:


src/convertToObject.js
```diff
@@ -6,7 +6,15 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString.split(';').reduce((result, style) => {
```
VenomMB, line 9:
```suggestion
  return sourceString.split(';').reduce((result, styleLine) => {
```

src/convertToObject.js
```diff
@@ -6,7 +6,15 @@
...
+    const [prop, value] = style.split(':').map((x) => x.trim());
```
VenomMB, line 10:
```suggestion
    const [prop, value] = styleLine.split(':').map((property) => property.trim());
```

VenomMB (approved):
Great job!
Leave small comments

PR #3390, opened 2024-07-13
Solution

vtroni:


src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .filter((item) => item.trim() !== '')
+    .map((item) => item.trim())
```
volodymyr-soltys97, line 12:
```suggestion
```

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+    .reduce((stylesObject, item) => {
+      const pair = item.split(':');
+
+      const key = pair[0].trim();
+      const value = pair[1].trim();
```
volodymyr-soltys97, line 17:
```suggestion
      const [key, value] = item.split(':').map(el => el.trim());
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

anastasiiavorobiova (approved):
Great work!

PR #3394, opened 2024-07-14
solution

MaxBorovique:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const declarations = sourceString.split(';');
+
+  const resultingObject = {};
```
ArturErmolenko, line 11:
```suggestion
  const styleObject = {};
```

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+
+  declarations.forEach((declaration) => {
```
ArturErmolenko, line 13: as for me `reduce `suits better here

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+    if (declaration.trim() !== '') {
+      const [key, value] = declaration.split(':').map((item) => item.trim());
```
ArturErmolenko, line 15:
```suggestion
      const [key, value] = declaration
         .split(':')
         .map((item) => item.trim());
```

ArturErmolenko (approved):

PR #3397, opened 2024-07-15
Solution

totum-revolutum:


src/convertToObject.js
```diff
@@ -6,7 +6,25 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  if (sourceString.length === 0) {
```
Moroz-Dmytro, line 9:
```suggestion
  if (!sourceString.length) {
```

Moroz-Dmytro: But is this `if` really needed?
I guess it will work correctly without it

Moroz-Dmytro (approved):

PR #3407, opened 2024-07-17
Solution

igor-romasiuk:


src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+  const lines = sourceString.split(';').filter((line) => line.trim() !== '');
```
BudnikOleksii, line 10:
```suggestion
  const lines = sourceString.split(';').filter((line) => line.trim());
```
You don't need to compare with an empty string, you can just check if value exists
https://developer.mozilla.org/en-US/docs/Glossary/Falsy

src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
...
+
+  lines.forEach((line) => {
```
BudnikOleksii, line 12: You can use reduce to generate an object

src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
...
+    const parts = line.split(':');
```
BudnikOleksii, line 13:
```suggestion
    const [property, value] = line.split(':');
```
You can use destructing

src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
...
+
+    if (parts.length >= 2) {
+      const property = parts[0].trim();
+      const value = parts.slice(1).join(':').trim();
```
BudnikOleksii, line 17: I think you don't need slice in this case as you can just get the second element, and why do you need to join ":"?

BudnikOleksii (changes requested):
Good work 👍 

BudnikOleksii (approved):
👍

PR #3408, opened 2024-07-17
convert to object

oliaShymanska:


etojeDenys (approved):
good job, but you can simplify that using the reduce method

PR #3421, opened 2024-07-22
Solution

IlliaAstakhov:


src/convertToObject.js
```diff
@@ -6,7 +6,29 @@
...
+
+      if (key && value) {
+        cssObject[key] = value;
+      }
+    }
+
+    return cssObject;
+  });
+
+  // console.log(cssObject);
```
etojeDenys, line 29: remove all comments

etojeDenys (changes requested):

IlliaAstakhov:
> you can simplify it using the reduce method

Do you mean, that I can use reduce method instead foreach?

etojeDenys (approved):
lgtm

PR #3430, opened 2024-07-28
Develop

MOODDDII:
Function that turns css strings to object

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -5,8 +5,37 @@
  *
  * @return {object}
  */
+'use strict';
+
+/**
+ * @param {string} sourceString
+ *
+ * @return {object}
+ */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+
+  sourceString.split('\n').forEach((s) => {
+    const str = s.trim();
+
+    if (str === '') {
+      return;
+    }
+
+    const colonIndex = str.indexOf(':');
+
+    const property = str.slice(0, colonIndex).trim();
+    const value = str.slice(colonIndex + 1).trim();
+
+    if (property && value) {
+      obj[property] = value;
+    }
+  });
+
+  return obj;
 }
+
+module.exports = convertToObject;
+
 
 module.exports = convertToObject;
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -25,10 +25,10 @@ function convertToObject(sourceString) {
     const colonIndex = str.indexOf(':');
 
     const property = str.slice(0, colonIndex).trim();
-    const value = str.slice(colonIndex + 1).trim();
+    let value = str.slice(colonIndex + 1).trim();
 
-    if (property && value) {
-      obj[property] = value;
+    if (value.endsWith(';')) {
+      value = value.slice(0, -1).trim();
     }
   });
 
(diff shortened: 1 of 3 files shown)


src/convertToObject.js
```diff
@@ -1,12 +1,40 @@
 'use strict';
 
+/**
+ * @param {string} sourceString
+ *
+ * @return {object}
+ */
+'use strict';
```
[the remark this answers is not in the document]
volodymyr-soltys97: Remove duplicates

src/convertToObject.js
```diff
@@ -1,12 +1,40 @@
...
+      obj[property] = value;
+    }
+  });
+
+  return obj;
 }
 
 module.exports = convertToObject;
+
+module.exports = convertToObject;
```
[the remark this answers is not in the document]
volodymyr-soltys97: Remove duplicates 
image

src/convertToObject.js
```diff
@@ -1,12 +1,40 @@
...
+  parts.forEach((str) => {
+    const cleanStr = str.trim();
+    const colonIndex = cleanStr.indexOf(':');
+
+    if (colonIndex === -1) {
+      return;
+    }
+
+    const property = cleanStr.slice(0, colonIndex).trim();
+    const value = cleanStr.slice(colonIndex + 1).trim();
```
volodymyr-soltys97, line 28: You can simplify it
```suggestion
    const [property, value] = str.split(':').map((item) => item.trim());
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/convertToObject.js
TarasHoliuk: I see you didn't apply the suggested changes
If it was unclear what exactly the suggested code does it's always better to ask it in the chat (or you have any other questions)

Generally, here you can use `split` method instead of code you've written
`const [property, value] =` - this part is destructuring assignment

src/convertToObject.js
```diff
@@ -5,8 +5,28 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  const StyleObj = {};
```
TarasHoliuk, line 10:
```suggestion
  const styleObj = {};
```
camelCase for variables
We use PascalCase for classes (or constructor functions) in JS

TarasHoliuk (changes requested):

TarasHoliuk (changes requested):
Hi! Let's fix naming

src/convertToObject.js
```diff
@@ -5,8 +5,27 @@
...
+
+  parts.forEach((str) => {
+    const cleanStr = str.trim();
+    const colonIndex = cleanStr.indexOf(':');
+
+    if (colonIndex === -1) {
+      return;
+    }
+
+    const [property, value] = cleanStr.split(':').map((item) => item.trim());
```
volodymyr-soltys97, line 21: You can simplify it
```suggestion
    const [property, value] = str.split(':').map((item) => item.trim());
```

volodymyr-soltys97 (changes requested):
Almost done!

lerastarynets (approved):
lgtm 🚀

PR #3435, opened 2024-07-30
add solution

George-Buzhenitsa:


src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
...
+    .filter((item) => item.length > 1)
+    .map((item) => {
+      return [item[0].trim(), item[1].trim()];
+    });
+
+  const finalObject = {};
+
+  updatedArray.forEach((item) => (finalObject[item[0]] = item[1]));
+
+  return finalObject;
```
volodymyr-soltys97, line 23:
```suggestion
  const updatedArray = sourceString.split(';')

  const finalObject = {};
  
  updatedArray.forEach((item) => {
    const [key, value] = item.split(':').map((el) => el.trim());
    
    if (key && value) {
      // write data into the object
    }
  })

  return finalObject;
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's simplify your code

volodymyr-soltys97 (approved):
Great 🔥

PR #3436, opened 2024-07-30
done

EvgenLutscenko:


etojeDenys (changes requested):
looks too difficult

you can rewrite this logic using the split method(split by `;` then by `:` and fill the object)

GUSILLUS (approved):
GJ!

PR #3444, opened 2024-08-05
add task solution

VladTk:


src/convertToObject.js
```diff
@@ -6,7 +6,31 @@
...
+  // replaceAll doesn't work on this version of Node
+  const clearedString = sourceString.replace(/[\t\n]/g, '');
+  const stylesArr = clearedString.split(';');
+  const stylesObj = {};
+
+  stylesArr.forEach((style) => {
+    let [name, value] = style.split(':');
+
+    // name and value can be empty or undefined
+    if (name && value) {
```
anastasiiavorobiova, line 24:
```suggestion
    if (name.trim() && value.trim()) {
```

anastasiiavorobiova (changes requested):
Great work! Consider fixing AI comments

SanyaBratashchuk (approved):

PR #3452, opened 2024-08-09
solution

VitaliiNez:


src/convertToObject.js
```diff
@@ -5,8 +5,20 @@
...
-  // write your code here
+  const obj = {};
+  const keyValuePairs = sourceString.trim().split(';');
+
+  keyValuePairs.forEach((el) => {
+    const [key, value] = el.split(':');
+
+    if (key && value) {
+      obj[key.trim()] = value.trim();
+    }
```
etojeDenys, line 18: you can simplify that using the reduce method

etojeDenys (approved):

PR #3457, opened 2024-08-12
Solution

ArturVyklynets:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedElements = {};
+  const styleElements = sourceString.split(';');
+
+  for (const element of styleElements) {
+    const [property, value] = element.split(':');
```
volodymyr-soltys97, line 13:
```suggestion
    const [property, value] = element.split(':').map(item => item.trim());
```

src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
+
+    if (value !== undefined) {
```
volodymyr-soltys97, line 15:
```suggestion
    if (property && value) {
```

src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
+      convertedElements[`${property.trim()}`] = `${value.trim()}`;
```
```suggestion
      convertedElements[property] = value;
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleProperties = {};
+  const styleElements = sourceString.split(';');
+
+  for (const element of styleElements) {
```
volodymyr-soltys97, line 12:
```suggestion
  styleElements.forEach((element) => {
```

volodymyr-soltys97 (approved):
Good job 👍

PR #3460, opened 2024-08-13
my solution "convert to obj"

Oksigenka:


lerastarynets (approved):
You're on the right track!

PR #3463, opened 2024-08-15
convert to object: solution

MaksymMohyla:


src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
...
+    return pair
+      .trim()
+      .split(':')
+      .map((part) => part.trim());
+  });
+
+  const stylesObject = {};
+
+  stylesArray.forEach(([key, value]) => {
+    stylesObject[key] = value;
```
etojeDenys, line 20: you can simplify it using the reduce method

etojeDenys (approved):
good job

src/convertToObject.js
MaksymMohyla: ok

PR #3467, opened 2024-08-16
Solution

Syogunok:


src/convertToObject.js
```diff
@@ -5,8 +5,26 @@
...
+
+  for (let i = 0; i < stylesArray.length; i++) {
+    const parts = stylesArray[i].split(':');
+
+    if (parts[0] && parts[1]) {
+      const property = parts[0].trim();
+      let value = parts[1].trim();
+
+      if (value.includes('!important')) {
+        value = value.replace(/\s*!important\s*/, '') + ' !important';
```
volodymyr-soltys97, line 20: Redundant
```suggestion
```

src/convertToObject.js
```diff
@@ -5,8 +5,26 @@
  *
  * @return {object}
  */
-function convertToObject(sourceString) {
-  // write your code here
+function convertToObject(stylesString) {
+  const stylesArray = stylesString.split(';');
+  const stylesObject = {};
+
+  for (let i = 0; i < stylesArray.length; i++) {
```
volodymyr-soltys97, line 12: Use forEach method instead of for loop

src/convertToObject.js
```diff
@@ -5,8 +5,26 @@
...
+    const parts = stylesArray[i].split(':');
+
+    if (parts[0] && parts[1]) {
```
volodymyr-soltys97, line 15: You can simplify like this 
```suggestion
    const [property, value] = stylesArray[i].split(':').map(item => item.trim());

    if (property && value) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/convertToObject.js
```diff
@@ -5,8 +5,19 @@
...
+  const stylesArray = stylesString.split(';');
+  const stylesObject = {};
+
+  stylesArray.forEach((style) => {
+    const parts = style.split(':');
+
+    if (parts[0] && parts[1]) {
+      stylesObject[parts[0].trim()] = parts[1].trim();
+    }
+  });
```
etojeDenys, line 18: you can simplify it using the reduce method

etojeDenys (approved):

PR #3468, opened 2024-08-16
Solution

nataliiasmirnova:


src/convertToObject.js
```diff
@@ -5,8 +5,74 @@
...
+    const indexEl = el.indexOf(':');
+    const key = el.slice(0, indexEl).replace(/\s+/g, '');
+    const value = el
+      .slice(indexEl, el.length)
+      .slice(1)
+      .trim()
+      .replace(/,\s{2,}/g, ',\n          ');
+
+    return [key, value];
+  });
```
volodymyr-soltys97, line 26:
```suggestion
  const filtredArr = sourceString.split(';')
```

src/convertToObject.js
```diff
@@ -5,8 +5,74 @@
...
+
+  pairs.forEach((pair) => {
+    styleObject[pair[0]] = pair[1];
+  });
```
volodymyr-soltys97, line 30:
```suggestion
  filtredArr.forEach((pair) => {
    const [property, value] = pair.split(':').map(item => item.trim())
    
    if (property && value) {
      styleObject[property] = value;
    }
  });
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's simplify your code

src/convertToObject.js
nataliiasmirnova: Thank you! It's really new level!

src/convertToObject.js
```diff
@@ -5,8 +5,63 @@
...
+  white-space: nowrap;
+  width: auto;
+`);
+
+convertToObject(`
+      box-shadow:
+          inset 0 -3em 3em rgb(0 200 0 / 30%),
+          0 0 0 2px white,
+          0.3em 0.3em 1em rgb(200 0 0 / 60%);
+    `);
```
volodymyr-soltys97, line 65: Remove it, it's redundant
```suggestion
```

volodymyr-soltys97 (changes requested):
Almost done!

TarasHoliuk (approved):
Nice solution 👍

PR #3471, opened 2024-08-19
Solution

anna-agerone:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
```
TarasHoliuk, line 9: good comment about naming:
image

FYI

TarasHoliuk (approved):

PR #3472, opened 2024-08-20
Solution

KrainovaAnastasia:


etojeDenys (approved):
you can simplify that using the reduce method, but looks good in general

PR #3473, opened 2024-08-21
Solution

rkoltunov:


src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
...
+  const result = {};
+  let styleRules = sourceString.split(';');
+
+  styleRules = styleRules.filter((rule) => rule.trim().length);
+
+  for (const line of styleRules) {
+    const [rule, value] = line.split(':');
 
+    result[rule.trim()] = value.trim();
+  }
```
etojeDenys, line 18: you can simplify it using the reduce method

etojeDenys (approved):
good job

PR #3474, opened 2024-08-21
js_style-to-object

Valentin-1988:


TarasHoliuk (approved):
Hi! Approved ✔️ 

BTW, all AI comments aren't critical, but fair - read them FYI

PR #3477, opened 2024-08-24
Solutuion

misulkin:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
-  // write your code here
+  const styles = {};
+
+  sourceString.split(';').forEach((declaration) => {
+    let [property, value] = declaration.split(':');
+
+    if (property && value) {
+      property = property.trim();
+      value = value.trim();
+      styles[property] = value;
```
volodymyr-soltys97, line 17: You can simplify it
```suggestion
    const [property, value] = declaration.split(':').map((item) => item.trim());

    if (property && value) {
      styles[property] = value;
```

volodymyr-soltys97 (approved):
Good job 👍

PR #3480, opened 2024-08-25
solved

Artem-Brui:


src/convertToObject.js
```diff
@@ -7,6 +7,29 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const result = {};
+  const regex = /\n+|\t+/g;
+
+  sourceString
+    .replaceAll(regex, ' ')
```
VitaliyBondarenko1982: replaceAll doesn't exists in 14 node version.
it is a reason of failed tests and in general redundant here.
Screenshot 2024-08-25 at 17 50 03

src/convertToObject.js
```diff
@@ -7,6 +7,29 @@
...
+    .split(';')
+    .map((item) => item.trim())
+    .filter((item) => item !== '')
+    .map((item) => {
+      return item.split(':').map((x) => {
```
VitaliyBondarenko1982, line 15: suggest to reduce array iterations count.
after split(';') it can be just one iteration (use reduce method or if it is too hard - forEach will be enough).

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some improvements and try to fix tests.

PR #3481, opened 2024-08-25
Solution

joaoguilherme000:


pedro-ruas (approved):
Nice job, João!
You could have omitted the variables created during the steps before the `.reduce` method and returning everything directly.
But your code has a good way for future debugging if needed. Great!

PR #3484, opened 2024-08-27
solution

Svidruk:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
+  const styleObject = {};
+  const declarations = sourceString.split(';');
+
+  declarations.forEach((declaration) => {
+    if (declaration.trim() !== '') {
+      const [key, value] = declaration.split(':');
+
+      styleObject[key.trim()] = value.trim();
+    }
+  });
```
VolodymyrKirichenko, line 18: u can use advance methods for it. Try something like this:
```suggestion
  declarations.filter(declaration => declaration.trim() !== '')
    .reduce((styleObject, declaration) => {
    
      YOUR VERSION
      
      return styleObject;
    }, {});
```

VolodymyrKirichenko (approved):
Good job 🥇 , but u can practice advance methods)
Look in my comment and improve it)

PR #3490, opened 2024-08-29
Solution

gordgelin:


okuzan (changes requested):
submitted to Java flow :(

gordgelin:
> submitted to Java flow :(

ok, sorry

i just wanted to go through some FS modules to refresh some JS staff, but it seems i should switch course first, it is just i do not want to see frontend/fs in my profile info ahahah, but now i see it does not work properly probably without switching main course on the platform

PR #3491, opened 2024-08-29
task solution

Dorosh90:


src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceArray = sourceString.split(';');
+
+  const convertsObject = {};
+
+  sourceArray.forEach((style) => {
+    const firstKeyIndex = style.indexOf(':');
```
volodymyr-soltys97, line 14:
```suggestion
    const [key, value] = style.split(':').map(item => item.trim());
```

src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
...
+
+    if (firstKeyIndex !== -1) {
+      const key = style.slice(0, firstKeyIndex).trim();
+      const firstValueIndex = firstKeyIndex + 1;
+      const value = style.slice(firstValueIndex).trim();
+
+      convertsObject[key] = value;
+    }
```
```suggestion
    if (key && value) {
      convertsObject[key] = value;
    }
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's simplify your code

volodymyr-soltys97 (approved):
Great! 🔥

PR #3498, opened 2024-09-01
solved task

PikalovAndrey:


src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+  sourceString
+    .trim()
+    .split(';')
+    .forEach((item) => {
+      const [key, value] = item.split(':').map((str) => str.trim());
+
+      if (key && value) {
+        cssProperties[key] = value;
+      }
+    });
```
BudnikOleksii, line 20: You can use `reduce` to generate an object

BudnikOleksii (changes requested):
Good job 👍 
Let's try to use `reduce` instead

BudnikOleksii (approved):
Looks great 🔥

PR #3504, opened 2024-09-02
task solution

W3zzie:


BogdanMaliuta (changes requested):
Not bad, but fix what AI suggested

src/convertToObject.js
```diff
@@ -7,6 +7,18 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  return sourceString
+    .split(';')
+    .filter((declaration) => declaration.trim())
```
zibi-dev, line 12: Be more explicit with conditions.

zibi-dev (approved):
LGTM!

PR #3505, opened 2024-09-02
Solution

kovaden414:


src/convertToObject.js
```diff
@@ -6,7 +6,16 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .map((value) => value.trim())
+    .filter((value) => value !== '')
+    .reduce((prev, string) => {
+      return {
+        ...prev,
+        [string.split(':')[0].trim()]: string.split(':')[1].trim(),
```
TarasHoliuk, line 16: it would be a bit more readable if you save key and value into a separate variables)

TarasHoliuk (approved):
GJ 👍

PR #3506, opened 2024-09-02
add solution

YuriiMykhailenko:


src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+    .split(';')
+    .map((property) => property.split(':'))
+    .filter((entry) => entry.length > 1);
+
+  sourceProperties.forEach((entry) => {
+    const entryKey = entry[0].trim();
+    const entryValue = entry[1].trim();
+
+    listOfProperties[entryKey] = entryValue;
+  });
```
VolodymyrKirichenko: u can use reduce for it.
the best method for working with objects
For example:
```suggestion
  return styles.reduce((acc, style) => {
    ADD YOUR LOGIC

    return acc;
  }, {});
```
After that, u won't need to create an `object` at the beginning and `return` it at the end of the function. The `object` will be ready and returned immediately after reducing

VolodymyrKirichenko (changes requested):
Good job 😮‍💨 
But let's make ur code more better! 🥹

BudnikOleksii (approved):
Good work 👍

PR #3507, opened 2024-09-02
Added solution

nklimovych:


.github/workflows/test.yml-template
```diff
@@ -0,0 +1,23 @@
...
+        node-version: [20.x]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm install
+    - run: npm test
```
SarakhmanFTMentor, line 23: redundant :)

SarakhmanFTMentor: but that's interesting trick to align your local and remote versions :)

SarakhmanFTMentor (approved):

PR #3509, opened 2024-09-03
Solution

DimaDamage91:


src/convertToObject.js
```diff
@@ -6,7 +6,16 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .map((value) => value.trim())
+    .filter((value) => value !== '')
```
volodymyr-soltys97: Redundant
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job!
Let's improve your code with AI comments

lerastarynets (approved):
You're on the right track!

PR #3511, opened 2024-09-03
task solution

Ivanavi808:


src/convertToObject.js
```diff
@@ -1,12 +1,19 @@
...
+
+  sourceString
+    .split(';')
+    .forEach((part) => {
+      if (part.trim()) {
+        const [command, value] = part.split(':');
+
+        finalStyle[command.trim()] = value.trim();
+      }
+  });
```
BudnikOleksii: You can use `reduce` to generate an object

BudnikOleksii (changes requested):
Good work 👍 
Let's try to use `reduce` instead of `forEach`

BudnikOleksii (approved):
Looks great 🚀

PR #3517, opened 2024-09-05
solution

YuliiaShatkovska:


src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+
+  const semicolon = ';';
+  const colon = ':';
```
VolodymyrKirichenko: Creating a variable for semicolon and colon  and using it only once is unnecessary and complicates the code without providing any benefits.
remove it

src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
...
+
+  const convertedToObject = sourceString
```
VolodymyrKirichenko: What does reduce return? Can u return it immediately?
```suggestion
  return sourceString
```

src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
```
VolodymyrKirichenko: u don't need this variable, because I see that u used `reduce` in this task)
u can create and object in the middle of it by setting the `initial value`(second parameter) and immediately return it after reducing

src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
...
+    .split(semicolon)
+    .reduce((styleAccumulator, declaration) => {
+      const [property, value] = declaration.trim().split(colon);
+
+      if (value) {
+        styleAccumulator[property.trim()] = value.trim();
+      }
+
+      return styleAccumulator;
+    }, styleObject);
```
VolodymyrKirichenko:
```suggestion
    }, {});
```

VolodymyrKirichenko (changes requested):
Good job 😮‍💨 
But let's improve ur code

VolodymyrKirichenko (approved):
Well done 🔥

PR #3518, opened 2024-09-05
Solution

EustasTheMonk:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const trimmedStyles = sourceString.split(';').map((style) => style.trim());
+
+  const clearedStyles = trimmedStyles
```
VolodymyrKirichenko: What does reduce return? Can u return it Immediately?
```suggestion
  return trimmedStyles
```

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+    .filter((element) => element)
+    .reduce((toObject, item) => {
```
VolodymyrKirichenko: `toObject`? Change the accumulator name

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+      const splitted = item.split(':');
```
VolodymyrKirichenko: splited what? idk) Change this name also

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const trimmedStyles = sourceString.split(';').map((style) => style.trim());
```
VolodymyrKirichenko: I think u can do it in middle of the `reduce`. Then u won't have to iterate over the string twice, but it's `Up to u`

VolodymyrKirichenko (changes requested):
Good job 👍 
But let's make ur code more readable and better!)

VolodymyrKirichenko (approved):
Well done 🔥 
nice code)

PR #3520, opened 2024-09-05
Develop

coldmelnyk:
Idk the reason but linter wants me adding some space between those two lines, so pushing like this:

Знімок екрана 2024-09-05 о 18 28 58
Знімок екрана 2024-09-05 о 18 28 52

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -9,7 +9,10 @@ function convertToObject(sourceString) {
   const cleared = sourceString.split(';');
 
   return cleared.reduce((accum, style) => {
-    const n = style.split(':').map((element) => element.trim());
+    const n = style
+      .split(':')
+
+      .map((element) => element.trim());
 
     const [key, value] = n;
 
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cleared = sourceString.split(';');
+
+  return cleared.reduce((accum, style) => {
+    const n = style.split(':').map((element) => element.trim());
+
+    const [key, value] = n;
+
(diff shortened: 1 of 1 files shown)


src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cleared = sourceString.split(';');
```
VolodymyrKirichenko: what the - `cleared`? idk)
Change this variable name

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+
+  return cleared.reduce((accum, style) => {
+    const n = style
```
VolodymyrKirichenko: `n`? what do u mean? `ninja` or maybe `nintendo`?)
Change this name also

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+      .split(':')
+
+      .map((element) => element.trim());
```
VolodymyrKirichenko: i think we can do it without map
At each iteration u have a `key` and a `value`.
u can work with them without additional methods.
`I'll leave a comment below for better understanding`

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+
+    const [key, value] = n;
+
+    accum[key] = value;
```
VolodymyrKirichenko: variant without map(additional sorting)
```suggestion
    if (value) {
      accum[key.trim()] = value.trim();
    }
```

VolodymyrKirichenko (changes requested):
Good job 👍 
But let's improve ur code! 😮‍💨

VolodymyrKirichenko (approved):
Well done 🔥

PR #3521, opened 2024-09-05
solution

Dabl-B:


src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .filter((style) => style.length > 1 && style.includes(':'));
+
+  return styles.reduce((acc, style) => {
+    const [rawKey, rawValue] = style.split(':');
+
+    acc[rawKey.trim()] = rawValue.trim();
```
VolodymyrKirichenko, line 16:
```suggestion
    if (rawKey && rawValue) {
      acc[rawKey.trim()] = rawValue.trim(); // instead filter method
    }  
```

src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .filter((style) => style.length > 1 && style.includes(':'));
```
VolodymyrKirichenko, line 11: here u don't need an extra iteration (filter method).
u can check ur `value` on each iteration in ur `reduce` method.
I'll leave comments below
```suggestion
```

VolodymyrKirichenko (approved):
Well done 🔥 
But plz, review my comments

PR #3522, opened 2024-09-05
Solution

KrivorotkoYurii:


src/convertToObject.js
```diff
@@ -1,12 +1,21 @@
 'use strict';
 
-/**
- * @param {string} sourceString
- *
- * @return {object}
- */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
```
VolodymyrKirichenko, line 4: u can do it immediately in ur case
```suggestion
  return sourceString
```

src/convertToObject.js
```diff
@@ -1,12 +1,21 @@
...
+    .map((style) => style.trim())
+    .filter((el) => el)
+    .reduce((accumulator, value) => {
+      const result = value.split(':');
+
+      const [styleKey, styleValue] = result;
+
+      accumulator[styleKey.trim()] = styleValue.trim();
+
+      return accumulator;
```
VolodymyrKirichenko, line 15: You're performing two unnecessary iterations (`filter` and `map`). In the `reduce` method, you're already processing each element, so u can use `trim()` inside.
Also, instead of using the `filter` method, u can apply a condition to check if `value !== undefined`.

```suggestion
    .reduce((accumulator, value) => {
      const result = value.trim().split(':'); // instead map method

      const [styleKey, styleValue] = result;

      if (styleValue) {
        accumulator[styleKey.trim()] = styleValue.trim(); // instead filter method
      }  

      return accumulator;
```

VolodymyrKirichenko (approved):
Good job 🥇 
Bur plz, review my comments and fix them 🥹

PR #3523, opened 2024-09-05
Develop

stasLiutyww:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceStringParts = sourceString.split(';');
+  const strWithoutSpaces = sourceStringParts.map((str) => str.trim());
```
VolodymyrKirichenko, line 10: u don't need an extra iteration since you're already using the `reduce` method. Below I'll show u how u can do this

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+  const result = strWithoutSpaces.reduce((prev, el) => {
+    const [key, value] = el.split(':');
```
VolodymyrKirichenko, line 12:
```suggestion
    const [key, value] = el.trim().split(':'); // Instead of the map method, since you already have all the elements in that place
```

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const sourceStringParts = sourceString.split(';');
+  const strWithoutSpaces = sourceStringParts.map((str) => str.trim());
+  const result = strWithoutSpaces.reduce((prev, el) => {
```
VolodymyrKirichenko, line 11: This is not the `previous` value. This is a `accumulator`. Change it.
Also u can return it immediately)

```suggestion
  return strWithoutSpaces.reduce((acc, el) => {
```

VolodymyrKirichenko (approved):
Good job 🔥 
But review my comments and fix them)

PR #3524, opened 2024-09-05
task solution

olshum8:


src/convertToObject.js
```diff
@@ -5,8 +5,21 @@
...
+  const result = {};
+
+  const stylesArr = sourceString.split(';');
+
+  stylesArr.forEach((style) => {
+    const [key, value] = style.split(':');
+
+    if (key && value) {
+      result[key.trim()] = value.trim();
+    }
```
volodymyr-soltys97, line 19: You can sicmplify it
```suggestion
    const [key, value] = style.split(':').map(item => item.trim());

    if (key && value) {
      result[key] = value;
    }
```

volodymyr-soltys97 (approved):
Great work! 🔥

PR #3525, opened 2024-09-05
solution

dokrod:
add task solution

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .filter((style) => style.length > 1 && style.includes(':'));
+
+  return styles.reduce((prev, style) => {
+    const [key, value] = style.split(':');
+
+    prev[key.trim()] = value.trim();
+
+    return prev;
+  }, {});
 }
 
 module.exports = convertToObject;


src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .filter((style) => style.length > 1 && style.includes(':'));
```
VolodymyrKirichenko, line 11: here u don't need an extra iteration (filter method).
u can check ur value on each iteration in ur reduce method.
I'll leave comments below
```suggestion
```

src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
...
+
+  return styles.reduce((prev, style) => {
+    const [key, value] = style.split(':');
+
+    prev[key.trim()] = value.trim();
```
VolodymyrKirichenko, line 16:
```suggestion
    if (key && value) {
      prev[key.trim()] = value.trim(); // instead filter method
    }  
```

src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .filter((style) => style.length > 1 && style.includes(':'));
+
+  return styles.reduce((prev, style) => {
```
VolodymyrKirichenko, line 13: It's not a `prev` value, it's an `accumulator`. Change it
```suggestion
  return styles.reduce((acc, style) => {
```

VolodymyrKirichenko (approved):
Good job 👍 
But plz, review my comments

PR #3526, opened 2024-09-05
Solution

Valeraom:


src/convertToObject.js
```diff
@@ -6,7 +6,14 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .filter((style) => style.trim() !== '')
+    .map((style) => style.split(':'));
+
+  return styles.reduce((prevStyles, [property, value]) => {
+    return { ...prevStyles, [property.trim()]: value.trim() };
+  }, {});
```
VolodymyrKirichenko: 1) This is not previous value. This is `accumulator`
2) use destructuring separately for better readability
3) u don't need create object on each iteration.
at each iteration you can add keys to the already created object (accumulator (second parameter)).

```suggestion
  return styles.reduce((acc, style) => { //1
    const [property, value] = style; // 2
    
    acc[property.trim()] = value.trim(); //3
  
    return acc;
  }, {});
```

src/convertToObject.js
```diff
@@ -6,7 +6,14 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .filter((style) => style.trim() !== '')
```
VolodymyrKirichenko: Before adding a key with a property in ur reduce method, u can check if `ur properties are !== undefined`. If not, add a key. Then u won't need a new search
```suggestion
```

src/convertToObject.js
```diff
@@ -6,7 +6,14 @@
...
+    .map((style) => style.split(':'));
```
VolodymyrKirichenko: u have each element in ur `reduce` method. Do this action in the middle
```suggestion
```

VolodymyrKirichenko (changes requested):
Looks good, but review my comments and fix them 😮‍💨

BudnikOleksii (approved):
👍

PR #3527, opened 2024-09-06
Solution

Hradivnyk:


src/convertToObject.js
```diff
@@ -6,7 +6,16 @@
...
+  const properties = sourceString.split(';');
+  const styleStringObject = {};
+
+  for (const property of properties) {
+    const [key, value] = property.split(':').map((item) => item.trim());
+
+    styleStringObject[key] = value;
+  }
+
+  return styleStringObject;
```
VolodymyrKirichenko, line 18: U can do it using `reduce` method 🥹 
```suggestion
   return sourceString.split(';').reduce((acc, style) => {

   `...`

    return styleAccumulator;
  }, {});
```

VolodymyrKirichenko (approved):
Well done 🔥 
But check my commets

PR #3528, opened 2024-09-06
Solution

AnastasiiaSavrasna:


src/convertToObject.js
```diff
@@ -6,7 +6,14 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  return sourceString.split(';').reduce((finalStyle, part) => {
```
VolodymyrKirichenko, line 9: This is `accumulator` in all the cases. Also, u are iterating over `styles`, so each element will be named accordingly, not just `part`. Part of what?
```suggestion
  return sourceString.split(';').reduce((acc, style) => {
```

VolodymyrKirichenko (approved):
Well done 🔥

PR #3530, opened 2024-09-06
Solution to the task

v1RnT:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+    const trimmedLine = line.trim();
+
+    const [property, value] = trimmedLine.split(':');
+
+    if (property && value) {
+      styleObject[property.trim()] = value.trim();
+    }
+  });
+
+  return styleObject;
```
VolodymyrKirichenko, line 21: U can use reduce method for it
```suggestion
  return sourceString.split(';').reduce((acc, style) => {
    const [property, value] = style.trim().split(':');

    if (property && value) {
      acc[property.trim()] = value.trim();
    }
    
    return acc;
  }, {});
```

VolodymyrKirichenko (approved):
Well done 🔥 
Check my comment)

PR #3531, opened 2024-09-06
solution

Ostkreuzzz:


src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let result = sourceString.split(';').map((arr) => arr.split(':'));
+
+  result = result.map(
+    (subArray) => subArray.map((string) => string.trim()),
+    // eslint-disable-next-line prettier/prettier
+  );
```
VolodymyrKirichenko: don't disable  prettier.
```suggestion
  result = result.map((subArray) => (
    subArray.map((string) => string.trim())
  ));
```
u can change ur syntax

src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let result = sourceString.split(';').map((arr) => arr.split(':'));
```
VolodymyrKirichenko: u do 3 iterations(map -> filter -> map) over the entire line or array. 
u can do all this in the middle of your reduce method
```suggestion
  const splitedString = sourceString.split(';');
```

src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
...
+    (subArray) => subArray.map((string) => string.trim()),
+    // eslint-disable-next-line prettier/prettier
+  );
+
+  result = result.filter((subArray) => subArray.length > 1);
+
+  result = result.reduce(
+    (aggr, [key, value]) => ({ ...aggr, [key]: value }),
+    {},
+  );
```
VolodymyrKirichenko: 1) This is not `aggr`. It will be `accumulator` in all the cases
2) At the beginning of ur reduce, you set an initial value - an empty object. At the end, u will have a completely finished object, so we can return it immediately
3) Do destructuring separately for better readability
4) At each iteration u have ur property, for example(`padding-left    : 18px    `), so u can use `trim` and `spit` at this stage and pull out the values ​​ready to be added to the object
5) instead of ur filtering(`filter method`), u can add a condition if `value !== undefined`
```suggestion
  return result.reduce((acc, style) => { // 1, 2
    const [key, value] = style.... ....; //3, 4(instead ur map method)
    
    if (key && value) { // 5(instead ur filter method)
      ADD NEW KEY WITH VALUE TO UR OBJECT(`acc`)
    }
    
    return acc;
  }, {});
```

src/convertToObject.js
VolodymyrKirichenko: u only need `reduce` here
```suggestion
  result = result.map(
    (subArray) => subArray.map((string) => string.trim()),
    // eslint-disable-next-line prettier/prettier
  );
```

src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  let result = sourceString.split(';').map((arr) => arr.split(':'));
+
+  result = result.map(
+    (subArray) => subArray.map((string) => string.trim()),
+    // eslint-disable-next-line prettier/prettier
+  );
+
+  result = result.filter((subArray) => subArray.length > 1);
```
VolodymyrKirichenko: u only need `reduce` here

VolodymyrKirichenko (changes requested):
Overall - good job 👍 
But let's improve ur code and practice new `reduce` method) 🥹

BudnikOleksii (approved):
Good work 👍

PR #3532, opened 2024-09-06
Solution

WatarJoy:


src/convertToObject.js
```diff
@@ -6,7 +6,40 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const resObj = {};
```
BudnikOleksii: `res/obj/sum/arr/etc` are bad names

src/convertToObject.js
```diff
@@ -6,7 +6,40 @@
...
+    if (styleString.includes(':')) {
+      const index = styleString.indexOf(':');
+      const property = styleString.slice(0, index).trim();
+      const value = styleString.slice(index + 1).trim();
+
+      if (currentProperty) {
+        resObj[currentProperty] = currentValue.trim();
+        currentProperty = '';
+        currentValue = '';
+      }
```
BudnikOleksii: Could be simplified, and let's try to use `reduce` instead of `forEach` to generate an object

src/convertToObject.js
```diff
@@ -6,7 +6,40 @@
...
+
+  styleStrings.forEach((styleString) => {
+    if (styleString.trim() === '') {
+      return;
+    }
+
+    if (styleString.includes(':')) {
+      const index = styleString.indexOf(':');
+      const property = styleString.slice(0, index).trim();
+      const value = styleString.slice(index + 1).trim();
```
BudnikOleksii: You don't need to find index, you can just split by `:` and get property and value

src/convertToObject.js
```diff
@@ -6,7 +6,40 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const resObj = {};
+  const cleanedSourceString = sourceString.replace(/\t+/, ' ').trim();
+  const styleStrings = cleanedSourceString.split(';');
+  let currentProperty = '';
+  let currentValue = '';
+
+  styleStrings.forEach((styleString) => {
+    if (styleString.trim() === '') {
```
BudnikOleksii: You don't need to compare with an empty string, just check if a value exists https://developer.mozilla.org/en-US/docs/Glossary/Falsy

src/convertToObject.js
```diff
@@ -6,7 +6,40 @@
...
+      if (currentProperty) {
+        resObj[currentProperty] = currentValue.trim();
+        currentProperty = '';
+        currentValue = '';
+      }
+
+      currentProperty = property;
+      currentValue = value;
+    } else if (currentProperty) {
+      currentValue += ' ' + styleString.trim();
```
BudnikOleksii: Prefer template literals over concatenation https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals
And you actually don't need these variables `currentProperty/currentValue`

BudnikOleksii (changes requested):
Good work 👍 
Left some comments to improve your solution

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const cleanedSourceString = sourceString.replace(/\t+/g, ' ').trim();
+  const styleStrings = cleanedSourceString.split(';');
+
+  return styleStrings.reduce((acc, styleString) => {
+    if (styleString.trim()) {
+      const [property, ...valueParts] = styleString.split(':');
+      const value = valueParts.join(':').trim();
```
```suggestion
      const [property, value] = styleString.split(':');
      const preparedValue = value.trim();
```
Probably should work the same

BudnikOleksii (approved):
👍

PR #3534, opened 2024-09-08
Solution

alena-bondar:


src/convertToObject.js
```diff
@@ -6,7 +6,27 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleArr = sourceString.split(';');
+  const newStyles = {};
+
+  styleArr.map((style) => {
```
etojeDenys: better to use reduce here instead of the map

etojeDenys (approved):

PR #3535, opened 2024-09-08
initial commit

michal-szwindowski:


src/convertToObject.js
```diff
@@ -6,7 +6,23 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  // Rozdzielamy ciąg na linie, usuwamy puste linie
+  // i linie zawierające tylko białe znaki
+  const lines = sourceString.split(';').filter((line) => line.trim());
+
+  // Używamy reduce do przekształcenia tablicy linii w obiekt stylów
+  return lines.reduce((stylesObject, line) => {
+    // Używamy wyrażenia regularnego do wyodrębnienia klucza i wartości z linii
+    const match = line.match(/^\s*([^:]+?)\s*:\s*([\s\S]+?)\s*$/);
```
zibi-dev, line 16: Here i would just simply use split method instead of regex. Its less performant and more complex code to read.

zibi-dev (approved):
Hey. Good job with implementation. I have left one comment for future reference.

PR #3536, opened 2024-09-08
Solution

vikshiyan:


src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const styleDeclarations = sourceString.split(';');
+
+  const resultObject = styleDeclarations.reduce((acumulator, item) => {
```
BudnikOleksii, line 11: result/object/array/res/obr/item/etc are bad names for variable

BudnikOleksii (approved):
Good work 👍 ~
Approved, but pay attention to variable names, they should be descriptive

PR #3546, opened 2024-09-15
add task solution

D-Ptashevskyi:


src/convertToObject.js
```diff
@@ -6,7 +6,28 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cleanedString = sourceString.trim();
+
+  const declarations = cleanedString.split(';');
+
+  const result = {};
+
+  for (const declaration of declarations) {
```
BudnikOleksii, line 15: - `result` is a bad name
- You can use `reduce` method to generate an object

BudnikOleksii (changes requested):
Good work 👍 
Let's try to use `reduce` method to generate an object instead

PR #3552, opened 2024-09-18
Solution

IvanVaniuk:


anastasiiavorobiova (changes requested):
Great start! Consider fixing AI comments. You can chain array methods and leverage the reduce method to get the result. It would be better to refactor the code this way

anastasiiavorobiova (approved):
Awesome!

PR #3579, opened 2024-09-30
Solution

ab3MN:
#### Implemented solution

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -6,7 +6,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
+    .split(';')
+    .map((style) => style.split(':'))
+    .reduce((acc, [key, value]) => {
+      if ((key, value)) {
+        const [_key, _value] = [key.trim(), value.trim()];
+
+        acc[_key] = _value;
+      }
+
+      return acc;
+    }, {});
+
+  return styles;
 }
 
 module.exports = convertToObject;
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -5,19 +5,17 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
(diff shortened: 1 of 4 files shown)


src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = sourceString
```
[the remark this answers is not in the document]
VolodymyrKirichenko: Good point 👍 
Change it

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+    .split(';')
+    .map((style) => style.split(':'))
```
VolodymyrKirichenko: I see u are using the `reduce` method. Why didn't u do it in the middle of the `reduce` method, on each iteration? + you can use convenient destructuring for this.
this is redundant
remove `map`

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+    .reduce((acc, [key, value]) => {
+      if ((key, value)) {
```
VolodymyrKirichenko:
```suggestion
    .reduce((acc, declaration) => {
    
      const [key, value] = declaration.trim().split(':'); // instead map
    
      if ((key, value)) {
```

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
+        const [_key, _value] = [key.trim(), value.trim()];
```
VolodymyrKirichenko: 1) why do u need these underlines?
is this some special code style?
share a link to it, and explain

2) why can't we immediately use it in our condition on the line?
do u need to take it as a separate term?

VolodymyrKirichenko (changes requested):
Good job, but i have a few improvements and  questions for u 🥹

VolodymyrKirichenko (approved):

PR #3584, opened 2024-10-01
my solution to upgrade slyles

minimal005:

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -1,12 +1,23 @@
 'use strict';
 
-/**
+/*
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleUpgrated = sourceString
+    .split(';')
+    .map((el) => el.split(':'))
+    .reduce((prev, oneStyle) => {
+      if (oneStyle[0] && oneStyle[1]) {
+        return { ...prev, [oneStyle[0].trim()]: oneStyle[1].trim() };
+      }
+
+      return prev;
+    }, {});
+
+  return styleUpgrated;
 }
 
 module.exports = convertToObject;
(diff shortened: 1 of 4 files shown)


src/convertToObject.js
```diff
@@ -1,12 +1,23 @@
...
 
-/**
+/*
  * @param {string} sourceString
  *
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleUpgrated = sourceString
```
VolodymyrKirichenko: u can return it immediately

```suggestion
  return sourceString
```

src/convertToObject.js
```diff
@@ -1,12 +1,23 @@
...
+    .split(';')
+    .map((el) => el.split(':'))
```
VolodymyrKirichenko: do u need this `map` method? u can do this on each iteration in the middle of the `reduce` method

src/convertToObject.js
```diff
@@ -1,12 +1,23 @@
...
+    .reduce((prev, oneStyle) => {
+      if (oneStyle[0] && oneStyle[1]) {
```
VolodymyrKirichenko: instead map

```suggestion
    .reduce((prev, oneStyle) => {
    
    const [property, value] = oneStyle.trim().split(':');
    
      if (oneStyle[0] && oneStyle[1]) {
```

src/convertToObject.js
```diff
@@ -1,12 +1,23 @@
...
+        return { ...prev, [oneStyle[0].trim()]: oneStyle[1].trim() };
```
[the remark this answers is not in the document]
VolodymyrKirichenko: good point 🔥 
don't do it
```suggestion
        prev[property.trim()] = value.trim();
```

VolodymyrKirichenko (changes requested):
Well done 🔥 
But let's improve ur code 🥹

VolodymyrKirichenko (approved):

PR #3599, opened 2024-10-07
Solution

Krykunov:


src/convertToObject.js
```diff
@@ -5,8 +5,23 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  // Remove tab characters and split the string into an array
+  const styleStringArray = sourceString.replaceAll('\t', '').split(';');
```
VolodymyrKirichenko, line 11: tab only?
I think this is redundant.
It should work without the `replaceAll` method, since u use `trim` later

src/convertToObject.js
```diff
@@ -5,8 +5,23 @@
...
+
+  // Reduce the array into an object, trimming the key and value
+  const styleObject = styleStringArray.reduce((acc, style) => {
```
VolodymyrKirichenko, line 14: u can return it immediately, but up to u)

```suggestion
  return styleStringArray.reduce((acc, style) => {
```

VolodymyrKirichenko (approved):
Well done 🔥

PR #3642, opened 2024-10-27
Solution

Artemida1609:


Artemida1609:
hello ?

src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
...
+
+  items.map((item) => {
+    const keyValue = item.split(':');
+
+    if (keyValue.length === 2) {
+      const key = keyValue[0].trim();
+      const value = keyValue[1].trim();
+
+      result[key] = value;
+    }
```
volodymyr-soltys97: You can simplify your code like this
```suggestion
    const [key, value] = item.split(':').map((it) => it.trim());

    if (key && value) {
      result[key] = value;
    }
```

src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
...
+  });
+
+  return result;//all done
```
volodymyr-soltys97: Remove comments
```suggestion
  return result;
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Great work! 🔥

PR #3644, opened 2024-10-27
function implemented

mfirka:


mfirka:
added property and value validation

PR #3654, opened 2024-11-03
Solution

OkMoroz:


src/convertToObject.js
```diff
@@ -6,7 +6,26 @@
...
+
+  cleanStrings.forEach((cleanString) => {
+    const partsOfString = cleanString
+      .split(':')
+      .map((partStyle) => partStyle.trim());
+
+    const key = partsOfString[0];
+    const value = partsOfString[1];
+
+    if (key && value) {
```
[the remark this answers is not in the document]
OkMoroz: if (partsOfString.length === 2) {...}

src/convertToObject.js
```diff
@@ -6,7 +6,26 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+  const cleanStrings = sourceString
+    .trim()
+    .split(';')
+    .filter((style) => style.length);
```
VolodymyrKirichenko: this is redundant.
u can do this check on each iteration in ur `forEach` method

src/convertToObject.js
```diff
@@ -6,7 +6,26 @@
...
+
+  cleanStrings.forEach((cleanString) => {
+    const partsOfString = cleanString
+      .split(':')
+      .map((partStyle) => partStyle.trim());
+
+    const key = partsOfString[0];
+    const value = partsOfString[1];
+
```
VolodymyrKirichenko: do u need `map` in this case?
```suggestion

    const key = partsOfString[0].trim();
    const value = partsOfString[1].trim();

```

src/convertToObject.js
```diff
@@ -6,7 +6,26 @@
...
+    if (key && value) {
+      styleObject[key] = value;
+    }
+  });
+
+  return styleObject;
```
VolodymyrKirichenko: u can do it easier using `reduce` method
u will avoid using 2 extra iterations of `filter ` and `map `
write all the logic in the middle of the `reduce' method

```suggestion
  return sourceString.split(';').reduce((style, partsOfString) => {
  
    ..............

    return style;
  }, {});
```

src/convertToObject.js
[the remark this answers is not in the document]
VolodymyrKirichenko: @OkMoroz 
It sounds like a check to avoid something that will never happen to u)
it's redundant)

VolodymyrKirichenko (changes requested):
Good job 👍 
But let's practice a new `reduce` method 🥹 
Review my comments and fix it

VolodymyrKirichenko (approved):

src/convertToObject.js
OkMoroz: Дякую, цікаво. Спробую :)

PR #3655, opened 2024-11-03
Solution

natalia646:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+  const props = sourceString.split(';').map((prop) => prop.trim());
+
+  for (const prop of props) {
+    const [key, value] = prop.split(':');
```
VolodymyrKirichenko, line 13: u can use `reduce` method for it. The best method for objects.
with its help, u will avoid unnecessary iteration (map method)

src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
```
VolodymyrKirichenko, line 9: Each variable must be responsible for what is contained within it.
Use the more clear name.
For example:

```suggestion
  const stylesObject = {};
```

src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
+  const props = sourceString.split(';').map((prop) => prop.trim());
```
VolodymyrKirichenko, line 10: It seems to me that it is not props 🤔

src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
+
+  for (const prop of props) {
+    const [key, value] = prop.split(':');
+
+    if (key) {
+      obj[key.trim()] = value.trim();
+    }
+  }
+
+  return obj;
```
VolodymyrKirichenko, line 20: we can do it in `reduce` method. Try it
just passed ur logic in the middle of the `reduce` method

```suggestion
  return sourceString.split(';').reduce((style, declaration) => {
    
    .........

    return style;
  }, {});
```

VolodymyrKirichenko (changes requested):
Well done 🔥 
But let's practice a new `reduce` method 🥹

VolodymyrKirichenko (approved):

PR #3661, opened 2024-11-04
Create object

AlyonaV22:


src/convertToObject.js
```diff
@@ -6,7 +6,22 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+
+  const splitString = sourceString.trim().split(';');
+
+  splitString.forEach((item) => {
+    const splitItem = item.trim().split(':');
```
VolodymyrKirichenko: u can use destructuring here
```suggestion
    const [key, value] = item.trim().split(':');
```

src/convertToObject.js
```diff
@@ -6,7 +6,22 @@
...
+
+    if (splitItem.length === 2) {
+      const key = splitItem[0].trim();
+      const value = splitItem[1].trim();
+
+      result[key] = value;
+    }
```
VolodymyrKirichenko: it looks weird)
ur if can look like this
```suggestion
    if (key && value) {
      result[key.trim()] = value.trim();
    }
```

src/convertToObject.js
```diff
@@ -6,7 +6,22 @@
...
+  });
+
+  return result;
```
VolodymyrKirichenko: u can use `reduce` method for it. Try it

```suggestion
   return sourceString.split(';').reduce((style, declaration) => {
    const [key, value] = declaration.trim().split(':');

   ..............
   
    return style;
  }, {});
```

VolodymyrKirichenko (approved):
Well done 🔥 
But plz, review my comments
I recommend using the reduce method

PR #3662, opened 2024-11-04
Solution

Mariagosp:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+  const str = sourceString;
+  const result = {};
+  const items = str.split(';').filter((item) => item.trim());
+
+  items.forEach((item) => {
+    const parts = item.split(':');
+    const key = parts[0].trim();
+    const value = parts.slice(1).join(':').trim();
+
+    result[key] = value;
```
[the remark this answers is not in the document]
VolodymyrKirichenko: Good point 🔥 
Do it

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const str = sourceString;
+  const result = {};
+  const items = str.split(';').filter((item) => item.trim());
```
VolodymyrKirichenko, line 11: I think this is an redundant. u can use the `reduce` method, and perform all operations in the middle of each iteration

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
+
+  items.forEach((item) => {
+    const parts = item.split(':');
+    const key = parts[0].trim();
+    const value = parts.slice(1).join(':').trim();
+
+    result[key] = value;
+  });
+
+  return result;
```
VolodymyrKirichenko, line 21: better option for u

```suggestion
  return sourceString.split(';').reduce((style, parts) => {
  
   ....................

    return style;
  }, {});
```

VolodymyrKirichenko (changes requested):
Good job 👍 
But let's practice the best method for objects - `reduce`

VolodymyrKirichenko (approved):

PR #3664, opened 2024-11-05
Develop

Krrampuss:


src/convertToObject.js
```diff
@@ -6,7 +7,24 @@
...
+      const key = parts[0].trim();
+      const value = parts.length > 1 ? parts.slice(1).join(':').trim() : '';
+
+      if (key) {
+        result[key] = value;
+      }
+    }
+  });
+
+  return result;
```
VolodymyrKirichenko, line 27: i think, that we can use `reduce` method for it. Let's try it
```suggestion
  return sourceString.split(';').reduce((style, declaration) => {
    
    ...........
    
    return style;
  }, {});
```

VolodymyrKirichenko (changes requested):
Good job, but we can do it better

VolodymyrKirichenko (approved):

PR #3665, opened 2024-11-05
solution

k0tlik02:


src/convertToObject.js
```diff
@@ -6,7 +6,25 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+  const lines = sourceString.replace('\n', '').split(';');
```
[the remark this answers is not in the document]
VolodymyrKirichenko: It looks like something extra and not necessary.
It will work without `replace`

src/convertToObject.js
```diff
@@ -6,7 +6,25 @@
...
+    line = line.trim();
+
+    if (!line.includes(':')) {
+      continue;
+    }
+
+    const [property, value] = line.split(':');
+
+    const cleanProperty = property.trim();
+    const cleanValue = value.replace(';', '').trim();
```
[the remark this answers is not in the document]
VolodymyrKirichenko: Good point. Remove it

src/convertToObject.js
```diff
@@ -6,7 +6,25 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+  const lines = sourceString.replace('\n', '').split(';');
+
+  for (let line of lines) {
+    line = line.trim();
```
VolodymyrKirichenko: why?
if u want to do this - create a separate const variable for this
```suggestion
  for (const line of lines) {
    const trimmedLine = line.trim();
```

src/convertToObject.js
```diff
@@ -6,7 +6,25 @@
...
+
+    const [property, value] = line.split(':');
+
+    const cleanProperty = property.trim();
+    const cleanValue = value.replace(';', '').trim();
+
+    styles[cleanProperty] = cleanValue;
+  }
+
+  return styles;
```
VolodymyrKirichenko: I think u can do it using the reduce method, I recommend refactoring it into a task using it. But up to u

```suggestion
  return sourceString.split(';').reduce((style, line) => {
    const [property, value] = line.trim().split(':');

    ............

    return style;
  }, {});
```

src/convertToObject.js
```diff
@@ -6,7 +6,25 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
```
VolodymyrKirichenko: Change the name of the object, which will not conflict with the `style` attribute
use `stylesObject`, `styleConfig`, `customStyles` for example

VolodymyrKirichenko (changes requested):
Good job 👍 
But let's make some fixes

VolodymyrKirichenko (approved):

PR #3672, opened 2024-11-06
Solution

koguts:


src/convertToObject.js
```diff
@@ -5,8 +5,20 @@
...
-function convertToObject(sourceString) {
-  // write your code here
+function convertToObject(styles) {
+  const styleObject = {};
+
+  styles
+    .split(';')
+    .map((style) => style.trim())
+    .filter((style) => style.length > 0)
+    .forEach((style) => {
```
BudnikOleksii, line 15: You can use `reduce` to generate an object

BudnikOleksii (approved):
👍

PR #3675, opened 2024-11-06
add task solution

taniabarkovskya:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const convertedString = sourceString
+    .split(';')
+    .filter((style) => style.trim() !== '')
```
```suggestion
    .filter((style) => style.trim())
```
You don't need to compare with an empty string, just check if value exists

src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
+    .reduce((prev, element) => {
+      const splittedStyle = element.split(':').map((el) => el.trim());
```
```suggestion
      const [property, value] = element.split(':').map((el) => el.trim());
```
You can use descturcting

BudnikOleksii (approved):
Good work 👍

PR #3678, opened 2024-11-06
solution

kami2693:


src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {};
+  const rules = sourceString.split(';');
+
+  rules.forEach((rule) => {
```
BudnikOleksii, line 12: You can use `reduce` to generate an object

BudnikOleksii (approved):
Good work 👍

PR #3680, opened 2024-11-06
solution

yuliiazubenko:


src/convertToObject.js
```diff
@@ -5,8 +5,29 @@
...
+    }
+
+    const colonIndex = trimmedDecl.indexOf(':');
+
+    if (colonIndex === -1) {
+      return accumulator;
+    }
+
+    const property = trimmedDecl.slice(0, colonIndex).trim();
+    const value = trimmedDecl.slice(colonIndex + 1).trim();
```
BudnikOleksii, line 23: You can use `split` instead of `indexOf` and `slice`

BudnikOleksii (approved):
Good work 👍

PR #3681, opened 2024-11-06
solution

Fridrif27:


src/convertToObject.js
```diff
@@ -6,7 +6,28 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const declarations = sourceString.trim().split(';');
+
+  const stylesObj = {};
+
+  declarations.forEach((declaration) => {
```
BudnikOleksii: You can use `reduce` to generate an object

src/convertToObject.js
```diff
@@ -6,7 +6,28 @@
...
+    const parts = declaration.split(':');
```
```suggestion
    const [property, value] = declaration.split(':');
```
You can use destructing

src/convertToObject.js
```diff
@@ -6,7 +6,28 @@
...
+
+    if (parts.length === 2) {
+      const property = parts[0].trim();
+      let value = parts[1].trim();
+
+      if (value.includes('!important')) {
+        value = value.replace('!important', '').trim() + ' !important';
```
BudnikOleksii: Do you actually need this?

BudnikOleksii (changes requested):
Good work 👍 

BudnikOleksii (approved):
LGTM 🚀

PR #3682, opened 2024-11-06
Solution

save-v:


src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  let res = sourceString.split(';');
+
+  res = res.map((el) => el.split(':'));
+  res = res.map((el) => el.map((item) => item.trim()));
+  res = res.filter((item) => item.length > 1);
```
BudnikOleksii: - res/result/el/elem/obj/object/arr/array and etc are bad names, name should describe variable
- Prefer const over let, you can call all iteration method in chain

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution

Moroz-Dmytro (approved):
👍

PR #3683, opened 2024-11-06
Solution

schyrva:


src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
...
+  return sourceString
+    .split(';')
+    .map((declaration) => declaration.trim())
+    .filter((declaration) => declaration)
+    .reduce((sourceObject, declaration) => {
+      const colonIndex = declaration.indexOf(':');
+
+      if (colonIndex !== -1) {
+        const property = declaration.slice(0, colonIndex).trim();
+        const value = declaration.slice(colonIndex + 1).trim();
```
BudnikOleksii: Just use `split` instead of `indexOf` and `slice`

BudnikOleksii (approved):
👍

PR #3685, opened 2024-11-06
Solution

LiliiaVol:


src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const res = sourceString.split(';').reduce((acc, str) => {
```
BudnikOleksii, line 9: res/result/el/elem/arr/array/obj/object and etc are bad names, name should describe variable

BudnikOleksii (approved):
Good job 👍

PR #3686, opened 2024-11-07
Solution

nataliia2211:


src/convertToObject.js
```diff
@@ -6,7 +6,26 @@
...
+
+  if (sourceString.length === 0) {
+    return styles;
+  }
+
+  sourceString
+    .split(';')
+    .map((el) => el.trim())
+    .filter((el) => el.includes(':'))
+    .forEach((el) => {
```
BudnikOleksii, line 19: - arr/array/str/string/el/elem/obj/object/res/result are bad names, name should describe variable
- You can use `reduce` to generate an object with it

BudnikOleksii (approved):
Good work 👍

PR #3689, opened 2024-11-07
create solution

Andrii-Yanchuk:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const transformedString = sourceString
+    .split(';')
+    .filter((style) => style.trim() !== '')
```
BudnikOleksii, line 11:
```suggestion
    .filter((style) => style.trim())
```
You don't need to compare with an empty string, just check if exists

src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
+    .reduce((acc, element) => {
+      const splitedEl = element.split(':').map((el) => el.trim());
```
BudnikOleksii, line 13:
```suggestion
      const [propery, value] = element.split(':').map((el) => el.trim());
```
You can use descturcting

BudnikOleksii (approved):
Good job 👍

PR #3690, opened 2024-11-07
solution

Yevhen-Filimonov:


src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const obj = {};
+
+  const stylesArray = stylesString
```
BudnikOleksii, line 11: res/result/el/elem/arr/array/obj/object and etc are bad names, variable name should be descriptive

src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
...
+    .split(';')
+    .map(style => style.trim())
+    .filter(style => style.length > 0);
+
+  stylesArray.forEach(style => {
```
BudnikOleksii, line 16: You can use `reduce` to generate an object.

src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
...
+    const [property, value] = style.split(':').map(item => item.trim());
```
BudnikOleksii, line 17: `item` is not informative name as well

BudnikOleksii (changes requested):
Good job 👍
Left a few comments to improve your solution

PR #3692, opened 2024-11-07
add solution

TargoniyAnatoliy:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const input = sourceString.split(';');
+  const result = {};
```
BudnikOleksii: res/result/el/elem/arr/array/obj/object and etc are bad names, variable name should be descriptive

src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
+
+  input.forEach((item) => {
```
BudnikOleksii: You can use `reduce` to generate an object. `item` is not informative name as well

src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
+    const [key, value] = item.split(':');
+
+    if (value !== undefined) {
```
```suggestion
    if (value) {
```
You don't need to compare with undefined, just check if a value exists

BudnikOleksii (changes requested):
Good job 👍 
Left a few comments to improve your solution

BudnikOleksii (changes requested):
.

src/convertToObject.js
```diff
@@ -6,7 +6,14 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const input = sourceString.split(';');
+  const result = input.reduce((acc, property) => {
```
BudnikOleksii, line 10: res/result/el/elem/arr/array/obj/object and etc are bad names, variable name should be descriptive

BudnikOleksii (approved):
Good work 👍

PR #3694, opened 2024-11-07
Solution

maxsabo:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
-}
+  const styleProperties = {};
+  const styleDeclarations = sourceString
+    .split(';')
+    .filter((style) => style.length !== 0);
+
+  styleDeclarations.forEach((style) => {
```
BudnikOleksii, line 14: You can use `reduce` to generate an object

BudnikOleksii (approved):
Good work 👍

PR #3696, opened 2024-11-07
solve task

KostiantynBond:


src/convertToObject.js
```diff
@@ -7,6 +7,22 @@
  */
 function convertToObject(sourceString) {
   // write your code here
+  const stylesArray = sourceString.trim().split(';');
+  const stylesObject = {};
+
+  stylesArray.forEach((style) => {
```
BudnikOleksii, line 13: You can use reduce to generate an object

BudnikOleksii (approved):
Good work 👍

PR #3698, opened 2024-11-08
Solution

so3r:


src/convertToObject.js
```diff
@@ -6,7 +6,22 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+
+  const lines = sourceString
+    .split(';')
+    .map((line) => line.trim())
+    .filter((line) => line.length > 0);
+
+  lines.forEach((line) => {
```
BudnikOleksii, line 16: You can use `reduce` to generate an object

src/convertToObject.js
```diff
@@ -6,7 +6,22 @@
...
+    const [property, value] = line.split(':').map((part) => part.trim());
+
+    if (property && value !== undefined) {
```
```suggestion
    if (property && value) {
```
You don't need to compare with undefined, just check if a value exists

BudnikOleksii (approved):
Good job 👍

PR #3699, opened 2024-11-08
Solution

AnderZerfall:


src/convertToObject.js
```diff
@@ -6,7 +6,21 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObj = {};
+
+  const styles = sourceString.split(';');
+
+  styles.forEach((style) => {
```
BudnikOleksii, line 13: You can use reduce to generate an object

BudnikOleksii (approved):
Good work 👍

PR #3700, opened 2024-11-08
done task

Jion353:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const cssStyle = {};
+  const arrStr = sourceString.split(';');
+
+  for (const pieceArr of arrStr) {
```
BudnikOleksii, line 12: - arr/array/str/string/el/elem/obj/object/res/result are bad names, name should describe variable
- The task requires using iteration methods, the best option is `reduce` in this case, you generate an object with it

BudnikOleksii (changes requested):
Good work 👍 
Let's fix naming and try to do the task with reduce

BudnikOleksii (changes requested):
.

PR #3701, opened 2024-11-08
solution

Ilya-Ilchuk:


src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+
+  sourceString.split(';').forEach((style) => {
```
BudnikOleksii, line 11: You can use `reduce` instead of `forEach` to generate an object

BudnikOleksii (approved):
Good work 👍

PR #3702, opened 2024-11-09
Solution

SemenVodolazskij:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleCss = {};
+  const changeCss = sourceString.split(';');
+
+  changeCss.forEach((item) => {
```
BudnikOleksii, line 12: You can use `reduce` to generate an object

src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
+    const [key, value] = item.split(':');
+
+    if (value !== undefined) {
```
```suggestion
    if (value) {
```
You don't need to compare with undefined, just check if value exists

BudnikOleksii (approved):
Good work 👍

PR #3705, opened 2024-11-10
Solution

StricTGod:


src/convertToObject.js
```diff
@@ -6,7 +6,18 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(';')
+    .map((style) => style.trim())
+    .reduce((acc, curr) => {
+      const [key, value] = curr.split(':').map((str) => str.trim());
+
+      if (key && value !== undefined) {
```
```suggestion
      if (key && value) {
```
You don't have to compare with undefined, just check if value exists

BudnikOleksii (approved):
LGTM 🔥

PR #3719, opened 2024-11-18
solution

dmdamyan:


src/convertToObject.js
```diff
@@ -6,7 +6,13 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const array = [...sourceString.split(';')];
```
etojeDenys:
```suggestion
  const array = sourceString.split(';');
```
looks redundant

etojeDenys (changes requested):
you need to pass all tests, feel free to ask for some help in the chat

src/convertToObject.js
```diff
@@ -6,7 +6,17 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const newObject = sourceString.split(';').reduce((acc, elem) => {
+    const [key, value] = elem.split(':');
+
+    if (key && value) {
+      acc[key.trim()] = value.trim();
```
volodymyr-soltys97, line 13: You can simplify it
```suggestion
    const [key, value] = elem.split(':').map((it) => it.trim());

    if (key && value) {
      acc[key] = value;
```

volodymyr-soltys97 (approved):
Great work! 🔥

PR #3731, opened 2024-11-22
Solution

naz4ik:


src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleObject = {};
+
+  const declarations = sourceString.split(';');
+
+  for (let declaration of declarations) {
```
volodymyr-soltys97, line 13: Don't need to use for loop here, use method instead

src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
...
+    // Trim spaces from each declaration
+    declaration = declaration.trim();
+
+    if (declaration === '' || !declaration.includes(':')) {
+      continue;
+    }
```
volodymyr-soltys97, line 19: Redundant
```suggestion
```

src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
...
+
+    const [key, value] = declaration.split(':');
```
volodymyr-soltys97, line 21: You can use trim method like this
```suggestion
    const [key, value] = declaration.split(':').map((it) => it.trim());
```

src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
...
+
+    styleObject[key.trim()] = value.trim();
```
volodymyr-soltys97, line 23: And after remove here
```suggestion
    styleObject[key] = value;
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

PR #3750, opened 2024-12-05
Solution

IvanVaniuk:


src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleArray = sourceString.replace(/^\n+/g, '').split(';');
```
volodymyr-soltys97, line 9: Redundant
```suggestion
  const styleArray = sourceString.split(';');
```

src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
...
+  const data = {};
+
+  styleArray.forEach((elem) => {
+    const style = elem.trim();
+
+    if (!style) {
+      return;
+    }
+
+    const [key, value] = style.split(':').map((item) => item.trim());
```
volodymyr-soltys97, line 19: You can simplify like this
```suggestion
    const [key, value] = elem.split(':').map((item) => item.trim());
```

volodymyr-soltys97 (approved):
Good job! 👍

volodymyr-soltys97 (approved):

PR #3782, opened 2024-12-20
string to object converter

danylivOstap:


src/convertToObject.js
```diff
@@ -1,12 +1,29 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .split(SEMICOLON)
+    .map((line) => line.trim())
+    .filter((line) => line && line.includes(COLON))
+    .map((line) => line.split(COLON))
+    .reduce((prev, word) => {
+      const key = word[FIRST_INDEX].trim();
+      let value = word[SECOND_INDEX].trim();
```
```suggestion
    .reduce((prev, line) => {
      const [key, value] = line.split(COLON).map(work => word.trim());
```
Could be simplified

BudnikOleksii (approved):
👍

PR #3920, opened 2025-02-13
Solution

IvanIlnytskyy:


src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleArray = sourceString.split(';').filter(Boolean);
+
+  return styleArray.reduce((styleObj, style) => {
+    const [key, value] = style.split(':').map((segment) => {
+      return segment.trim();
+    });
```
volodymyr-soltys97: You can simplify it
```suggestion
    const [key, value] = style.split(':').map((segment) => segment.trim());
```

src/convertToObject.js
```diff
@@ -6,7 +6,19 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styleArray = sourceString.split(';').filter(Boolean);
```
volodymyr-soltys97: Redundant
```suggestion
  const styleArray = sourceString.split(';');
```

volodymyr-soltys97 (approved):
Great work! 🔥

PR #3951, opened 2025-02-19
Solution

DariaRadtkina:


src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const result = sourceString
+    .split(';')
+    .map((rows) => rows.trim())
+    .filter((newRows) => newRows.trim())
+    .map((row) => row.split(':'))
```
igorkruz, line 13: Reduce Unnecessary `.map()` Calls

src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
...
+    .map((element) => ({
+      [element[0]]: element[1],
+    }))
+    .reduce((newObj, thatObj) => {
+      for (const key in thatObj) {
+        newObj[key] = thatObj[key];
+      }
+
+      return newObj;
+    }, {});
```
igorkruz, line 24:
```suggestion
    .reduce((acc, rule) => {
      const [property, value] = rule.split(':').map(str => str.trim());

      if (property && value) {
        acc[property] = value;
      }

      return acc;
    }, {});
```

src/convertToObject.js
```diff
@@ -6,7 +6,24 @@
...
+  const result = sourceString
+    .split(';')
+    .map((rows) => rows.trim())
+    .filter((newRows) => newRows.trim())
+    .map((row) => row.split(':'))
+    .map(([word, value]) => [word.trim(), value.trim()])
+    .map((element) => ({
+      [element[0]]: element[1],
+    }))
+    .reduce((newObj, thatObj) => {
```
igorkruz, line 18: Use descriptive variable names

igorkruz (approved):
Great job 👍 
Please consider the added comments.

PR #3955, opened 2025-02-19
converter solution

reb0rned:


src/convertToObject.js
```diff
@@ -5,8 +5,21 @@
...
+
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .trim()
+    .split(';')
+    .filter((el) => el.trim() !== '')
+    .reduce((acc, el) => {
+      const [key, value] = el.split(':');
+      const keyResult = key.trim();
```
igorkruz, line 16: Add a check to ensure key and value exist before processing.

src/convertToObject.js
```diff
@@ -5,8 +5,21 @@
  *
  * @return {object}
  */
+
 function convertToObject(sourceString) {
-  // write your code here
+  return sourceString
+    .trim()
+    .split(';')
+    .filter((el) => el.trim() !== '')
```
igorkruz, line 13: `el `should clearly describe what each item represents in the context of the function. Choose a better name instead

igorkruz (approved):
Great job 👍 
Please consider the added comments.

PR #3962, opened 2025-02-19
add js

glebanty19:


igorkruz (approved):
Good Job 👍
You can improve your solution using reduce() Instead of forEach()

PR #3996, opened 2025-03-01
Solution

KvashaIhor:


etojeDenys (approved):
you can simplify it using the reduce, but looks good

PR #4065, opened 2025-03-26
add solution

NazarMykhailiv:


NazarMykhailiv:
@mateacademy-ai-mentor
I read your review — it's not correct.
The project uses ESLint + Prettier, which automatically format the code and enforce indentation rules.
I'm using 4 spaces, and all chained methods (.map, .filter, .reduce) are properly aligned according to the configured standard.
If anything was wrong, ESLint or Prettier would block the push.
Formatting is already correct, and the changes were made earlier. I didn’t commit anything new because nothing needed to be changed.

PR #4112, opened 2025-04-12
attempt 1

A1cmae0n:
Used AI help only when it came to the edge-case where there is a CSS property without a colon. Before, I did not check for whether a property had a colon at all, and it seems to pass the case in the given example. However, because of difficulties with backwards-compatability of Mate's in-house testing scripts, I am unable to run `npm test1`.

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -6,24 +6,24 @@
  * @return {object}
  */
 function convertToObject(sourceString) {
-  const cssDeclarations = sourceString.split(';');
-  let obj = {};
+  return sourceString
+    .split(';')
+    .map((declaration) => declaration.trim())
+    .filter((declaration) => declaration)
+    .reduce((result, declaration) => {
+      const colonPos = declaration.indexOf(':');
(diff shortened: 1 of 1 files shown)


A1cmae0n:
Admittedly, this attempt was AI-heavy. While my first attempt was low on AI and only integrated a check for an edge case that may not have been relevant, the requirement to use array iteration methods had me going to AI for help. It told me how to string split(), filter(), map(), and finally reduce() together. Reduce was the most insightful: it showed how I can use an empty object as an accumulator, and interact with this object at each iteration: for instance to add a new entry. It also showed me how I could simply return this object if a condition was not met, making a workaround for the continue() function that is otherwise unavailable in array iteration methods. Admittedly, I do not quite know what filter() is doing, simply filtering if each declaration exists. I suppose this could eliminate white space. If so, it is a useful tool. I believe that going over AI-generated code is useful to understand it. While I could have struggled to string together array iteration methods, sometimes it is more useful to learn from an example, especially when the coder has not been exposed to best-practices.

PR #4152, opened 2025-04-30
Solution

izuwiiii:


izuwiiii:
I check for empty strings here     if (s[1]) {
      obj[`${s[0].trim()}`] = s[1].trim()
    }

PR #4184, opened 2025-05-04
sol CHECK desc

vladyslavchichkan:
SOLUTION was push by Chat gpt because linters gives f.. issues

ORIGINAL SOLUTION is here

function removeSpaces(string) {
    return string.split(" ").filter(Boolean).join("")
};

function getObjectFromArrOfTwoStrings(s1 = null, s2 = null) {
    return {[s1]: s2} 
}

  // write your code here
  return (sourceString
    .split(";")
    .map(
      (string) => getObjectFromArrOfTwoStrings(...
          removeSpaces(string)
          .replace(/\n/g, "")
          .split(":")
      )
    )
    .filter(obj => Object.keys(obj)[0])
  )
}


src/convertToObject.js
```diff
@@ -2,11 +2,24 @@
...
+    const property = match[1].trim();
+    let value = match[2];
+
+    value = value.replace(/[\r\t]+/g, '');
+    value = value.trim(); 
+
+    stylesObject[property] = value;
+  }
+
+  return stylesObject;
```
BudnikOleksii, line 22: - You don't need a `regex` and `while` loop in this case
- If you have a `regex` specify name so everyone understand what it does

BudnikOleksii (approved):
If you have a linter errors just fix them instead of changing solution

PR #4228, opened 2025-05-19
Solution 19.05.25

Alya-ya:


src/convertToObject.js
```diff
@@ -6,7 +6,25 @@
...
-  // write your code here
+  const lines = sourceString.split(';');
+
+  const result = lines.reduce((acc, curr) => {
+    const clean = curr.trim();
+
+    if (clean.includes(':')) {
+      const parts = clean.split(':', 2);
+      const clean1 = parts[0].trim();
+      const clean2 = parts[1].trim();
```
TarasHoliuk, line 17: FYI: We know that in the first part we have `key` or `property` in terms of CSS and in the second part we have value, so such names are better than `clean1` and `clean2`

src/convertToObject.js
```diff
@@ -6,7 +6,25 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const lines = sourceString.split(';');
+
+  const result = lines.reduce((acc, curr) => {
+    const clean = curr.trim();
+
+    if (clean.includes(':')) {
+      const parts = clean.split(':', 2);
```
TarasHoliuk, line 15:
```suggestion
      const [key, value] = clean.split(':', 2);
```
You can do it like that -immediately destructure it (but since we anyway have to trim it not very useful)

TarasHoliuk (approved):
GJ! Added 2 comments just FYI

src/convertToObject.js
Alya-ya: Okey , Thanks 🙏

PR #4244, opened 2025-05-27
Solution

MaxOnRem:
Implemented a convertToObject function that takes a string with styles and returns an object where CSS declarations are converted to keys and values. Applied Checklist code style recommenadtions. Importantly, all chained methods called on the same level should have same indentation, that makes code much more readable.

### DIFF
--- a/src/convertToObject.js
+++ b/src/convertToObject.js
@@ -27,7 +27,12 @@ function convertToObject(sourceString) {
     // If we have both property and value, add them to the result object
     const [property, value] = propertyPair;
 
-    if (property && value) {
+    // if (property && value) {
+    //   result[property] = value;
+    // }
+
+    // ensure that empty property names are not included
+    if (property !== '' && value) {
       result[property] = value;
     }
   });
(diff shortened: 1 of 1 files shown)


src/convertToObject.js
```diff
@@ -1,12 +1,45 @@
...
+const string = `
+  cursor: pointer;
+  float: left;
+  font-family: inherit;`;
+
 function convertToObject(sourceString) {
-  // write your code here
+  const result = {}; // Initialize an empty object to store key-value pairs
+
+  // Split the string into individual declarations using ';' as the separator
```
maxim2310, line 18: remove comments

maxim2310 (changes requested):
Good solution, just clean it up

src/convertToObject.js
MaxOnRem: plese don't decline my soultion, those comments are just for my understanding the code. But soultion, I think, is pretty decent

anastasiiavorobiova (approved):

PR #4648, opened 2025-12-09
solution added

Olena-Dolna:


2pasha (changes requested):
hi!

right now core functionality is valid, but your solution is to complex. You have looped through array near 4 times, it can be optimized. 

If you do not understand how to do this correctly - you may ask in fe_chat for help and we will help you to understand how to make this solution more optimized.

vitalii-bondarenko-ua (approved):
Looks good

PR #4659, opened 2025-12-15
Solution

MaksOther:


src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+  const splitStrings = sourceString.split(';');
+
+  splitStrings
+    .map((line) => line.trim())
+    .filter((line) => line.length)
+    .filter((line) => line.includes(':'))
+    .forEach((line) => {
```
dothearts, line 16: Better option use variables, not chaining.

src/convertToObject.js
```diff
@@ -6,7 +6,20 @@
...
  */
 function convertToObject(sourceString) {
-  // write your code here
+  const styles = {};
+  const splitStrings = sourceString.split(';');
+
+  splitStrings
+    .map((line) => line.trim())
+    .filter((line) => line.length)
+    .filter((line) => line.includes(':'))
```
dothearts, line 15: You can check in forEach iteration.
In general, you can use one method that help you do all the stuff using 1 iteration. 
Another way, use several itteration, but move condition (validation) in forEach in this case.

dothearts (changes requested):
Good job. 
Add some comments to improve your solution. 
Many thanks!

PR #4725, opened 2026-01-12
Solution

NeustrVk:


maxim2310 (approved):
Very clean work, I even suspected that it wasn't a student who did it) but if I'm still wrong, then it's really very good code

PR #4841, opened 2026-03-20
solution

thmoro:


thmoro:
@mateacademy-ai-mentor I tried this before (' A simple value = value.trim(); would be sufficient for all cases described in the task and would make the code simpler.'), but it causes a fail test which is this one 'convertToObject should › support multiline values'

